PR #3, opened 2015-08-10
Sort command line options and class fields

antiagainst:
* Sort command line options in `main.c` to make it consistent with the manual.
* Sort and group fields in compiler-related classes.
* Add missing blurbs for some fields.
* Add the `-I` option in `glslc` manual.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ cd $SOURCE_DIR/
 cd $BUILD_DIR
 cmake -GNinja -DCMAKE_BUILD_TYPE={Debug|Release|RelWithDebInfo} $SOURCE_DIR
 ninja
-ctest #optional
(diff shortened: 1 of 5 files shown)


dmircevski:
Looks good, except for one general concern: does this mean we intend to permanently sync the ordering in main.cc and the manual?  So every time I change something in the manual, I have to update main.cc as well?

antiagainst:
@dmircevski I would think that's reasonable. It will keep options in the same group placed together, better for locating & maintenance.

dmircevski:
I bet you it falls out of sync over time (even if we document it somewhere prominent) and nobody notices because it makes little difference.

+2 anyway, because it doesn't hurt.  And :+1: for adding blurbs. :)

PR #4, opened 2015-08-10
Fix a problem in command line option parsing.

antiagainst:
When a command line option ends with an equal sign, it's argument
is what follows that equal sign, even if it's empty.

### DIFF
--- a/glslc/src/main.cc
+++ b/glslc/src/main.cc
@@ -68,11 +68,10 @@ void PrintHelp(std::ostream* out) {
        << std::endl;
 }
 
-// Parses the argument for the option at index in argv. On success, returns
-// true and writes the parsed argument into option_argument. Returns false
-// if any error occurs. option is the expected option at argv[index].
-// After calling this function, index will be at the last command line
-// argument consumed.
+// Gets the option argument for the option at *index in argv in a way consistent
+// with clang/gcc. On success, returns true and writes the parsed argument into
+// *option_argument. Returns false if any errors occur. After calling this
+// function, *index will the index of the last command line argument consumed.
 bool GetOptionArgument(int argc, char** argv, int* index,
                        const std::string& option,
                        string_piece* option_argument) {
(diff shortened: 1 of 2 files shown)


glslc/test/option_std.py
```diff
@@ -29,6 +29,30 @@ def core_frag_shader_without_version():
 
 
 @inside_glslc_testsuite('OptionStd')
+class TestStdNoArg(expect.ErrorMessage):
+    """Tests -std."""
```
dmircevski: Consider "Tests -std alone."

glslc/src/main.cc
```diff
@@ -82,12 +82,13 @@ bool GetOptionArgument(int argc, char** argv, int* index,
     *option_argument = arg.substr(option.size());
     return true;
   } else {
-    if (++(*index) >= argc) {
-      return false;
-    } else {
-      *option_argument = argv[*index];
+    if (option.back() == '=') {
```
dmircevski: This merits a mention in the blurb.

antiagainst:
@dmircevski: done

glslc/src/main.cc
```diff
@@ -69,10 +69,11 @@ void PrintHelp(std::ostream* out) {
...
 
 // Parses the argument for the option at index in argv. On success, returns
-// true and writes the parsed argument into option_argument. Returns false
-// if any error occurs. option is the expected option at argv[index].
-// After calling this function, index will be at the last command line
-// argument consumed.
+// true and writes the parsed argument into option_argument. Returns false if
+// any error occurs. The option parameter is the expected option at argv[index].
+// If the option parameter ends with an equal sign, its argument is whatever
+// follows that equal sign, even empty string. After calling this function,
```
dmircevski: And if it's not an equal sign, then the argument is either the rest of argv[index] or argv[++index]?

antiagainst:
@dmircevski done

antiagainst:
@dmircevski revised.

dmircevski:
+2

PR #5, opened 2015-08-11
Support for -working-directory

antiagainst:
This patch enables `glslc` to accept `-working-directory` as an option.
Input and output files are resolved against the working directory if
necessary. Setting working directory with the existence of the `#include`
directive is not implemented yet.

### DIFF
--- a/glslc/src/file_compiler.h
+++ b/glslc/src/file_compiler.h
@@ -42,6 +42,9 @@ class FileCompiler : public shaderc_util::Compiler {
   bool CompileShaderFile(const std::string& input_file,
                          EShLanguage shader_stage);
 
+  // Sets the working directory for the compilation.
+  void SetWorkingDirectory(const std::string& dir);
+
   // Adds a directory to be searched when processing #include directives.
   //
   // Best practice: if you add an empty string before any other path, that will
@@ -74,6 +77,10 @@ class FileCompiler : public shaderc_util::Compiler {
  private:
   // Returns the name of the output file, given the input_filename string.
   std::string GetOutputFileName(std::string input_filename);
+
+  // Resolves relative paths against this working directory. Must always end
+  // in '/'.
+  std::string workdir_;
 
   // A FileFinder used to substitute #include directives in the source code.
   shaderc_util::FileFinder include_file_finder_;
--- a/glslc/src/main.cc
+++ b/glslc/src/main.cc
@@ -62,6 +62,9 @@ void PrintHelp(std::ostream* out) {
        << "  -S                Only run preprocess and compilation steps.\n"
        << "  -w                Suppresses all warning messages.\n"
        << "  -Werror           Treat all warnings as errors.\n"
+       << "  -working-directory <dir>\n"
+       << "                    Resolve file paths relative to the specified "
+       << "directory.\n"
        << "  -x <language>     Treat subsequent input files as having type "
        << "<language>.\n"
        << "                    The only supported language is glsl."
@@ -116,6 +119,18 @@ int main(int argc, char** argv) {
         return 1;
       }
       compiler.SetOutputFileName(file_name);
+    } else if (arg.starts_with("-working-directory")) {
+      // Following Clang, both -working-directory and -working-directory= are
+      // accepted.
+      std::string option = "-working-directory";
+      if (arg[option.size()] == '=') option = "-working-directory=";
+      string_piece workdir;
+      if (!GetOptionArgument(argc, argv, &i, option, &workdir)) {
+        std::cerr << "glslc: error: argument to '-working-directory' is "
+                     "missing (expected 1 value)"
+                  << std::endl;
+      }
+      compiler.SetWorkingDirectory(workdir.str());
     } else if (arg.starts_with("-fshader-stage=")) {
       const string_piece stage = arg.substr(std::strlen("-fshader-stage="));
       force_shader_stage = glslc::GetShaderStageFromCmdLine(arg);
--- a/glslc/test/working_directory.py
+++ b/glslc/test/working_directory.py
@@ -1,0 +1,167 @@
+# Copyright 2015 The Shaderc Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os.path
+
+import expect
+from environment import File, Directory
+from glslc_test_framework import inside_glslc_testsuite
+from placeholder import FileShader
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirNoArg(expect.ErrorMessage):
+    """Tests -working-directory. Behavior cribbed from Clang."""
+
+    glslc_args = ['-working-directory']
+    expected_error = [
+        "glslc: error: argument to '-working-directory' is missing "
+        '(expected 1 value)\n',
+        'glslc: error: no input files\n']
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirEqNoArg(expect.ErrorMessage):
+    """Tests -working-directory=<empty>. Behavior cribbed from Clang."""
+
+    glslc_args = ['-working-directory=']
+    expected_error = ['glslc: error: no input files\n']
+
+
+EMPTY_SHADER_IN_SUBDIR = Directory(
+    'subdir', [File('shader.vert', 'void main() {}')])
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirEqNoArgCompileFile(expect.ValidNamedObjectFile):
+    """Tests -working-directory=<empty> when compiling input file."""
+
+    environment = Directory('.', [EMPTY_SHADER_IN_SUBDIR])
+    glslc_args = ['-c', '-working-directory=', 'subdir/shader.vert']
+    # Output file should be generated into subdir/.
+    expected_object_filenames = ('subdir/shader.vert.spv',)
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirEqNoArgLinkFile(expect.ValidNamedObjectFile):
+    """Tests -working-directory=<empty> when linking input file."""
+
+    environment = Directory('.', [EMPTY_SHADER_IN_SUBDIR])
+    glslc_args = ['-working-directory=', 'subdir/shader.vert']
+    # Output file should be generated into the current directory, not
+    # subdir/, where the source is.
+    expected_object_filenames = ('a.spv',)
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirCompileFile(expect.ValidNamedObjectFile):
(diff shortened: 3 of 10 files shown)


glslc/src/file_compiler.h
```diff
@@ -75,6 +78,9 @@ class FileCompiler : public shaderc_util::Compiler {
   // Returns the name of the output file, given the input_filename string.
   std::string GetOutputFileName(std::string input_filename);
 
+  // Resoloves relative paths against this working directory.
```
dmircevski: Must always end in '/'?

dmircevski: Consider "Relative paths of files on the command line are resolved against this directory."

glslc/src/main.cc
```diff
@@ -116,6 +119,16 @@ int main(int argc, char** argv) {
         return 1;
       }
       compiler.SetOutputFileName(file_name);
+    } else if (arg.starts_with("-working-directory")) {
+      std::string option = "-working-directory";
+      if (arg[option.size()] == '=') option = "-working-directory=";
```
dmircevski: Is this how it works in clang?  The '=' is optional here but required for -std?

glslc/test/working_directory.py
```diff
@@ -0,0 +1,210 @@
...
+        '(expected 1 value)\n',
+        'glslc: error: no input files\n']
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirEqNoArg(expect.ErrorMessage):
+    """Tests -working-directory=<empty>."""
+
+    glslc_args = ['-working-directory=']
+    expected_error = ['glslc: error: no input files\n']
```
dmircevski: Should be "argument missing," like above.

libshaderc_util/src/io.cc
```diff
@@ -40,6 +40,11 @@ void OutputFileErrorMessage(int errno_value) {
 
 namespace shaderc_util {
 
+bool IsAbsolutePath(const std::string& path) {
+  // TODO(antiagainst): add support for Windows.
+  return !path.empty() && path.front() == '/';
```
dmircevski: That's not how Windows works. ;)

libshaderc_util/src/io_test.cc
```diff
@@ -34,6 +35,19 @@ class ReadFileTest : public testing::Test {
...
+  EXPECT_FALSE(IsAbsolutePath(""));
+  EXPECT_TRUE(IsAbsolutePath("/"));
+  EXPECT_FALSE(IsAbsolutePath("."));
+  EXPECT_FALSE(IsAbsolutePath(".."));
+  EXPECT_TRUE(IsAbsolutePath("/bin/echo"));
+  EXPECT_TRUE(IsAbsolutePath("//etc/shadow"));
+  EXPECT_TRUE(IsAbsolutePath("/../../../lib"));
+  EXPECT_FALSE(IsAbsolutePath("./something"));
+  EXPECT_FALSE(IsAbsolutePath("input"));
+  EXPECT_FALSE(IsAbsolutePath("../test"));
```
dmircevski: Add a case of C:\.  Consider paths with spaces or Unicode glyphs in them.

dmircevski:
Also need to update README.asciidoc.

glslc/test/working_directory.py
dmircevski: Add a comment that this behaviour is cribbed from clang.

glslc/test/working_directory.py
```diff
@@ -0,0 +1,210 @@
...
+class TestWorkDirEqNoArgLinkFile(
+    expect.SuccessfulReturn, expect.CorrectObjectFilePreamble):
+    """Tests -working-directory=<empty> when linking input file."""
+
+    environment = Directory('.', [
+        Directory('subdir', [File('shader.vert', 'void main() {}')])
+    ])
+    glslc_args = ['-working-directory=', 'subdir/shader.vert']
+
+    def check_object_file_preamble(self, status):
```
dmircevski: Comment on why override the default.

glslc/test/working_directory.py
```diff
@@ -0,0 +1,210 @@
...
+        return self.verify_object_file_preamble(
+            # output file should be generated into the current directory.
```
dmircevski: Move this comment up to the method level.

glslc/test/working_directory.py
```diff
@@ -0,0 +1,210 @@
...
+            os.path.join(status.directory, 'a.spv'))
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirCompileFile(
+    expect.SuccessfulReturn, expect.CorrectObjectFilePreamble):
+    """Tests -working-directory=<dir> when compiling input file."""
+
+    environment = Directory('.', [
```
dmircevski: DRY.  This same code repeats throughout.

glslc/test/working_directory.py
```diff
@@ -0,0 +1,210 @@
...
+class TestWorkDirLinkFile(
+    expect.SuccessfulReturn, expect.CorrectObjectFilePreamble):
+    """Tests -working-directory=<dir> when linking input file."""
+
+    environment = Directory('.', [
+        Directory('subdir', [File('shader.vert', 'void main() {}')])
+    ])
+    glslc_args = ['-working-directory=subdir', 'shader.vert']
+
+    def check_object_file_preamble(self, status):
```
dmircevski: DRY

glslc/test/working_directory.py
```diff
@@ -0,0 +1,210 @@
...
+class TestWorkDirCompileFile(
+    expect.SuccessfulReturn, expect.CorrectObjectFilePreamble):
+    """Tests -working-directory=<dir> when compiling input file."""
+
+    environment = Directory('.', [
+        Directory('subdir', [File('shader.vert', 'void main() {}')])
+    ])
+    glslc_args = ['-c', '-working-directory=subdir', 'shader.vert']
+
+    def check_object_file_preamble(self, status):
```
dmircevski: This could go into a base class (together with `environment`), parameterized by directory.

glslc/src/main.cc
dmircevski: Just tried with clang, and this is, indeed, correct (see below).  Maybe we should put a comment in, for the future reader.

The following commands all all work:
    clang++ -working-directory=.  /tmp/x.cc 
    clang++ -working-directory.  /tmp/x.cc 
    clang++ -working-directory .  /tmp/x.cc

But with -std:
    deki>clang++ -std=c++11 /tmp/x.cc 
    deki>clang++ -stdc++11 /tmp/x.cc 
    clang: error: unknown argument: '-stdc++11'
    deki>clang++ -std c++11 /tmp/x.cc 
    clang: error: unknown argument: '-std'
    clang: error: no such file or directory: 'c++11'

glslc/src/file_compiler.h
antiagainst: Updated the blurb to mention the invariant.

Not only relative paths of files on the command line. relative include files should also be resolved against this.

glslc/src/main.cc
antiagainst: Added comment for noting this.

libshaderc_util/src/io_test.cc
antiagainst: Done.

glslc/test/working_directory.py
antiagainst: Done.

glslc/test/working_directory.py
antiagainst: Done.

glslc/test/working_directory.py
antiagainst: I've created another expectation in `expect.py`, which simplifies this.

glslc/test/working_directory.py
antiagainst: Done.

antiagainst:
@dmircevski Updated `README.asciidoc`, DRY'd and addressed issues.

glslc/src/file_compiler.h
```diff
@@ -75,6 +78,10 @@ class FileCompiler : public shaderc_util::Compiler {
   // Returns the name of the output file, given the input_filename string.
   std::string GetOutputFileName(std::string input_filename);
 
+  // Resoloves relative paths against this working directory. Must always ends
```
dmircevski: Typo: "Resolves"

antiagainst: Done

glslc/test/working_directory.py
```diff
@@ -0,0 +1,166 @@
...
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirLinkFileOutput(expect.ValidNamedObjectFile):
+    """Tests -working-directory=<dir> when linking input file and specifying
+    output filename."""
+
+    environment = Directory('.', [
+        EMPTY_SHADER_IN_SUBDIR,
+        Directory('bin', [])
+    ])
```
dmircevski: Why is this different from line 91?  What would happen if `bin` was put inside `subdir`?

glslc/test/working_directory.py
```diff
@@ -0,0 +1,166 @@
...
+    glslc_args = ['-o', 'bin/spv', '-working-directory=subdir',
+                  'shader.vert']
+    # output file should be generated into bin/, not subdir/bin/.
+    expected_object_filenames = ('bin/spv',)
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirLinkFileOutputNotExist(expect.ErrorMessage):
+    """Tests -working-directory=<dir> when linking input file and specifying
+    output filename."""
```
dmircevski: Copy pasta :)

glslc/test/working_directory.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirEqInArg(expect.ValidNamedObjectFile):
+    """Tests -working-directory=<dir-with-equal-sign-inside>."""
+
+    environment = Directory('.', [
+        Directory('=subdir', [File('shader.vert', 'void main() {}')]),
+    ])
+    glslc_args = ['-working-directory==subdir', 'shader.vert']
```
dmircevski: Nice! :)

dmircevski:
+1

Just one question and one nitpick.  It's basically ready.

glslc/test/working_directory.py
antiagainst: The previous test case (Line 91) has `-c` in `glslc_args`. With `-c`, we are generating an intermediate (not the final linked) object file. The path given by `-o` should be resolved against `-working-directory`: `subdir/bin/spv`. While in this test case, we are generating a final linked object file. The path given by `-o` should be resolved against the **current** working directory: `bin/spv`. So I need create the current holding directory: in the previous test, its `subdir/bin/`, here it is `bin/`.

Moving `bin` into `subdir` will case the test case to fail with `No such file or directory` like the following case (starting from Line 118).

glslc/test/working_directory.py
antiagainst: Fixed :)

antiagainst:
@dmircevski Done.

glslc/test/working_directory.py
dmircevski: Then -working-directory doesn't affect the linked output file at all.  All those linked-file tests are basically just testing that -working-directory has no impact on linked-file designation.  After you check this in, I think I'd like to group them all together and name them better -- they only differ by their -wd and -o arguments.

dmircevski:
+2

glslc/test/working_directory.py
antiagainst: Yes, you are right. That's my understanding by trying clang. Feel free to reorganize it ;)

antiagainst:
Fixed the non-existing filename vs. dir name problem.

Rebased and squashed. Ready to merge.

PR #9, opened 2015-08-17
Sync with upstream glslang.

antiagainst:
`GL_OES_sample_variables` are now implemented in glslang, which
brings `gl_SampleID` into ES profile.


dneto0:
LGTM

PR #11, opened 2015-08-18
Rework link-file tests in working_directory.py.

dmircevski:
The link-file tests currently follow closely the compile-file tests, which is incorrect because `-working-directory` _shouldn't_ impact the link-file location.  Pull them into their own section, make them cover only `-o` variations (without varying the `-working-directory` forms, which are tested elsewhere), and rename them so its obvious what's different between them.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,0 +1,6 @@
+ROOT_SHADERC_PATH := $(call my-dir)
+
+include $(ROOT_SHADERC_PATH)/third_party/Android.mk
+include $(ROOT_SHADERC_PATH)/libshaderc_util/Android.mk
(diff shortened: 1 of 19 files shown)


antiagainst:
@dmircevski rebase needed ;)

dneto0:
What does "reworking" mean?
Are you just reordering the tests?  If so you should say so so that I know how to review the change.

dmircevski:
@dneto0 Not just reordering, but actually changing the tests.  I'll update the top comment.

dmircevski:
@antiagainst rebased.

dmircevski:
Abandoning, because I got screwed by rewritten history.  Will create a fresh pull request for this.

PR #14, opened 2015-08-19
Add secondary language around "fork."

dmircevski:
This will allow us to keep saying "fork" in the right context (though
still not in the top-level README).


antiagainst:
+1.

However, I'm not an expert on wording of this kind of stuff. Somebody else should have a double check.

dneto0:
+2

PR #15, opened 2015-08-19
Add tests for -working-directory involving multiple files.

antiagainst:


glslc/test/working_directory.py
```diff
@@ -81,10 +81,35 @@ class TestWorkDirPosition(expect.ValidNamedObjectFile):
     """Tests that -working-directory=<dir> affects all files including the
     one before it."""
```
dmircevski: s/including...it/, before and after it on the command line/

antiagainst: Done

dneto0:
+2

PR #17, opened 2015-08-19
Improve link-file tests in working_directory.py.

dmircevski:
The link-file tests currently follow closely the compile-file tests, which is incorrect because `-working-directory` _shouldn't_ impact the link-file location.  Pull them into their own section, make them cover only `-o` variations (without varying the `-working-directory` forms, which are tested elsewhere), and rename them so its obvious what's different between them.

### DIFF
--- a/glslc/test/working_directory.py
+++ b/glslc/test/working_directory.py
@@ -167,7 +167,8 @@ class WorkDirDoesntAffectLinkedFile(expect.ValidNamedObjectFile):
     environment = Directory('.', [
         Directory('subdir', [
             File('shader.vert', 'void main(){}'),
-            # This bin/ is in the wrong location, so glslc should ignore it.
+            # Try to fake glslc into putting the linked file here, though it
+            # shouldn't (because -working-directory doesn't impact -o).
             Directory('bin', [])]),
         File('shader.vert', "fake file, doesn't compile."),
         Directory('bin', [])])
--- a/glslc/test/working_directory.py
+++ b/glslc/test/working_directory.py
@@ -53,17 +53,6 @@ class TestWorkDirEqNoArgCompileFile(expect.ValidNamedObjectFile):
 
 
 @inside_glslc_testsuite('WorkDir')
-class TestWorkDirEqNoArgLinkFile(expect.ValidNamedObjectFile):
-    """Tests -working-directory=<empty> when linking input file."""
-
-    environment = Directory('.', [EMPTY_SHADER_IN_SUBDIR])
-    glslc_args = ['-working-directory=', 'subdir/shader.vert']
-    # Output file should be generated into the current directory, not
-    # subdir/, where the source is.
-    expected_object_filenames = ('a.spv',)
-
-
-@inside_glslc_testsuite('WorkDir')
 class TestMultipleWorkDir(expect.ValidNamedObjectFile):
     """Tests that if there are multiple -working-directory=<dir> specified,
     only the last one takes effect."""
@@ -123,17 +112,6 @@ class TestWorkDirCompileFile(expect.ValidNamedObjectFile):
 
 
 @inside_glslc_testsuite('WorkDir')
-class TestWorkDirLinkFile(expect.ValidNamedObjectFile):
-    """Tests -working-directory=<dir> when linking input file."""
-
-    environment = Directory('.', [EMPTY_SHADER_IN_SUBDIR])
-    glslc_args = ['-working-directory=subdir', 'shader.vert']
-    # Output file should be generated into the current directory, not
-    # subdir/, where the source is.
-    expected_object_filenames = ('a.spv',)
-
-
-@inside_glslc_testsuite('WorkDir')
 class TestWorkDirCompileFileOutput(expect.ValidNamedObjectFile):
     """Tests -working-directory=<dir> when compiling input file and specifying
     output filename."""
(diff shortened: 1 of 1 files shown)


glslc/test/working_directory.py
```diff
@@ -207,11 +157,63 @@ class TestWorkDirCompileFileAbsolutePath(expect.ValidObjectFile):
...
-class TestWorkDirLinkFileAbsolutePath(expect.ValidNamedObjectFile):
-    """Tests -working-directory=<dir> when linking input file with absolute
-    path."""
+# The -working-directory flag should not affect the placement of the link file.
+# The following tests ensure that.
 
-    shader = FileShader('void main() {}', '.vert')
-    glslc_args = ['-working-directory=subdir', shader]
+class WorkDirDoesntAffectLinkedFile(expect.ValidNamedObjectFile):
+    """A base class for tests asserting that -workind-directory has no impact
```
dneto0: Typo:  "-working-directory".

More importantly, working-directory *does* affect where the output file goes.  We want to match clang behaviour.  (Try it.)
But this is a matter of incomplete feature, rather than this CL.

glslc/test/working_directory.py
```diff
@@ -207,11 +157,63 @@ class TestWorkDirCompileFileAbsolutePath(expect.ValidObjectFile):
...
+        Directory('bin', [])])
+
+
+@inside_glslc_testsuite('WorkDir')
+class TestWorkDirLinkFileDefaultLocation(WorkDirDoesntAffectLinkedFile):
+    """Tests that -working-directory doesn't impact the default link-file
+    location.
+    """
+    glslc_args = ['-working-directory=subdir', 'shader.vert']
     expected_object_filenames = ('a.spv',)
```
dneto0: I think this should be 'subdir/a.spv'?!

antiagainst: For linked final output, `-working-directory` has no effect.

dmircevski: @dneto0 No, see line 160.

antiagainst: Btw, the rules I used (which is from my experience with clang):
https://github.com/google/shaderc/tree/master/glslc#413--working-directory

glslc/test/working_directory.py
```diff
@@ -207,11 +157,63 @@ class TestWorkDirCompileFileAbsolutePath(expect.ValidObjectFile):
...
-    shader = FileShader('void main() {}', '.vert')
-    glslc_args = ['-working-directory=subdir', shader]
+class WorkDirDoesntAffectLinkedFile(expect.ValidNamedObjectFile):
+    """A base class for tests asserting that -workind-directory has no impact
+    on the location of the output link file.
+    """
+    environment = Directory('.', [
+        Directory('subdir', [
+            File('shader.vert', 'void main(){}'),
+            # This bin/ is in the wrong location, so glslc should ignore it.
```
antiagainst: I'd suggest to reword this sentence to something like "This bin/ is not where linked files are generated...". "In the wrong location" sounds strange for me. We are freely to create a bin/ directory anywhere.

glslc/test/working_directory.py
dmircevski: I tried it, and it doesn't:
```
deki@deki1>ls subdir/
x.c
deki@deki1>ls a.out 
ls: cannot access a.out: No such file or directory
deki@deki1>clang-3.5 -working-directory subdir x.c 
deki@deki1>ls subdir/
x.c
deki@deki1>ls a.out 
a.out*
deki@deki1>
```

See also https://github.com/google/shaderc/pull/5/files#diff-1f63fb1c30dcf688e73a16603d4169daR111

antiagainst:
`TestWorkDirEqNoArgLinkFile` got deleted. But I don't find an equivalent one added?

glslc/test/working_directory.py
dneto0: ok, acknowledged I was thinking of a different case.

glslc/test/working_directory.py
dneto0: ack.

dmircevski: @dneto0 Fixed the typo.

dneto0:
+1

glslc/test/working_directory.py
dmircevski: @antiagainst Reworded.

glslc/test/working_directory.py
```diff
@@ -53,17 +53,6 @@ class TestWorkDirEqNoArgCompileFile(expect.ValidNamedObjectFile):
 
 
 @inside_glslc_testsuite('WorkDir')
-class TestWorkDirEqNoArgLinkFile(expect.ValidNamedObjectFile):
```
dneto0: This test has no equivalent in the new version.

dneto0:
-1.  As Lei pointed out, a test has gone missing.

dmircevski:
@antiagainst That test doesn't cover anything new in terms of link-file location.  It merely varies the -working-directory format, which is orthogonal and tested elsewhere.

We don't need to cover the full cross-product of -o and -working-directory variations, and it's weird to cover some combos but not others.

dmircevski:
@antiagainst @dneto0 The equivalent new test is `TestWorkDirLinkFileDefaultLocation`.

dneto0:
+1

antiagainst:
LGTM

Please squash and merge. :)

dmircevski:
Went in squashed as b2420a3b9f78cf0a735523988bb1dda2ecb70dc5.

PR #25, opened 2015-08-25
Fix Windows build.

dmircevski:
Make EXPECT_DEBUG_DEATH_IF_SUPPORTED take a fixed number of arguments,
since MSVC apparently doesn't compile \__VA_ARGS__ the same as gcc and
clang.


dmircevski:
Well, at least this made progress on the Windows build. :)  The build itself succeeds now, but the glslang tests fail.

dmircevski:
Win build fixed by other PRs.

PR #27, opened 2015-08-25
Turn off debug build on Window temporarily

antiagainst:
There is an infinite loop when testing for the debug configuration
on Windows. We don't want to waste AppVeyor resources on this issue
for every commit. Will bring back once the infinite loop is solved.


dmircevski:
+2

PR #28, opened 2015-08-26
Remove Unicode characters from path tests.

antiagainst:
We use C++ `std::string` for handling paths. Library functions based
on `std::string` doesn't play nicely with Unicode.


dmircevski:
+2

Nit: "don't play nicely."

antiagainst:
@dmircevski done

PR #33, opened 2015-09-15
glslc: avoid crash on failure to open the output file stream

dneto0:
Also, update the glslc test framework so that any test will fail if the subprocess was terminated by a signal.  This is especially important to distinguish for a test where the subprocess is expected to exist with non-zero status code.


AWoloszyn:
+2

PR #34, opened 2015-09-16
Fixes for build in VS2013.

AWoloszyn:
Use initializer in atomic_int constructor, since VS2013 does not support
the constexpr constructor.

Change definition of EXPECT_DEBUG_DEATH_IF_SUPPORTED to match how
gtest does it, in order to avoid inconsistent handling of __VA_ARGS__.

Changed include_file.1 to not have a new-line at the end, because
this may get checked out differently on different platforms.


libshaderc_util/include/libshaderc_util/counting_includer.h
```diff
@@ -24,7 +24,7 @@ namespace shaderc_util {
 // An Includer that counts how many #include directives it saw.
 class CountingIncluder : public glslang::TShader::Includer {
  public:
-  CountingIncluder() : num_include_directives_(0) {}
+  CountingIncluder() : num_include_directives_({0}) {}
```
dneto0: This is fine because this atomic_int object is *created* by this constructor.  No race is possible until a client can see this object, which occurs after construction.
At first I thought would have to use in-class member initialization with ATOMIC_VAR_INIT macro call, but we don't need to go that far.

dneto0:
+2

PR #35, opened 2015-09-21
Catch up with glslang development

antiagainst:
Glslang has been migrated to use the Rev. 31 `spirv.hpp` header.
`GLSL450Lib.h` was replaced by `GLSL.std.450.h`.


AWoloszyn:
+2! Goodbye GlslStd450DebugNames!

PR #38, opened 2015-10-09
Travis CI: test correct CC for uploading coverage report.

antiagainst:
We set `CC` to `gcc-4.8` previously; so checking whether `CC` is `gcc`
doesn't work any more.


dneto0:
+2

dmircevski:
+2

PR #39, opened 2015-10-15
Add missing shader kinds

mgadams:
Currently only vertex and fragment shader types are exposed; this adds
mappings for the remainder

### DIFF
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -261,4 +261,77 @@ TEST(CompileString, MultipleThreadsCalling) {
   EXPECT_THAT(results, Each(true));
 }
 
+TEST(CompileKinds, Vertex) {
+  Compiler compiler;
+  ASSERT_NE(nullptr, compiler.get_compiler_handle());
+  const std::string kVertexShader = "void main(){ gl_Position = vec4(0);}";
+  EXPECT_TRUE(CompilationSuccess(compiler.get_compiler_handle(), kVertexShader,
+                                 shaderc_glsl_vertex_shader));
+}
+
+TEST(CompileKinds, Fragment) {
+  Compiler compiler;
+  ASSERT_NE(nullptr, compiler.get_compiler_handle());
+  const std::string kFragShader = "void main(){ gl_FragColor = vec4(0);}";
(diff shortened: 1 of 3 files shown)


dmircevski:
Thanks, @mgadams !

Would it be possible to also put in a test or two?  Say, in libshaderc/src/shaderc*test.cc.

mgadams:
I’ve been added to the NVIDIA Corporation CLA, so contributions should be covered now.

mgadams:
Yes, I can take a look into adding some basic tests. It'll likely be a few days before I can scare up the time though.

mgadams:
I've updated the commit to include a new set of tests which test a simple shader of each type can be compiled. I put them in a new test group so it's easy to tell from the output what shader types are currently enabled & tested.

libshaderc/src/shaderc_test.cc
```diff
@@ -261,4 +261,74 @@ TEST(CompileString, MultipleThreadsCalling) {
...
+      "#version 310 es\n"
+      "void main(){}\n";
+  EXPECT_TRUE(CompilationSuccess(compiler.get_compiler_handle(), kCompShader,
+                                 shaderc_glsl_compute_shader));
+}
+
+TEST(CompileKinds, Geometry) {
+  Compiler compiler;
+  ASSERT_NE(nullptr, compiler.get_compiler_handle());
+  const std::string kGeoShader =
```
antiagainst: I'd suggest to use C++11 raw strings to improve readability for these code snippets. E.g.,

```
const std::string kGeoShader = R"(
  #version 310 es
  #extension ...
  ...
)";
```

mgadams: That's a new syntax for me, but sure. Note that due to the GLES requirement of no characters prior to the #version token, the R"( had to go on the same line as the first shader line.

antiagainst: Yeah, you are right. Sorry, neglected that point earlier. :)

dmircevski:
LGTM, merged.  @mgadams: can you please add yourself to the CONTRIBUTORS file? :)

mgadams:
Thanks! And yeah, I didn't think about the CONTRIBUTORS file. I have another patch to clean up & pass your way soon; I'll edit CONTRIBUTORS as part of that commit.

PR #41, opened 2015-10-22
Add shaderc_get_spv_version() to retrieve SPIR-V version/revision

mgadams:
This is useful for diagnostic purposes when embedded in test apps.

### DIFF
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -21,6 +21,8 @@
 #include <sstream>
 #include <vector>
 
+#include "SPIRV/spirv.hpp"
+
 #include "libshaderc_util/compiler.h"
 #include "libshaderc_util/resources.h"
 
@@ -204,4 +206,9 @@ void shaderc_module_release(shaderc_spv_module_t module) { delete module; }
 const char* shaderc_module_get_error_message(
     const shaderc_spv_module_t module) {
   return module->messages.c_str();
+}
+
+void shaderc_get_spv_version(unsigned int *version, unsigned int *revision) {
+    *version = spv::Version;
+    *revision = spv::Revision;
 }
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -52,6 +52,14 @@ TEST(Init, MultipleThreadsCalling) {
   shaderc_compiler_release(compiler3);
 }
 
+TEST(Init, SPVVersion) {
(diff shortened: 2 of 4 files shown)


libshaderc/src/shaderc_test.cc
```diff
@@ -52,6 +52,14 @@ TEST(Init, MultipleThreadsCalling) {
   shaderc_compiler_release(compiler3);
 }
 
+TEST(Init, SPVVersion) {
+  unsigned int version = 0;
+  unsigned int revision = 0;
+  shaderc_get_spv_version(&version, &revision);
+  EXPECT_TRUE(version == spv::Version);
```
dmircevski: Use EXPECT_EQ or EXPECT_THAT(version, Eq(spv::Version)).

Same below.

libshaderc/src/shaderc.cc
```diff
@@ -23,6 +23,7 @@
 
 #include "libshaderc_util/compiler.h"
 #include "libshaderc_util/resources.h"
+#include "SPIRV/spirv.hpp"
```
dmircevski: Our style requires this be moved up, as "other libraries' .h file":

http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Names_and_Order_of_Includes

mgadams:
Both updates done as requested.

libshaderc/src/shaderc_test.cc
```diff
@@ -52,6 +52,14 @@ TEST(Init, MultipleThreadsCalling) {
   shaderc_compiler_release(compiler3);
 }
 
+TEST(Init, SPVVersion) {
+  unsigned int version = 0;
+  unsigned int revision = 0;
+  shaderc_get_spv_version(&version, &revision);
+  EXPECT_EQ(version, spv::Version);
```
dmircevski: Looks like you got the arguments swapped -- spv::Version is the expected, version the actual.

mgadams: Ah, I didn't realize order mattered (never used this test framework before). I'll send another update later this morning.

dmircevski: Yeah, it matters for correct failure messages -- it's pretty nifty when a failing test says something like
```
Value of: version
  Actual: 456
Expected: spv::Version
Which is: 123
```

mgadams: Definitely a useful feature. I've updated the commit accordingly.

PR #42, opened 2015-10-26
Ensure prototypes don't have empty parameter lists

mgadams:
When built in environments with warnings enable, newer versions of
GCC complain about two of the function definitions not being prototypes. This
is solved by using (void) instead of ()


antiagainst:
LGTM, merged. Thanks, @mgadams!

PR #44, opened 2015-10-27
DRY setup_build.cmake's use of find_host_program.

dmircevski:


dmircevski:
Hm, the modified find_program invocation isn't working quite right, so continuous builds failed.  Please hold off on reviewing this until I figure it out...

dneto0:
I suspect beta reduction doesn't work in CMake.  ;-P
And there's a tension between DRY and simplicity here.

dmircevski:
Nah, variable substitution works fine.  Now ready for review.  Hopefully this is still simple enough to grok.

dneto0:
+2

dneto0:
Merged

dneto0:
Merged after squashing.

PR #45, opened 2015-10-30
Set Vulkan and SPIR-V rules bits when invoking glslang

jekstrand:
Without these, it assumes some GL-based set of rules and doesn't enable
everything.  Some day, it might be good if these came from command-line
flags or a #extension declaration or something.

### DIFF
--- a/libshaderc_util/src/compiler.cc
+++ b/libshaderc_util/src/compiler.cc
@@ -148,7 +148,9 @@ bool Compiler::Compile(
   bool success =
       shader.parse(&shaderc_util::kDefaultTBuiltInResource, default_version_,
                    default_profile_, force_version_profile_,
-                   kNotForwardCompatible, EShMsgDefault, includer);
+                   kNotForwardCompatible,
+                   static_cast<EShMessages>(EShMsgSpvRules | EShMsgVulkanRules),
+                   includer);
 
   success &= PrintFilteredErrors(error_tag, error_stream, warnings_as_errors_,
                                  suppress_warnings_, shader.getInfoLog(),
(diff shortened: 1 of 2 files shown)


dmircevski:
Thanks, Jason!  Once you take care of the CLA, could you please also add yourself to the CONTRIBUTORS file in the root directory?

libshaderc_util/src/compiler.cc
```diff
@@ -148,7 +148,8 @@ bool Compiler::Compile(
   bool success =
       shader.parse(&shaderc_util::kDefaultTBuiltInResource, default_version_,
                    default_profile_, force_version_profile_,
-                   kNotForwardCompatible, EShMsgDefault, includer);
+                   kNotForwardCompatible,
+                   (EShMessages)(EShMsgSpvRules | EShMsgVulkanRules), includer);
```
dmircevski: Please use C++-style casts.

https://google.github.io/styleguide/cppguide.html#Casting

dmircevski:
Code LGTM.  @googlebot: "CLAs are signed, but unable to verify author consent"??  @jekstrand: perhaps you need to reply here saying something like "I signed it."  I'm not sure how exactly the process works.

dneto0:
Jason, thanks for the patch.
Currently Shaderc isn't very flexible about setting options to further describe the target, including dialect and things like target specific limits.  I've got some ideas in that direction, but thanks for moving the ball forward with this pull request.

jekstrand:
Unfortunately, the CLA brought up some Intel management approval issues.  I'm in the process of getting those sorted out which is why this is taking so long.  I'll formally acknowledge the CLA as soon as I get things sorted out internally.

dneto0:
No problem about the delay.  Completely understand.

jekstrand:
I think all the legal road-blocks are taken care of.

I should now be under the Intel-signed non-Android Google NDA.

dmircevski:
@googlebot: can you please confirm that Jason's CLA is good?  I'd like to merge this PR soon.

antiagainst:
Squashed and merged as 68b4a047ea70e57294ed66026f059219129a18cc. Thanks, @jekstrand!

antiagainst:
closing since already merged.

PR #46, opened 2015-11-16
Update SPIR-V binary version and generator number.

antiagainst:
This is required since glslang has updated to support SPIR-V v1r2.


dneto0:
+2

Upstream glslang was updated to support SPIR-V 1.0.  Also, glslang registered tool ID 8, and is using misc half-word of 1.  So those values are right.

PR #48, opened 2015-12-01
Fixed android ndk-build from change to paths in glslang.

AWoloszyn:
OSDependent moved into Unix and Windows rather than
Linux and Windows.

It also requires that -DGLSLANG_OSINCLUDE_UNIX is specified
for Unix builds.


dmircevski:
+2

antiagainst:
+2 if checks pass.

PR #49, opened 2015-12-01
Suppress all warnings from third-party projects.

antiagainst:


third_party/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+# Suppress all warnings from third-party projects.
+set_property(DIRECTORY APPEND PROPERTY COMPILE_OPTIONS -w)
```
dmircevski: Does this work for all compilers?

antiagainst: Works on GCC and Clang. "[When compiling with MSVC the “-w” will be automatically converted to “/w” which has the same function.](https://www.johnlamp.net/cmake-tutorial-4-libraries-and-subdirectories.html)"

dmircevski: Perfect, thx!

dmircevski:
+2

PR #51, opened 2015-12-01
Fix cmake on mac

Whiteseeker:
Replace "ar -M" on mac to be able to merge static libraries.
It now use "libtool" which is also available on linux, maybe we could remove the if later once the libtool method is tested on linux, in the meantime I left the linux code untouched.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -127,7 +127,6 @@ function(combine_static_lib new_target target)
   get_transitive_libs(${target} all_libs)
   
   if(APPLE)
-
     string(REPLACE ";" " " all_libs_string "${all_libs}")
 
     add_custom_command(OUTPUT lib${new_target}.a
@@ -136,7 +135,6 @@ function(combine_static_lib new_target target)
     add_custom_target(${new_target}_genfile ALL
       DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/lib${new_target}.a)
   else()
-
     string(REPLACE ";" "\\\\naddlib " all_libs_string "${all_libs}")
 
     set(ar_script
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -123,11 +123,11 @@ function(combine_static_lib new_target target)
     message(FATAL_ERROR "MSVC does not yet support merging static libraries")
   endif()
 
-  # Should remove the if and use libtool on MacOS and Linux
+  set(all_libs "")
+  get_transitive_libs(${target} all_libs)
+  
   if(APPLE)
 
-    set(all_libs "")
-    get_transitive_libs(${target} all_libs)
     string(REPLACE ";" " " all_libs_string "${all_libs}")
 
     add_custom_command(OUTPUT lib${new_target}.a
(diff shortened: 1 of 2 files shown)


cmake/utils.cmake
```diff
@@ -123,18 +123,34 @@ function(combine_static_lib new_target target)
...
-  set(all_libs "")
-  get_transitive_libs(${target} all_libs)
-  string(REPLACE ";" "\\\\naddlib " all_libs_string "${all_libs}")
+  # Should remove the if and use libtool on MacOS and Linux
+  if(APPLE)
 
-  set(ar_script
-    "create lib${new_target}.a\\\\naddlib ${all_libs_string}\\\\nsave\\\\nend")
+    set(all_libs "")
+    get_transitive_libs(${target} all_libs)
```
dmircevski: Aren't the above two lines identical to 141 & 142?  If so, this needs DRYing.

cmake/utils.cmake
```diff
@@ -123,18 +123,34 @@ function(combine_static_lib new_target target)
     message(FATAL_ERROR "MSVC does not yet support merging static libraries")
   endif()
 
-  set(all_libs "")
-  get_transitive_libs(${target} all_libs)
-  string(REPLACE ";" "\\\\naddlib " all_libs_string "${all_libs}")
+  # Should remove the if and use libtool on MacOS and Linux
```
dmircevski: I kinda doubt libtool and cmake play well together.  Also, didn't libtool use to require only libraries that it has built itself?

Whiteseeker: You may be write, it's the first time I hear about it so I'm unsure.

cmake/utils.cmake
Whiteseeker: Yes they are. Will fix

cmake/utils.cmake
dmircevski: I'm just wondering if this comment is appropriate, since it's not clear to me we want to necessarily go in that direction.  Perhaps we should leave it out?

Whiteseeker: Sure I'll remove it too.

Whiteseeker:
Both comment should be fixed. Welcome any other (^_^). I still wonder why the test 14/14 fail on mac.

cmake/utils.cmake
```diff
@@ -125,16 +125,29 @@ function(combine_static_lib new_target target)
...
+
+    string(REPLACE ";" "\\\\naddlib " all_libs_string "${all_libs}")
+
+    set(ar_script
+      "create lib${new_target}.a\\\\naddlib ${all_libs_string}\\\\nsave\\\\nend")
+
+    add_custom_command(OUTPUT lib${new_target}.a
+      DEPENDS ${all_libs}
+      COMMAND ${ECHO_EXE} -e ${ar_script} | ${CMAKE_AR} -M)
+    add_custom_target(${new_target}_genfile ALL
```
antiagainst: I think the `add_custom_target(...)` command can also be DRY'd?

cmake/utils.cmake
```diff
@@ -125,16 +125,29 @@ function(combine_static_lib new_target target)
 
   set(all_libs "")
   get_transitive_libs(${target} all_libs)
-  string(REPLACE ";" "\\\\naddlib " all_libs_string "${all_libs}")
+  
+  if(APPLE)
```
antiagainst: Unnecessary empty line after `if(...)` and `else()`. :)

antiagainst:
@Whiteseeker: I've tried to build on Mac. `glslc_tests` passes after I installed `nose` via `pip`. You may want to check whether it also solves your problem.

antiagainst:
@Whiteseeker: It turns out there was a bug in `glslc_test_framework.py`. #53 should have it fixed. :)

Whiteseeker:
Actually installed nose did the trick and test 14 pass properly on mac too.
Thx !

antiagainst:
LGTM. I'll squash and merge this.

antiagainst:
squashed and committed as 4f86339d10beb2220970554cec56728be13d8239. closing.

PR #52, opened 2015-12-01
Git should ignore third_party/googletest

dneto0:


.gitignore
```diff
@@ -5,4 +5,5 @@ compile_commands.json
 .ycm_extra_conf.py
 cscope.*
 third_party/glslang/
+third_party/googletest/
 third_party/gmock-1.7.0/
```
dmircevski: Is this line still necessary?

antiagainst: +1. The old gtest dir should be deleted.

dneto0: I intentionally left it in for the transition period where people will have both trees.  But I can see it's misleading.  Will remove it.

antiagainst:
LGTM.

dneto0:
rebased and pushed onto master

PR #53, opened 2015-12-02
Skip scanning Python files for nose tests in glslc test framework.

antiagainst:
Those files contain tests for the test framework; they should be
added separately.


antiagainst:
This solves an import failure when `nose` is not found.

Reported in #51.

dmircevski:
+2

antiagainst:
@dmircevski: The previous commit has a defect. Pushed out a new commit for fixing it.

dmircevski:
+2

Issue #40, opened 2015-10-16
Add command-line option to select SPIR-V generation mode (Vulkan or OpenGL)


antiagainst:
glslang [has two command-line options](https://github.com/KhronosGroup/glslang/blob/master/StandAlone/StandAlone.cpp#L954-L958) to generate SPIR-V:
* `-V`: under Vulkan semantics
* `-G`: under OpenGL semantics

shaderc should also support some kind of command-line option to select the mode.

dneto0:
We already have "-x glsl".  Normally language dialect is chosen with "-std=<name>"  but shaderc already has "-std=310es" for example.  Let's keep this open until Vulkan ships then we see how it expresses version numbers.

antiagainst:
Already done by [`--target-env=`](https://github.com/google/shaderc/tree/master/glslc#423---target-env).

capitalknew:
But anyways not possible to define version of SPIR-V (1.0, 1.1, 1.2).

PR #54, opened 2015-12-08
Add shaderc_compile_options_set_message_rules API

mgadams:
This provides an API for specifying which set of rules should be used for warning/error generation. The default remains spirv/vulkan, but this allows using more relaxed rules for legacy or non-Vulkan shaders.

This was previously possible until commit 68b4a047 when the default rules used by shaderc were changed.

### DIFF
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -21,6 +21,7 @@ extern "C" {
 
 #include <stddef.h>
 #include <stdbool.h>
+#include <stdint.h>
 
 typedef enum {
   shaderc_glsl_vertex_shader,
@@ -30,6 +31,13 @@ typedef enum {
   shaderc_glsl_tess_control_shader,
   shaderc_glsl_tess_evaluation_shader
 } shaderc_shader_kind;
+
+typedef enum {
+  shaderc_target_env_vulkan,      // create SPIR-V under Vulkan semantics
+  shaderc_target_env_glsl,        // create SPIR-V under OpenGL semantics
+  shaderc_target_env_glsl_compat, // create SPIR-V under OpenGL semantics, including compatibility profile functions
+  shaderc_target_env_default = shaderc_target_env_vulkan
+} shaderc_target_env;
 
 // Usage examples:
 //
@@ -104,6 +112,12 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 // deleted after this function has returned.
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
+
+// Set the target shader environment, affecting which warnings or errors will be issued.
+// The version will be for distinguishing between different versions of the target environment.
+// "0" is the only supported version at this point
+void shaderc_compile_options_set_target_env(
+    shaderc_compile_options_t options, shaderc_target_env target, uint32_t version);
 
 // An opaque handle to the results of a call to shaderc_compile_into_spv().
 typedef struct shaderc_spv_module* shaderc_spv_module_t;
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -33,9 +33,9 @@ typedef enum {
 } shaderc_shader_kind;
 
 typedef enum {
-  shaderc_target_env_vulkan,      // create SPIR-V under Vulkan semantics
-  shaderc_target_env_glsl,        // create SPIR-V under OpenGL semantics
-  shaderc_target_env_glsl_compat, // create SPIR-V under OpenGL semantics, including compatibility profile functions
+  shaderc_target_env_vulkan,        // create SPIR-V under Vulkan semantics
+  shaderc_target_env_opengl,        // create SPIR-V under OpenGL semantics
+  shaderc_target_env_opengl_compat, // create SPIR-V under OpenGL semantics, including compatibility profile functions
   shaderc_target_env_default = shaderc_target_env_vulkan
 } shaderc_target_env;
 
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -232,6 +232,36 @@ TEST(CompileStringWithOptions, IfDefCompileOption) {
                                  shaderc_glsl_vertex_shader, options.get()));
 }
 
+TEST(CompileStringWithOptions, TargetEnv) {
+  Compiler compiler;
+  compile_options_ptr options(shaderc_compile_options_initialize());
+  ASSERT_NE(nullptr, compiler.get_compiler_handle());
+
+  // Confirm that this shader compiles with shaderc_target_env_glsl_compat;
+  // if targeting Vulkan, glslang will fail to compile it
+  const std::string kGlslShader =
+    R"(#version 100
+       uniform highp sampler2D tex;
+       void main() {
+         gl_FragColor = texture2D(tex, vec2(0.0,0.0));
+       }
+  )";
+
+  EXPECT_FALSE(CompilesToValidSpv(compiler.get_compiler_handle(),
+                                  kGlslShader,
+                                  shaderc_glsl_fragment_shader, options.get()));
+
+  shaderc_compile_options_set_target_env(options.get(), shaderc_target_env_glsl_compat, 0);
+  EXPECT_TRUE(CompilesToValidSpv(compiler.get_compiler_handle(),
+                                 kGlslShader,
+                                 shaderc_glsl_fragment_shader, options.get()));
+
+  shaderc_compile_options_set_target_env(options.get(), shaderc_target_env_vulkan, 0);
+  EXPECT_FALSE(CompilesToValidSpv(compiler.get_compiler_handle(),
+                                  kGlslShader,
+                                  shaderc_glsl_fragment_shader, options.get()));
+}
+
 TEST(CompileString, ShaderKindRespected) {
   Compiler compiler;
   ASSERT_NE(nullptr, compiler.get_compiler_handle());
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -237,7 +237,7 @@ TEST(CompileStringWithOptions, TargetEnv) {
   compile_options_ptr options(shaderc_compile_options_initialize());
   ASSERT_NE(nullptr, compiler.get_compiler_handle());
 
-  // Confirm that this shader compiles with shaderc_target_env_glsl_compat;
+  // Confirm that this shader compiles with shaderc_target_env_opengl_compat;
   // if targeting Vulkan, glslang will fail to compile it
   const std::string kGlslShader =
     R"(#version 100
@@ -251,7 +251,7 @@ TEST(CompileStringWithOptions, TargetEnv) {
                                   kGlslShader,
                                   shaderc_glsl_fragment_shader, options.get()));
 
-  shaderc_compile_options_set_target_env(options.get(), shaderc_target_env_glsl_compat, 0);
+  shaderc_compile_options_set_target_env(options.get(), shaderc_target_env_opengl_compat, 0);
   EXPECT_TRUE(CompilesToValidSpv(compiler.get_compiler_handle(),
                                  kGlslShader,
                                  shaderc_glsl_fragment_shader, options.get()));
--- a/libshaderc/include/shaderc.hpp
+++ b/libshaderc/include/shaderc.hpp
@@ -104,6 +104,13 @@ class CompileOptions {
     AddMacroDefinition(name.c_str(), value.c_str());
   }
 
+  // Set the target shader environment, affecting which warnings or errors will be issued.
+  // The version will be for distinguishing between different versions of the target environment.
+  // "0" is the only supported version at this point
+  void SetTargetEnvironment(shaderc_target_env target, uint32_t version) {
+    shaderc_compile_options_set_target_env(options_, target, version);
+  }
+
  private:
   CompileOptions& operator=(const CompileOptions& other) = delete;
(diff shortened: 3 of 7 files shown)


libshaderc/include/shaderc.h
```diff
@@ -105,6 +111,11 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
 
+// Set flags for the rules on which warnings or errors will be issued. This can
+// either be an individual flag, or a set of flags or'd together
+void shaderc_compile_options_set_message_rules(
+    shaderc_compile_options_t options, shaderc_msg_rules rules);
+
 // An opaque handle to the results of a call to shaderc_compile_into_spv().
 typedef struct shaderc_spv_module* shaderc_spv_module_t;
```
dmircevski: What about `shaderc.hpp`?  We strive to have feature parity in the C++ API.

libshaderc/include/shaderc.h
```diff
@@ -105,6 +111,11 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
 
+// Set flags for the rules on which warnings or errors will be issued. This can
+// either be an individual flag, or a set of flags or'd together
```
dmircevski: If `rules` can be a combo of enum values, then its declared type can't be the enum.

libshaderc/include/shaderc.h
mgadams: I actually wasn't aware of shaderc.hpp; I'll take a look at that now.

libshaderc/include/shaderc.h
```diff
@@ -105,6 +111,11 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
...
+void shaderc_compile_options_set_message_rules(
+    shaderc_compile_options_t options, shaderc_msg_rules rules);
+
```
dmircevski: Would it make sense to have a getter, too?

libshaderc/include/shaderc.h
mgadams: That's actually following the precise pattern used by glslang and EShMessages which this enum maps to. See glslang/Public/ShaderLang.h line 127 and then subsequent uses. I do agree it's unusual, but my thinking was that it'd be best to keep it close to the underlying options we're exposing from glslang.

libshaderc/include/shaderc.h
mgadams: No strong opinion on this. We don't have a getter for shaderc_compile_options_add_macro_definition(), so I followed the same approach. I don't have a problem with adding one if that's desired.

libshaderc/src/shaderc_test.cc
```diff
@@ -232,6 +232,42 @@ TEST(CompileStringWithOptions, IfDefCompileOption) {
...
+                                 kGlslShader,
+                                 shaderc_glsl_fragment_shader, options.get()));
+
+  shaderc_compile_options_set_message_rules(options.get(), shaderc_msg_spirv_rules);
+  EXPECT_FALSE(CompilesToValidSpv(compiler.get_compiler_handle(),
+                                  kGlslShader,
+                                  shaderc_glsl_fragment_shader, options.get()));
+
+  shaderc_compile_options_set_message_rules(options.get(),
+                                            static_cast<shaderc_msg_rules>(shaderc_msg_spirv_rules | shaderc_msg_vulkan_rules));
```
dmircevski: I don't like this `static_cast`, because the value it produces isn't a member of the enum.

libshaderc/include/shaderc.h
AWoloszyn: I think it's probably fine to not have getters for the options I don't think they will actually provide too much use.

libshaderc/src/shaderc_test.cc
mgadams: As above, this again stems from glslang and EShMessages. While I could solve the immediate problem by adding a third member to the enum, my concern is making the API future-proof. There are other glslang EShMessages options we may choose to expose in the future. Alternatively, I could see glslang adding an option down the road to generate different versions of SPIR-V for example.

I see two main alternatives here:

1) change the shaderc_msg_* values to #defines,  and pass in an unsigned int for the parameter. That's less clean, but makes it easier to expose additional glslang options in the future

2) change the enum values to entries to shaderc_msg_default, shaderc_msg_spirv_glsl_rules and shaderc_msg_spirv_vulkan_rules. This is cleaner and fulfills the immediate need, but doesn't leave an easy avenue for exposing additional glslang options without adding additional API entrypoints

Since you're looking to break from the underlying glslang/EShMessages pattern, which approach do you prefer?

libshaderc/include/shaderc.h
dmircevski: I'm not sure Shaderc should follow that particular pattern.  Why don't we make this parameter an unsigned int, which can happily take a bitmask of `shaderc_msg_rules`?

libshaderc/include/shaderc.h
dmircevski: Ah, fair point.

libshaderc/src/shaderc_test.cc
dmircevski: I agree we want to stay with a bitmask for keeping the API future-proof.  But if the parameter type were an integer, you could just drop the `static_cast` here and keep everything else the same, no?

libshaderc/include/shaderc.h
mgadams: I've add SetMessageRules() to the C++ interface and a test for that.

libshaderc/src/shaderc_test.cc
mgadams: Agreed; I've gone the route of using an unsigned int parameter.

libshaderc/include/shaderc.h
```diff
@@ -31,6 +31,12 @@ typedef enum {
   shaderc_glsl_tess_evaluation_shader
 } shaderc_shader_kind;
 
+typedef enum {
+  shaderc_msg_default           = 0x00,
+  shaderc_msg_spirv_rules       = 0x01,
```
dneto0: This presents the wrong abstraction.
What you have for "shaderc_msg_spirv_rules" is "generate SPIR-V using OpenGL conventions".   The default has always been to target OpenGL conventions, but glslang needed the extra enum to control whether generate SPIR-V.

See in glslang's Standalone.cpp, function SetMessageOptions:
https://github.com/KhronosGroup/glslang/blob/master/StandAlone/StandAlone.cpp#L601
It says "Translate the meaningful subset of command-line options to parser-behaviour options."

Look further inside glslang and you see that it controls what set of builtin functions are available.  

So it's really about selecting dialect depending on target environment: either OpenGL or Vulkan. These are mutually exclusive.

So I'd rather have something like this:

```
// Selects which target environment conventions to follow in generated SPIR-V.
// Each target environment such as Vulkan or OpenGL should define the conventions
// expected of a SPIR-V module.
enum {
shaderc_target_env_vulkan
shaderc_target_env_opengl,
shaderc_target_env_default = shader_target_env_vulkan
} shaderc_target_env;
```

Eventually this would also be exposed on the glslc command line as --target-env=<something> where that <something> is either vulkan or opengl.  (Default is Vulkan.)

libshaderc/include/shaderc.h
```diff
@@ -105,6 +111,11 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
 
+// Set flags for the rules on which warnings or errors will be issued. This can
+// either be an individual shaderc_msg_rules flag, or a set of flags or'd together
+void shaderc_compile_options_set_message_rules(
```
dneto0: Correspondingly, should be 
```
// Sets the target environment and version, overriding the default or any previous call to this
// method. Default is shaderc_target_env_vulkan.
void shader_compile_options_set_target_env(..., shaderc_target_env);
```

The target can be an enum, since they are exclusive choices.
(If it were a bit mask, then we'd want an unsigned int to convey it.)

I'm waffling over whether we should also have a version parameter to state which version of the target spec sets the rules.  So far no OpenGL spec sets rules for SPIR-V, but has an established numbering convention such as 450 for OpenGL 4.5.   Vulkan is still under development, but later versions might have different rules than what goes into Vulkan 1.0.   For now I think we can address any future movement in the rules with a separate shader_compile_options_set_target_env_version(..., int version);  if we need it.

dneto0: Looking into the future once more, if we have two different calls to set target env and target env spec version, then we need to define when validation occurs. Since you can't set both at once, you can only validate at compile time, or we could add a "validate" method on the compiler options to check all settings for mutual consistency.

libshaderc/src/shaderc_test.cc
```diff
@@ -232,6 +232,37 @@ TEST(CompileStringWithOptions, IfDefCompileOption) {
...
 }
 
+TEST(CompileStringWithOptions, MessageRules) {
+  Compiler compiler;
+  compile_options_ptr options(shaderc_compile_options_initialize());
+  ASSERT_NE(nullptr, compiler.get_compiler_handle());
+
+  // Confirm that this shader only compiles with shaderc_msg_default;
+  // if using SPIR-V rules, glslang will fail to compile it
+  const std::string kGlslShader =
```
dneto0: Thanks for the test.  It was really illuminating.  It made me think hard about what the option was really doing.

dneto0:
-1  since I think it's presenting the wrong abstraction.

libshaderc/include/shaderc.h
mgadams: It's actually not quite that cut & dried, as glslang has a couple of different OpenGL modes. Currently, if you just pass in EShMsgDefault, it permits using compatibility builtins (e.g. ES2 shaders).

If you pass in EShMsgSpvRules, it uses a more restrictive set of builtins which correspond to more modern GL profiles. This is equivalent to running glslang with -G.

If you pass in EShMsgSpvRules | EShMsgVulkanRules, you can only use the subset which is valid for Vulkan. This is equivalent to running glslang with -V.

If the goal here is to expose target profiles, then we'd need to have 3 options in the enum above. Note that this is a very different goal from what I was aiming to achieve; however, I don't have a fundamental objection if that's the direction the Shaderc maintainers want to take instead.

libshaderc/include/shaderc.h
mgadams: Two other alternatives here:

1) add an "int version" parameter now to the target env function and document it as unused for now; that at least gets the API in place and avoids having to change it in the future.

2) down the road, add a second set_target_env variant which has the "int version" parameter; document the original set_target_env version to use the highest supported version.

Either of those would avoid having separate target & version functions and the "when to validate" issue. I don't have a strong preference offhand.

dneto0: I think you're right: better to add the int version parameter now.  Make it a uint32_t, since version numbers shouldn't be negative (and despite the Google C++ style guide).  (Entails adding #include <stdint.h>)
Each target env (external) spec determines what the numbers mean.

libshaderc/include/shaderc.h
antiagainst: Issue #40 is related to this. (Just want to create a link :D)

mgadams:
Thanks for the feedback; I've updated the pull request based on the recommendations.

libshaderc/include/shaderc.h
```diff
@@ -105,6 +113,12 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
 
+// Set the target shader environment, affecting which warnings or errors will be issued.
+// The version is currently unused, but will be for distinguishing between different
```
dneto0: If we say version is unused, then people will put random values there.  Better to say only 0 is supported for now.

libshaderc/include/shaderc.h
```diff
@@ -31,6 +32,13 @@ typedef enum {
   shaderc_glsl_tess_evaluation_shader
 } shaderc_shader_kind;
 
+typedef enum {
+  shaderc_target_env_vulkan,
+  shaderc_target_env_glsl,
+  shaderc_target_env_glsl_compat,
```
dneto0: Should say how glsl_compat differs from glsl.   I'm confused, and would have to dig into the glslang code to see the differences again.

libshaderc/include/shaderc.h
mgadams: Good point; I've updated the documentation accordingly in both header files.

libshaderc/include/shaderc.h
mgadams: I've added documentation for each target environment here.

In a nutshell - the compat mode allows use of compatibility profile built-in functions (e.g. texture2D). To see the list, grep for "spv == 0" in glslang/MachineIndependent/Initialize.cpp.

I agree it's not clearly defined, but IMO this is still useful functionality to expose. This matches what Shaderc accepted prior to commit 68b4a04.

mgadams:
Both updates done & ready for review.

Regarding the message_rules_ member variable - you're quite correct about the tension there. I had debated with myself originally whether to just pass through the bitmask, or to add an internal enum for passing down the target environment. I decided to go the bitmask route to avoid a second type conversion, but I agree it's not as clean semantically as most of the other members in that class.

dneto0:
+2.  Waiting for Travis results before merging into master.

dneto0:
I think Travis is stuck.
Squashed and pushed into master.
Closing.

PR #55, opened 2015-12-22
Add set_preprocessing_only_mode() to libshaderc.h API

Qining:
Add set_preprocessing_only_mode() to libshaderc C API (shaderc.h, shaderc.cc).

Add test for set_preprocessing_only_mode() in shaderc_test.cc

### DIFF
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -165,7 +165,7 @@ size_t shaderc_module_get_length(const shaderc_spv_module_t module);
 // Returns a pointer to the start of the SPIR-V bytes, either SPIR-V binary or
 // char string. When the source string is compiled into SPIR-V binary, this is
 // guaranteed to be castable to a uint32_t*. If the source string is compiled in
-// disassembly mode or preprocessing only mode, the pointer will points to the
+// disassembly mode or preprocessing only mode, the pointer will point to the
 // result char string.
 const char* shaderc_module_get_bytes(const shaderc_spv_module_t module);
 
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -115,7 +115,7 @@ void shaderc_compile_options_add_macro_definition(
 
 // Set the compiler to do only preprocessing. The byte array result in the
 // module returned by the compilation is the text of the preprocessed shader.
-// This option override all other compilation modes, such as disassembly mode
+// This option overrides all other compilation modes, such as disassembly mode
 // and the default mode of compilation to SPIR-V binary.
 void shaderc_compile_options_set_preprocessing_only_mode(
     shaderc_compile_options_t options);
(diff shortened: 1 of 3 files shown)


libshaderc/include/shaderc.h
```diff
@@ -113,6 +113,13 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
 
+// Set the compiler to only do preprocessing. When both preprocessing only mode
+// and disassembly mode are turned on, the compiler will respect preprocessing
+// only
```
dneto0: Bad line break in comment.
Preprocessing-only mode also prevents SPIR-V code generation.  You should generalize the comment that way.
E.g.
"
Set the compiler to do only preprocessing.  The byte array result in the module returned by the compilation method is the text of the preprocessed shader."

Also, you now must update the descriptions on the methods: shaderc_compile_into_spv, shaderc_module_get_bytes, shaderc_module_get_length.  They all say the result is a SPIR-V binary. But that's no longer true.

dneto0:
-1.  Need to fix the method comments.
Also, a followup CL should update the shaderc C++ API.

libshaderc/include/shaderc.h
dneto0: And yes, it's true the preprocessing-only mode overrides all other compilation modes (disassembly, and the default mode of compilation to SPIR-V binary).

libshaderc/include/shaderc.h
```diff
@@ -113,6 +113,13 @@ void shaderc_compile_options_release(shaderc_compile_options_t options);
 void shaderc_compile_options_add_macro_definition(
     shaderc_compile_options_t options, const char* name, const char* value);
 
+// Set the compiler to do only preprocessing. The byte array result in the
+// module returned by the compilation is the text of the preprocessed shader.
+// This option override all other compilation modes, such as disassembly mode
```
dneto0: "overrides" instead of "override".  I can't explain English grammar, sorry.  :-)

libshaderc/include/shaderc.h
```diff
@@ -146,11 +157,16 @@ void shaderc_module_release(shaderc_spv_module_t module);
...
+// string is a char string. Otherwise, the result string is binary string.
 size_t shaderc_module_get_length(const shaderc_spv_module_t module);
 
-// Returns a pointer to the start of the SPIR-V bytes.
-// This is guaranteed to be castable to a uint32_t*.
+// Returns a pointer to the start of the SPIR-V bytes, either SPIR-V binary or
+// char string. When the source string is compiled into SPIR-V binary, this is
+// guaranteed to be castable to a uint32_t*. If the source string is compiled in
+// disassembly mode or preprocessing only mode, the pointer points to the result
+// char string, and it will not castable to a uint32_t*.
```
dneto0: Just leave off the last part of the last sentence: ", and it will not be castable to a uint32_t*".  Becuase it might be doable, but you have no guarantee.   Better to just not say anything.

dneto0:
-1 just some comment fixups.  Getting close. :-)

dneto0:
+2
I will squash an merge into master.

dneto0:
Merged into master

PR #57, opened 2015-12-22
Expose set disassembly mode c api

Qining:


libshaderc/src/shaderc_test.cc
```diff
@@ -216,6 +216,36 @@ TEST(CompileStringWithOptions, MacroCompileOptions) {
...
+                   shaderc_glsl_vertex_shader, options.get());
+  EXPECT_TRUE(shaderc_module_get_success(comp.result()));
+  // This should work with both the glslang native disassembly format and the
+  // SPIR-V tools assembly format.
+  EXPECT_THAT(shaderc_module_get_bytes(comp.result()),
+              HasSubstr("Capability Shader"));
+  EXPECT_THAT(shaderc_module_get_bytes(comp.result()),
+              HasSubstr("MemoryModel"));
+
+  const std::string kMinimalShaderCloneOption = "void main(){}\n";
```
dneto0: Optional: You don't need to make a new string for the second test.

dneto0:
Don't make a new std::string variable when you don't need one.

dneto0:
+2
Merged into master.

PR #58, opened 2015-12-23
Add cpp interface for set disassembly mode and preprocessing only mode

Qining:


libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -249,6 +249,52 @@ TEST(CppInterface, MacroCompileOptions) {
...
 
+TEST(CppInterface, DisassemblyOption) {
+  shaderc::Compiler compiler;
+  shaderc::CompileOptions options;
+  options.SetDisassemblyMode();
+  const std::string kMinimalShader = "void main(){}\n";
+  shaderc::SpvModule result = compiler.CompileGlslToSpv(
+      kMinimalShader, shaderc_glsl_vertex_shader, options);
+  EXPECT_TRUE(result.GetSuccess());
+  // This should work with both the glslang native disassembly formt and the
```
dneto0: spelling "format"

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -249,6 +249,52 @@ TEST(CppInterface, MacroCompileOptions) {
...
+  // SPIR-V tools assembly format.
```
dneto0: Captialize "SPIR-V Tools" since that's a proper project name, and the name of a GitHub repo.

libshaderc/include/shaderc.hpp
```diff
@@ -104,6 +104,23 @@ class CompileOptions {
     AddMacroDefinition(name.c_str(), value.c_str());
   }
 
+  // Set the compiler to emit a disassembly text instead of a binary. In
```
dneto0: Today I feel like asking for the blurbs to use standard style. So:
"Sets the compiler mode to emit disassembly text instead of a binary."   Ok, to just do for the C++ API.  We can do the C API another time.

libshaderc/include/shaderc.hpp
```diff
@@ -104,6 +104,23 @@ class CompileOptions {
...
+  // this mode, the byte array result in the shaderc_spv_module returned
+  // from shaderc_compile_into_spv() will consist of SPIR-V assembly text.
+  // Note the preprocessing only mode overrides this option, and this option
+  // overrides the default mode generating a SPIR-V binary.
+  void SetDisassemblyMode(){
+    shaderc_compile_options_set_disassembly_mode(options_);
+  }
+
+  // Set the compiler to do only preprocessing. The byte array result in the
```
dneto0: "Sets the compiler mode to only do preprocessing."

dneto0:
-1 for comment fixups.
Also, we promise that preprocessing mode overrides disassembly mode.  So we should have a test that proves it.  That is, a test that sets both options, but only gets preprocessing mode. Test both orders: set preprocessing first in one test, and another test with disassembly mode first.

dneto0:
+2  
Will squash and merge

dneto0:
merged into master

PR #61, opened 2016-01-04
Make Compilation::result() const.

dekimir:
And run clang-format on libshaderc/src/shaderc_test.cc.

### DIFF
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -72,7 +72,7 @@ class Compilation {
 
   ~Compilation() { shaderc_module_release(compiled_result_); }
 
-  shaderc_spv_module_t result() { return compiled_result_; }
+  shaderc_spv_module_t result() const { return compiled_result_; }
 
  private:
   shaderc_spv_module_t compiled_result_;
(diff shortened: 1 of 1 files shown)


libshaderc/src/shaderc_test.cc
```diff
@@ -222,8 +222,8 @@ TEST(CompileStringWithOptions, DisassemblyOption) {
   shaderc_compile_options_set_disassembly_mode(options.get());
   ASSERT_NE(nullptr, compiler.get_compiler_handle());
   const std::string kMinimalShader = "void main(){}\n";
-  Compilation comp(compiler.get_compiler_handle(), kMinimalShader,
-                   shaderc_glsl_vertex_shader, options.get());
+  const Compilation comp(compiler.get_compiler_handle(), kMinimalShader,
```
dneto0: This isn't mentioned by the commit comment.
But it's *enabled* by the functional change described in the commit comment (making the result() method const).

I'm ok with this in this CL since it's in test code only.  (No public semantic impact.)  Otherwise I'd prefer a slightly expanded commit comment.

dneto0:
+2
AppVeyor issue looks independent.

PR #66, opened 2016-01-05
Simplify shaderc_cpp tests.

dekimir:
Move common code into a fixture so tests can be brief and direct.

### DIFF
--- a/libshaderc/src/shaderc_cpp_test.cc
+++ b/libshaderc/src/shaderc_cpp_test.cc
@@ -47,7 +47,8 @@ const std::string kMinimalUnknownVersionShader =
 class CppInterface : public testing::Test {
  protected:
   // Compiles a shader and returns true on success, false on failure.
-  bool CompilationSuccess(const std::string& shader, shaderc_shader_kind kind) {
+  bool CompilationSuccess(const std::string& shader,
+                          shaderc_shader_kind kind) const {
     return compiler_.CompileGlslToSpv(shader.c_str(), shader.length(), kind)
         .GetSuccess();
   }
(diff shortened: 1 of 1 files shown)


libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -72,121 +127,75 @@ TEST(CppInterface, MultipleThreadsInitializing) {
...
+  ASSERT_FALSE(compiler_.IsValid());
   ASSERT_TRUE(compiler2.IsValid());
 }
 
-TEST(CppInterface, EmptyString) {
-  shaderc::Compiler compiler;
-  ASSERT_TRUE(compiler.IsValid());
-  EXPECT_FALSE(CompilationSuccess(compiler, "", shaderc_glsl_vertex_shader));
-  EXPECT_FALSE(CompilationSuccess(compiler, "", shaderc_glsl_fragment_shader));
+TEST_F(CppInterface, EmptyString) {
```
dneto0: All the early tests of the compiler call copmiler_.IsValid() before trying to compile anything.
So it's not clear that you should be able to compile code *without* calling IsValid first.  There should not be any hidden state initialized in the IsValid method.  It should all be ready via construction (RAII).

Please add an early test, such as CanCompileEmptyStringWithoutCheckingIsValidFirst or something like that.

dekimir: How does that relate to this PR?  The tests are identical, just DRYer.

dneto0: It's something I noticed while reviewing this PR.

Agreed, it's primarly a DRYness change.

dneto0:
+2

dekimir:
Merged via bc7d444.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -44,14 +44,69 @@ const std::string kMinimalUnknownVersionShader =
...
+    return IsValidSpv(compiler_.CompileGlslToSpv(shader, kind, options));
+  }
+
+  // Compiles a shader, asserts compilation success, and returns the error
+  // message.
+  std::string CompilationErrors(const std::string& shader,
+                                shaderc_shader_kind kind,
+                                const shaderc::CompileOptions& options) {
+    const auto module = compiler_.CompileGlslToSpv(shader, kind, options);
+    EXPECT_TRUE(module.GetSuccess()) << kind << '\n' << shader;
```
Qining: I think there are cases that the compilation can fail and we want to check the specific error message.

E.g. when an error is caused by a wrong version-profile, we need to make sure the compilation fail and the error message shows the wrong version-profile.

I think we don't need to assert compilation success here.

dekimir: That makes sense.  I only left it in there because it was common to all the call sites.  But now it strikes me that these call sites are actually looking for compilation _warnings_ not errors.  I'll rename this method to `CompilationWarnings`, and @Qining will add a new `CompilationErrors` that asserts failure (and use it in the `ErrorsReported` test case, among others).

PR #73, opened 2016-01-06
Fix several compilation warnings/errors with VS2013

mgadams:
These were encountered when building in an environment with many warnings enabled.


antiagainst:
LGTM

PR #74, opened 2016-01-06
Fix warning caused by braces around scalar initializer.

antiagainst:
Reported by Clang with version number higher than 3.6.


dneto0:
+2

PR #75, opened 2016-01-06
Set debug info

Qining:
Add set_generate_debug_info() to libshaderc api.

Update/add CompilationOutput() to construct output string from 'char*' and 'length',so that we can check the content of the complete output binary array.

Fix some format problems.

### DIFF
--- a/libshaderc/src/shaderc_cpp_test.cc
+++ b/libshaderc/src/shaderc_cpp_test.cc
@@ -121,6 +121,10 @@ class CppInterface : public testing::Test {
                                 const CompileOptions& options) const {
     const auto module = compiler_.CompileGlslToSpv(shader, kind, options);
     EXPECT_TRUE(module.GetSuccess()) << kind << '\n';
+    // Use string(const char* s, size_t n) constructor instead of
+    // string(const char* s) to make sure the string has complete binary data.
+    // string(const char* s) assumes a null-terminated C-string, which will cut
+    // the binary data when it sees a '\0' byte.
     return std::string(module.GetData(), module.GetLength());
   }
 
--- a/libshaderc/src/shaderc_cpp_test.cc
+++ b/libshaderc/src/shaderc_cpp_test.cc
@@ -51,6 +51,12 @@ const std::string kCoreVertShaderWithoutVersion =
     "gl_ClipDistance[0] = 5.;\n"
     "}\n";
 
+// Debug information should contain the name of the vector: debug_info_sample.
+const std::string kMinimalDebugInfoShader =
+    "void main(){\n"
+    "vec2 debug_info_sample = vec2(1.0, 1.0);\n"
+    "}\n";
+
 class CppInterface : public testing::Test {
  protected:
   // Compiles a shader and returns true on success, false on failure.
@@ -106,6 +112,16 @@ class CppInterface : public testing::Test {
     const auto module = compiler_.CompileGlslToSpv(shader, kind, options);
     EXPECT_FALSE(module.GetSuccess()) << kind << '\n' << shader;
     return module.GetErrorMessage();
+  }
+
+  // Compiles a shader, expects compilation success, and returns the output
+  // bytes.
+  std::string CompilationOutput(const std::string& shader,
+                                shaderc_shader_kind kind,
+                                const CompileOptions& options) const {
+    const auto module = compiler_.CompileGlslToSpv(shader, kind, options);
+    EXPECT_TRUE(module.GetSuccess()) << kind << '\n';
+    return std::string(module.GetData(), module.GetLength());
   }
 
   // For compiling shaders in subclass tests:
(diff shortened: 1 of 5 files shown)


libshaderc/src/shaderc_test.cc
```diff
@@ -175,7 +180,8 @@ class CompileStringTest : public testing::Test {
                            options);
     EXPECT_TRUE(shaderc_module_get_success(comp.result())) << kind << '\n'
                                                            << shader;
-    return shaderc_module_get_bytes(comp.result());
+    return std::string(shaderc_module_get_bytes(comp.result()),
```
Qining: If we don't use the 'length' to build the string and return char*, EXPECT_THAT(...HasSubstr()) will cut the output binary array when it sees a '\0' byte.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -102,6 +108,16 @@ class CppInterface : public testing::Test {
     return module.GetErrorMessage();
   }
 
+  // Compiles a shader, expects compilation success, and returns the output
+  // bytes.
+  std::string CompilationOutput(const std::string& shader,
```
dneto0: In SPIR-V Tools tests we've found it much more useful to return this as a vector of uint32_t.
We had a bunch of utilities in the test code for manually creating vectors that matched SPIR-V instructions, then we could check element subsequence equality.
Please change this interface to return a vector of uint32_t.

dneto0: Ah. You use this for both binary and disassembled compilation.
In that case I'd recommend making two methods, one for each output form.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -290,6 +306,44 @@ TEST_F(CppInterface, DisassemblyOption) {
...
+              HasSubstr("debug_info_sample"));
+}
+
+TEST_F(CppInterface, GenerateDebugInfoDisassemblyClonedOptions) {
+  options_.SetGenerateDebugInfo();
+  // Generate debug info mode should be carried to the cloned options.
+  CompileOptions cloned_options(options_);
+  EXPECT_THAT(CompilationOutput(kMinimalDebugInfoShader,
+                                shaderc_glsl_vertex_shader, cloned_options),
+              HasSubstr("debug_info"));
```
dneto0: May as well check the full name "debug_info_sample", like in other places.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -290,6 +306,44 @@ TEST_F(CppInterface, DisassemblyOption) {
   EXPECT_THAT(result_from_cloned_options.GetData(), HasSubstr("MemoryModel"));
 }
 
+TEST_F(CppInterface, GenerateDebugInfoBinary) {
+  options_.SetGenerateDebugInfo();
+  // The output binary should contain the name of the vector: debug_info_sample as char
+  // array.
+  EXPECT_THAT(CompilationOutput(kMinimalDebugInfoShader,
+                                shaderc_glsl_vertex_shader, options_),
+              HasSubstr("debug_info_sample"));
```
dneto0: We could and should make this check stronger.

I would favour making a utility function that takes a SPIR-V binary, and a name string, and checks for an OpName instruction for an ID for that name. 
You can use the spvBinaryParse function from SPIR-V Tools to do the parsing for you, and send in a lambda for the instruction handler. 

Also add a TODO to check for OpLine annotations.  That can only be added once glslang supports full debug output (including OpLine)

libshaderc/src/shaderc_test.cc
dneto0: This doesn't belong in the MR comments.  Put the comment in the code.
Also, see above, I argue that you should have two CompilationOutput methods: one for string, and another for the binary, so you can more naturally check the binary contents.

libshaderc/src/shaderc_test.cc
```diff
@@ -283,6 +291,39 @@ TEST_F(CompileStringWithOptionsTest, DisassemblyOption) {
   EXPECT_THAT(disassembly_text_cloned_options, HasSubstr("MemoryModel"));
 }
 
+TEST_F(CompileStringWithOptionsTest, GenerateDebugInfoBinary) {
+  shaderc_compile_options_set_generate_debug_info(options_.get());
+  ASSERT_NE(nullptr, compiler_.get_compiler_handle());
+  // The binary output should contain the name of the vector: debug_info_sample.
+  EXPECT_THAT(CompilationOutput(kMinimalDebugInfoShader,
+                                shaderc_glsl_vertex_shader, options_.get()),
+              HasSubstr("debug_info_sample"));
```
dneto0: Again, we should have  stronger debug info test.

dneto0:
-1

libshaderc/src/shaderc_cpp_test.cc
Qining: I think we may have two Compilation**Output for binary and disassembly output respectively. But how to use the 'MakeVector' in Spirv-tools' test? It is an internal header, I think I shouldn't refer to that header, maybe I should copy the function to shaderc's tests? It is only a utility function depends only on STL.

Then, if I use the spvBinaryParse(), we will need to add SPIRV-Tools as a dependent library for shaderc's test, and only for the test. Is this what you mean, or I missed something?

libshaderc/src/shaderc_cpp_test.cc
dneto0: Ok.  Requireing the spvBinaryParse for the test adds a link dependency to shaderc that we don't want yet.
So keep the test as is.  Maybe add a TODO if you like saying how we could strengthen the test.

libshaderc/src/shaderc_cpp_test.cc
dneto0: Right.  I'll back down on this for now.
Maybe add a TODO with the idea for how to strengthen the test.

libshaderc/src/shaderc_test.cc
dneto0: Backing off on this suggestion.

dneto0:
squashed and merged

PR #76, opened 2016-01-07
Add set_forced_version_profile to libshaderc.

Qining:
This code doesn't parse the version-profile pair from string, it only pass the int/enum value down to libshaderc_util. The parsing should be done before calling SetForcedVersionProfile().

### DIFF
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -127,8 +127,10 @@ void shaderc_compile_options_add_macro_definition(
 void shaderc_compile_options_set_disassembly_mode(
     shaderc_compile_options_t options);
 
-// Sets the compiler mode to force (without any verification) the default
-// version and profile for compilation.
+// Forces the GLSL language version and profile to a given pair. The version
+// number is the same as would appear in the #version annotation in the source.
+// Version and profile specified here overrides the #version annotation in the
+// source.
 void shaderc_compile_options_set_forced_version_profile(
     shaderc_compile_options_t options, int version, shaderc_profile profile);
 
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -40,6 +40,8 @@ typedef enum {
 } shaderc_target_env;
 
 typedef enum {
+  shaderc_profile_none,  // Used if and only if GLSL version did not specify
+                         // profiles.
   shaderc_profile_core,
   shaderc_profile_compatibility,
   shaderc_profile_es,
@@ -130,7 +132,10 @@ void shaderc_compile_options_set_disassembly_mode(
 // Forces the GLSL language version and profile to a given pair. The version
 // number is the same as would appear in the #version annotation in the source.
 // Version and profile specified here overrides the #version annotation in the
-// source.
+// source. Versions before 150 do not allow a profile token, in that case
+// shaderc_profile_none should be passed down as the profile parameter, and
+// shaderc_profile_none should only be used when none profile is specified along
+// with version.
 void shaderc_compile_options_set_forced_version_profile(
     shaderc_compile_options_t options, int version, shaderc_profile profile);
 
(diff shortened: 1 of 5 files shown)


libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -290,6 +296,62 @@ TEST_F(CppInterface, DisassemblyOption) {
...
+                                 shaderc_glsl_vertex_shader, options_));
+}
+
+TEST_F(CppInterface, ForcedVersionProfileCorrectStdClonedOptions) {
+  // Forces the version and profile to 450core, which fixes the missing
+  // #version.
+  options_.SetForcedVersionProfile(450, shaderc_profile_core);
+  CompileOptions cloned_options(options_);
+  EXPECT_TRUE(CompilesToValidSpv(kCoreVertShaderWithoutVersion,
+                                 shaderc_glsl_vertex_shader, options_));
```
dneto0: Should have used cloned_options here.

libshaderc/include/shaderc.h
```diff
@@ -121,6 +127,11 @@ void shaderc_compile_options_add_macro_definition(
 void shaderc_compile_options_set_disassembly_mode(
     shaderc_compile_options_t options);
 
+// Sets the compiler mode to force (without any verification) the default
+// version and profile for compilation.
```
dneto0: I'd like this to say that it's forcing the *GLSL* language version and profile.  Should also say that the version number is the same as would appear in the #version annotation in the source. 

The blurb should describe how this setting interacts with the #version annotation in the source, if any such annotation exists.
(You have it in your test cases)

dneto0:
-1 for bug in the cloning test case.  Also, please strengthen the API method blurb.

libshaderc/src/shaderc_test.cc
```diff
@@ -283,6 +289,78 @@ TEST_F(CompileStringWithOptionsTest, DisassemblyOption) {
...
+  compile_options_ptr cloned_options(
+      shaderc_compile_options_clone(options_.get()));
+  ASSERT_NE(nullptr, compiler_.get_compiler_handle());
+  EXPECT_TRUE(CompilesToValidSpv(kCoreVertShaderWithoutVersion,
+                                 shaderc_glsl_vertex_shader,
+                                 cloned_options.get()));
+}
+
+TEST_F(CompileStringWithOptionsTest, ForcedVersionProfileWrongStd) {
+  // Forces the version and profile to 310es, which is a wrong pair.
```
dneto0: Fix the comment, and maybe the test name.
It's not that 310es is a wrong pair.  It's that the module is invalid for that version of GLSL.

libshaderc/src/shaderc_test.cc
```diff
@@ -283,6 +289,78 @@ TEST_F(CompileStringWithOptionsTest, DisassemblyOption) {
...
+  // Warning message should complain about the unknown version.
+  EXPECT_THAT(CompilationWarnings(kMinimalShader, shaderc_glsl_vertex_shader,
+                                  options_.get()),
+              HasSubstr("warning: version 4242 is unknown.\n"));
+}
+
+TEST_F(CompileStringWithOptionsTest, ForcedVersionProfileRedundantProfileStd) {
+  // Forces the version and profile to 100core. But versions before 150 do not
+  // allow a profile token, compilation should fail.
+  shaderc_compile_options_set_forced_version_profile(options_.get(), 100,
```
dneto0: This begs the question:  How are you supposed to set version 100?  What profile enum do you send down?
Maybe this proves you need a "none" option.  Or maybe it can only be the compatibility profile.
Since GLSL version 1.0 doesn't have profiles, I'd lean toward adding a "none" option.
In either case we'd have to document that behaviour.

dneto0:
-1
Looking again, it looks like we need  a "none" profile for use with versions lower than 150.
And we'd have to document its use.  It's uses if and only if the GLSL version did not specify profiles.

dneto0:
+2 
I'll squash and merge

dneto0:
squashed and merged

Issue #79, opened 2016-01-08
Disassembled output should be in standard form, as emitted by SPIRV-Tools disassembler


dneto0:
It almost certainly implies linking to SPIRV-Tools itself.

antiagainst:
An interesting issue related here is duplicate SPIR-V headers. If we link SPIRV-Tools in, then we have duplicate SPIR-V headers. One is `spirv.hpp` in glslang, The other is `spirv.h` in SPIRV-Tools. Sadly, both of these two headers have `#ifndef spirv_H` & `#define spirv_H` inside them.

antiagainst:
refer to https://github.com/KhronosGroup/SPIRV-Tools/issues/24

dneto0:
Until the header guards are made different, the workaround is that if you need both, then you have to include them from separate compilation units.  So you can wrap the calls to the SPIR-V APIs in a different compilation unit in Shaderc.   It's a bit ugly, but certainly doable.

antiagainst:
@dneto0 It turns out that we don't have some source code pulling in these two headers together. (Yay!) But I've something ugly which should work:

```
// after including header 1:
#ifdef spirv_H
#undef spirv_H
#undef SPIRV_VERSION
#undef SPIRV_REVISION
#endif 
// include header 2
```

It's lucky that everything in `spirv.hpp` is in the `spv` namespace. So we just need to `#undef` the two macros.

AWoloszyn:
@antiagainst  That will unfortunately break the header-guards for header2. So if you end up including something else that includes header2 you will still get double-definitions.
#ifndef spirv_H
#inlclude header1
#undef spirv_H
#undef SPIRV_VERSION
#undef SPIRV_REVISION
#include header2
#endif 
Is a BIT more correct, but it fails if something you include #includes header1 or header2 elsewhere.

antiagainst:
@AWoloszyn: yeah. That's better. Luckily we don't need this kind of hack, at least for now. :)

antiagainst:
Addressed via #81.

PR #81, opened 2016-01-09
Pull in SPIRV-Tools as a third-party project and disassemble using it

antiagainst:
* Addresses Issue #79.
* Configuration in `Android.mk` works (tested using `ndk-build`).
* More tests for `-S` (both `shaderc` & `glslc`) TODO


libshaderc_util/src/compiler.cc
```diff
@@ -20,16 +20,16 @@
...
 #include "libshaderc_util/shader_stage.h"
 #include "libshaderc_util/string_piece.h"
 #include "libshaderc_util/version_profile.h"
-#include "libshaderc_util/message.h"
 
-#include "SPIRV/disassemble.h"
-#include "SPIRV/doc.h"
 #include "SPIRV/GlslangToSpv.h"
 
+#include "libspirv/libspirv.h"
```
dneto0: include order.  This should go before SPIRV/*

libshaderc_util/src/compiler.cc
```diff
@@ -67,6 +67,45 @@ std::pair<int, string_piece> DecodeLineDirective(string_piece directive) {
...
+      spvtools_context, binary.data(), binary.size(),
+      /* options = */ 0u, &disassembled_text, &spvtools_diagnostic);
+  if (result == SPV_SUCCESS) {
+    text_or_error->assign(disassembled_text->str, disassembled_text->length);
+  } else {
+    OutputSpvToolsDiagnostic(spvtools_diagnostic, text_or_error);
+  }
+
+  if (spvtools_diagnostic != nullptr) {
+    spvDiagnosticDestroy(spvtools_diagnostic);
```
dneto0: This should be safe to call on nullptr.
I now see that that behviour is neither advertised nor tested.  :-(

libshaderc_util/src/compiler.cc
```diff
@@ -67,6 +67,45 @@ std::pair<int, string_piece> DecodeLineDirective(string_piece directive) {
...
+  }
+  if (disassembled_text != nullptr) {
```
dneto0: spvTextDestroy is safe to call on nullptr, so you don't need this guard.

libshaderc_util/src/compiler.cc
dneto0: I filed https://github.com/KhronosGroup/SPIRV-Tools/issues/62 for the SPIR-V Tools side enhancement.

libshaderc_util/src/compiler.cc
```diff
@@ -67,6 +67,45 @@ std::pair<int, string_piece> DecodeLineDirective(string_piece directive) {
...
+  os << diagnostic->position.index << ": " << diagnostic->error;
+  *dest = os.str();
+}
+
+// Disassembles the given binary. Returns SPV_SUCCESS and writes the
+// disassembled text to *text_or_error if successful. Otherwise, writes the
+// error message to *text_or_error.
+spv_result_t DisassembleBinary(const std::vector<uint32_t>& binary,
+                               std::string* text_or_error) {
+  auto spvtools_context = spvContextCreate();
```
dneto0: The SPIRV-Tools context could be made a member of the Compiler object.  Then we wouldn't have to create and destroy the SPIR-V Tools context object for each disassembled module.

Right now it does an allocation, but long term, I'd like to precompile all the data in the SPIR-V Tools context object so it wouldn't need extra memory or time to set up.  
But that hasn't happened just yet, so I'd feel better pushing this context object out to the Compiler object.

dneto0:
-1 mostly minor things
Also the README needs a sentence or two about the dependency on SPIR-V Tools.  (Similar to dependency on glslang, and googletest.)

libshaderc_util/src/compiler.cc
antiagainst: I'm afraid that won't work because `shaderc_util::Compiler` is a part of `struct shaderc_compile_options`. The option struct supports copying. We don't want the SPIRV-Tools context object to be copied around.

Instead, I have another [commit](https://github.com/antiagainst/shaderc/commit/bc0999f51dd833a189280bfe22962ef0592c3940) to wrap this implementation using a singleton, which should solve the frequent construction/destruction.

libshaderc_util/src/compiler.cc
antiagainst: Done.

libshaderc_util/src/compiler.cc
antiagainst: Good catch! Done.

libshaderc_util/src/compiler.cc
antiagainst: Done.

libshaderc_util/src/compiler.cc
dneto0: Oh, ok.  Please keep the current create/destroy pair.  I don't like singletons.

dneto0:
+2 to what you have now.  I will squash and merge.

dneto0:
Hm.  I'm having trouble integrating this into a project that already includes both Shaderc and SPIRV-Tools.  Is there a way to allow shaderc to compile automatically if spirv-tools is already present in the Cmake-based project?  
Something like what SPIRV-Tools does to automatically detect gmock or use a local copy?

So -1 again for this integration.

antiagainst:
@dneto0 I've updated another commit to use CMake mechanisms to figure out SPIRV-Tools binary directory. In this way, SPIRV-Tools can be placed in a different place.

As for your question, I think the reason is directory including order. Both shaderc and spirv-tools depend on googletest. And shaderc depends on spirv-tools. So you'll need to carefully include these three projects. E.g., if you place them in the same dir, make sure
```cmake
add_subdirectory(googletest)
add_subdirectory(spirv-tools)
add_subdirectory(shaderc)
```

antiagainst:
rebased and cleaned the git history.

README.md
```diff
@@ -41,6 +41,9 @@ copy only to stage those changes for review, and to provide something for
 Shaderc to build against in the meantime.  Please see
 [DEVELOPMENT.howto.md](DEVELOPMENT.howto.md) for more details.
 
+Shaderc depends on [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools)
```
dekimir: Either insert "also" here or remove "also" from line 47.

README.md
```diff
@@ -54,8 +57,9 @@ Shaderc into.
 ```sh
 git clone https://github.com/google/shaderc $SOURCE_DIR
 cd $SOURCE_DIR/third_party
-git clone https://github.com/google/googletest.git
-git clone https://github.com/google/glslang glslang
+git clone https://github.com/google/googletest.git googletest
+git clone https://github.com/google/glslang.git glslang
+git clone https://github.com/KhronosGroup/SPIRV-Tools.git third_party/spirv-tools
```
dekimir: Remove `third_party/` from this line.

dekimir: Also, why do we need the last argument on the previous three lines?  Were the instructions wrong before?

dneto0: Good catch:  I don't think we need the last arguments glslang, googletest.
But I think we do need the last argument for spirv-tools since the case is different between the repo name and the local directory name.

README.md
antiagainst: Done.

README.md
antiagainst: Good catch about `third_party`.

re: last argument: It's not necessary, but thus we have consistency and it looks better. But I removed it since there are concerns around it.

dekimir:
+1

third_party/Android.mk
```diff
@@ -83,3 +85,33 @@ LOCAL_C_INCLUDES:=$(GLSLANG_LOCAL_PATH)
...
+		source/diagnostic.cpp \
+		source/ext_inst.cpp \
+		source/opcode.cpp \
+		source/operand.cpp \
+		source/print.cpp \
+		source/spirv_endian.cpp \
+		source/table.cpp \
+		source/text.cpp \
+		source/text_handler.cpp \
+		source/validate.cpp \
```
dneto0: There are some new validator source files.  :-(

dneto0:
My last objection is that there are new validator source files that need to be listed in the Android makefile.

Once that's in, I'd give this a +2

antiagainst:
@dneto0 That is done. Also rebased against master.

antiagainst:
OMG. Clicked on the wrong button...

dneto0:
+2

libshaderc_util/src/compiler.cc
```diff
@@ -67,6 +67,41 @@ std::pair<int, string_piece> DecodeLineDirective(string_piece directive) {
   return std::make_pair(line, directive);
 }
 
+// Writes the message contained in the diagnostic parameter to *dest. Assumes
+// the diagnostic message is reported for for a binary location.
```
dekimir: Typo "for for".

antiagainst: Done.

PR #83, opened 2016-01-11
Use a newer version of GCC for Travis CI.

antiagainst:


dneto0:
+1
It would be nice to know why we want this, especially since it's a newer GCC than what's in Ubuntu 14.04

AWoloszyn:
If anything I would prefer using older versions of the toolchains.

For the record the NDK (r10e) supports gcc-4.8 and clang-3.5.

antiagainst:
@dneto0 @AWoloszyn 

GCC 4.8 provided by Travis is 4.8.1. It has internal errors when compiling SPIRV-Tools:

https://travis-ci.org/google/shaderc/jobs/101674113
https://travis-ci.org/antiagainst/SPIRV-Tools/jobs/100751593

antiagainst:
GCC 4.8.4 is tested and works fine. But I haven't tested 4.8.1 on my local machine yet.

dneto0:
I filed an issue against Shaderc to document which version of GCC will work.
So
+2

antiagainst:
Committed as 99a202aa83be25bd0808f665f8a128e8a242ce14.

PR #87, opened 2016-01-12
Set includer

Qining:


libshaderc/include/shaderc.h
```diff
@@ -47,6 +47,7 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
+
```
dekimir: Extra line.

libshaderc/include/shaderc.h
```diff
@@ -141,6 +142,56 @@ void shaderc_compile_options_set_disassembly_mode(
 void shaderc_compile_options_set_forced_version_profile(
     shaderc_compile_options_t options, int version, shaderc_profile profile);
 
+// When including a file, libshaderc initiates a query to get the full path
```
dekimir: Why "the full path"? There is no guarantee that a particular includer even has the notion of a path.

Qining: You are right. I saw the comments in glslang::Includer say 'include()' returns the 'full path' and 'content', so I use 'fullpath' here. I think we can change the variable name to anything make sense. It seems that Glslang doesn't require the 'full path' to have any specific meaning.

libshaderc/include/shaderc.h
```diff
@@ -141,6 +142,56 @@ void shaderc_compile_options_set_disassembly_mode(
...
+typedef void (*shaderc_includer_fullpath_and_content_finalizer_fn)(
+    void* user_data, const char* filename,
+    shaderc_includer_fullpath_content data);
+
+// A struct that contains all the elements that a includer requires from client
+// side, including two function pointers and user_data as contexts for calling
+// each of the functions.
+struct shaderc_includer_callbacks {
+  shaderc_includer_fullpath_and_content_getter_fn get_fullpath_and_content;
+  void* getter_user_data;
```
AWoloszyn: This could probably be a single user_data object rather than 2.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +504,51 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+  // Create a fake file object, with specified fullpath and content.
+  struct FakeFile {
+    std::string fullpath = "Full/Path/Sample";
+    std::string content = "void main() {foo();}\n";
+  };
+  FakeFile fake_file;
+  shaderc_includer_callbacks callbacks{
+      // A response method that returns fake full path and file content as
+      // defined above.
+      [](void* user_data, const char* filename) {
```
AWoloszyn: I have a slight preference for 
```+[](void*....){}``` so that the error message that will be shown if/when we ever add captures to this lambda are more clear.

Comment (Qining) [inline libshaderc/src/shaderc_cpp_test.cc]: Yes, will add it now.

File: libshaderc/include/shaderc.h (hunk 3 above)
Comment (Qining) [inline libshaderc/include/shaderc.h]: Originally I use one object, but I think we can not guarantee the client code to use same user_data for both 'getting the data' and 'release the data'.

File: libshaderc/src/shaderc_cpp_test.cc (hunk 4 above)
Comment (AWoloszyn) [inline libshaderc/src/shaderc_cpp_test.cc]: It is scattered around the spec, but http://stackoverflow.com/questions/17822131/resolving-ambiguous-overload-on-function-pointer-and-stdfunction-for-a-lambda has a good summary.

Comment (Qining) [inline libshaderc/src/shaderc_cpp_test.cc]: I found appveyor fails on that '+' usage. Maybe VSC++ doesn't support this usage?

File: libshaderc/include/shaderc.h [hunk 5]
```diff
@@ -141,6 +141,55 @@ void shaderc_compile_options_set_disassembly_mode(
 void shaderc_compile_options_set_forced_version_profile(
     shaderc_compile_options_t options, int version, shaderc_profile profile);
 
+// When including a file, libshaderc initiates a query to get the full
+// path and the content of a file. The query is initiated through calling a
```
dneto0: I normally call these "callbacks". So you could simplify the wording.
You could simplify the first two sentences with:


```
To support file inclusion, libshaderc invokes a callback into its client to resolve the full path and content of the included file.   
```
Later phrases of 'client-side implemented method' can be replaced with 'client callback'.

libshaderc/include/shaderc.h
dneto0: We can impose that requirement.  It will make it simpler. The client can always make that user data be a struct with two user-data pointers.  :-)

Basically, rule of thumb should be one user-data pointer for each interaction protocol.

libshaderc/include/shaderc.h
```diff
@@ -141,6 +141,55 @@ void shaderc_compile_options_set_disassembly_mode(
...
+// The function signature of the client-side implemented reponse method. It
+// returns shaderc_includer_fullpath_content struct.
+typedef shaderc_includer_fullpath_content (
+    *shaderc_includer_fullpath_and_content_getter_fn)(void* user_data,
+                                                      const char* filename);
+
+// The function signature of the client-side implemented clean-up method.
+// Includer will call this callback function when the including process is done
+// with the fullpath and content data.
+typedef void (*shaderc_includer_fullpath_and_content_finalizer_fn)(
```
dneto0: This seems overly complex.  Why do you need to provide the original filename?  

I expected the original query to return a *pointer* to a shadec_includer_response, then this "finalizer" would really be
  typedef void (*shaderc_includer_response_release_fn)(void* user_data, shaderc_includer_response* data);

libshaderc/include/shaderc.h
```diff
@@ -141,6 +141,55 @@ void shaderc_compile_options_set_disassembly_mode(
...
+    void* user_data, const char* filename,
+    shaderc_includer_fullpath_content data);
+
+// A struct that contains all the elements that a includer requires from client
+// side, including two function pointers and user_data as contexts for calling
+// each of the functions.
+struct shaderc_includer_callbacks {
```
dneto0: It seems overweight to make a struct just for this.
Just make three arguments to the setter: the user_data, the request callback, and the relase-response callback.

libshaderc/include/shaderc.hpp
```diff
@@ -125,6 +125,10 @@ class CompileOptions {
     shaderc_compile_options_set_generate_debug_info(options_);
   }
 
+  void SetIncluderCallbacks(shaderc_includer_callbacks* callbacks_ptr) {
```
dneto0: You would need the same 3 arguments to the C API setter.

libshaderc/include/shaderc.hpp
```diff
@@ -125,6 +125,14 @@ class CompileOptions {
     shaderc_compile_options_set_generate_debug_info(options_);
   }
 
+  void SetIncluderCallbacks(
```
AWoloszyn: If we are exposing a C++ api, would it make sense to not use raw C functions and either
A) have this take an object that implements an INCLUDER interface or
```
void SetIncluder(std::unique_ptr<Includer>&& includer)
```
or at least

B) void SetIncluder(std::function<...> get_includer_response,
                                std::function<...> release_includer_response)
// Note there is no user_data* since std::function can contain a
// struct/function/lambda.

Qining: Just updated the PR, option A is selected.

libshaderc/include/shaderc.h
```diff
@@ -141,6 +141,46 @@ void shaderc_compile_options_set_disassembly_mode(
...
+// the underlying data is owned by client code.
+struct shaderc_includer_response {
+  const char* path;
+  size_t path_length;
+  const char* content;
+  size_t content_length;
+};
+
+// The function signature of the client-side implemented reponse method. It
+// returns shaderc_includer_reponsestruct.
```
dneto0: ..."returns a pointer to a shaderc_includer_response."

Missed the pointer part.

libshaderc/include/shaderc.h
```diff
@@ -141,6 +141,46 @@ void shaderc_compile_options_set_disassembly_mode(
 void shaderc_compile_options_set_forced_version_profile(
     shaderc_compile_options_t options, int version, shaderc_profile profile);
 
+// To support file inclusion, libshaderc invokes a callback into its client to
+// resolve the full path and content of the included file.
+// Client callback should follow the specified function signature below, and
```
dneto0: Need a "the" here:
"The client callback..."

libshaderc/include/shaderc.h
```diff
@@ -141,6 +141,46 @@ void shaderc_compile_options_set_disassembly_mode(
...
+// it should be passed to libshaderc through the corresponding setter function.
+// When the including of a file is done, libshaderc will call another client
+// callback to clean up the resources used for the including process.  Client
```
dneto0: "The client should ..."

libshaderc/include/shaderc.hpp
```diff
@@ -125,6 +126,35 @@ class CompileOptions {
     shaderc_compile_options_set_generate_debug_info(options_);
   }
 
+  // Define an includer interface, on which libshaderc will invoke the methods
+  // to resolve the full path and content of the file to be included, and clean
+  // up the resources used for the including process. Client code should
+  // implement this interface and configure the includer instance then set it to
+  // libshaderc through following api.
```
dneto0: api -> API

libshaderc/include/shaderc.hpp
```diff
@@ -125,6 +126,35 @@ class CompileOptions {
...
+  class Includer {
+   public:
+    virtual shaderc_includer_response* GetIncluderResponse(
```
dneto0: The "Includer" part of these methods is redundant with the class name. You already know you're working with an includer.

I think a more natural naming is:
```
GetInclude
ReleaseInclude
```

libshaderc/include/shaderc.hpp
dneto0: :-(
The solution for (A) creates a new class, and forces a class-type based hierarchy on the clients.  I actually prefer the std::function solution.  It's more elemental and does not impose a solution structure on the client.

libshaderc/include/shaderc.hpp
dneto0: If you want to keep the current solution, please rename Includer -> IncluderInterface
That's more in line with the advice in the Google style guide for permitted uses of multiple inheritance (it may be that a client uses multiple inheritance as part of this solution)

libshaderc/src/shaderc.cc
```diff
@@ -85,21 +85,55 @@ struct {
 
 std::mutex compile_mutex;  // Guards shaderc_compile_*.
 
-// Rejects #include directives.
-class ForbidInclude : public shaderc_util::CountingIncluder {
+class InternalFileIncluder : public shaderc_util::CountingIncluder {
  private:
-  // Returns empty contents and an error message.  See base-class version.
+  bool AreValidCallbacks() const {
```
dneto0: Needs a blurb.

libshaderc/include/shaderc.h
```diff
@@ -141,6 +141,46 @@ void shaderc_compile_options_set_disassembly_mode(
...
+  const char* path;
+  size_t path_length;
+  const char* content;
+  size_t content_length;
+};
+
+// The function signature of the client-side implemented reponse method. It
+// returns shaderc_includer_reponsestruct.
+typedef shaderc_includer_response* (
+    *shaderc_includer_response_get_fn)(void* user_data,
```
dneto0: Wait a second.  This interface is missing information.
It should take two things:
-   const char * context_path that gives the full path of the file that is doing the including.   
- the filename being requested.

That's required to support stuff like this:
  first file includes   "subdir/a.h"
  and then subdir/a.h  includes "b.h"
For normal processing, the includer should try to find "b.h" as "subdir/b.h".  But to know it must do that, it must know which file is doing the including. This is the whole reason we wanted the path element in the includer response in the first place.

We also have to spell out what happens on the *first* inclusion.  We could make the rule that context_path is either nullptr or an empty string.  (Empty string would play nicer in the C++ space where the context path would probably be a const std::string& parameter.)  So probably empty string is better.

libshaderc/include/shaderc.hpp
dneto0: Here again, would need a context_path argument, like in the C API.

libshaderc/src/shaderc.cc
```diff
@@ -85,21 +85,55 @@ struct {
...
+    return GetIncluderResponse_ != nullptr &&
+           ReleaseIncluderResponse_ != nullptr;
+  }
+
+  // Find filename in search path and returns its contents.
   std::pair<std::string, std::string> include_delegate(
       const char* filename) const override {
```
dneto0: Should have public first, then protected, private.
https://google.github.io/styleguide/cppguide.html#Declaration_Order

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+ public:
+  // Create a fake file object, with specified path and content.
+  struct FakeFile {
+    std::string path;
+    std::string content;
+  };
+  // Use hashmap as a fake file system to store fake files to be included.
+  using FakeFS = std::unordered_map<std::string, FakeFile>;
+
+  TestIncluder(FakeFS& fake_fs) : fake_fs_(fake_fs){};
```
dneto0: The semicolon is redunant.
Also, I suspect this isn't clang-formatted.    I expected a space before {}

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+
+  // Get path and content from the fake file system.
+  shaderc_includer_response* GetIncluderResponse(
```
dneto0: You can't just reuse the same response object.  The client might request many independent or stacked includes before releasing any of them.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+
+  // Response data is owned as private property, no need to release explicitly.
+  void ReleaseIncluderResponse(shaderc_includer_response* data) override {
+    (void)data;
+  };
+
+ private:
+  FakeFS& fake_fs_;
+  // Includer response data is stored as private property.
+  shaderc_includer_response response_;
```
dneto0: You'll need at least one object per requested but un-released response.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+        "#include \"file_1\"\n"
+        "void main() {foo(); bar();}\n"}},
+      {"file_1", {"path/to/file_1", "void bar() {}\n"}},
+  });
+
+  options_.SetIncluder(
+      std::unique_ptr<TestIncluder>(new TestIncluder(fake_fs)));
+
+  // Expect the compilation to succeed.
+  EXPECT_TRUE(CompilesToValidSpv(kMainIncludingShader,
```
dneto0: Compiling all the way down to object code expands the scope of the test too far.  Should only look at the preprocessed output.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+    response_.path = file_to_include.path.c_str();
+    response_.path_length = file_to_include.path.size();
+    response_.content = file_to_include.content.c_str();
+    response_.content_length = file_to_include.content.size();
+    return &response_;
+  };
+
+  // Response data is owned as private property, no need to release explicitly.
+  void ReleaseIncluderResponse(shaderc_includer_response* data) override {
+    (void)data;
```
dneto0: Shouldn't have to do the (void)data thing.   To avoid a warning, delete the parameter name, i.e.
do:
```
ReleaseIncluderResponse(shaderc_includer_response*) override {}
```

libshaderc/src/shaderc_test.cc
```diff
@@ -482,6 +483,190 @@ TEST_F(CompileStringWithOptionsTest, PreprocessingOnlyOption) {
   EXPECT_THAT(preprocessed_text_cloned_options, HasSubstr("void main(){ }"));
 }
 
+// A mock class that simulate a client of the includer.
+class IncluderResponsor {
```
dneto0: IncluderResponder

libshaderc/src/shaderc_test.cc
```diff
@@ -482,6 +483,190 @@ TEST_F(CompileStringWithOptionsTest, PreprocessingOnlyOption) {
...
+    response_.content = file_to_include.content.c_str();
+    response_.content_length = file_to_include.content.size();
+    return &response_;
+  };
+
+  // Response data is owned as private property, no need to release explicitly.
+  void ReleaseIncluderResponse(shaderc_includer_response* data) { (void)data; };
+
+  // Wrapper for the corresponding member function.
+  static shaderc_includer_response* GetIncluderResponseCb(
```
dneto0: What does "Cb" mean?

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+                                              "#line 3"));
+}
+
+TEST_F(CppInterface, IncluderNestedInclude) {
+  const std::string kMainIncludingShader =
+      "void foo() {}\n"
+      "#include \"file_0\"\n";
+
+  TestIncluder::FakeFS fake_fs({
+      {"file_0",
```
dneto0: Should have a nested include as given in my previous comments, where an included file is in a subdirectory, and references another file in the same subdir, but the #include doesn't have the subdir name.

libshaderc/include/shaderc.hpp
```diff
@@ -125,6 +126,35 @@ class CompileOptions {
...
+  // implement this interface and configure the includer instance then set it to
+  // libshaderc through following API.
+  class Includer {
+   public:
+    virtual shaderc_includer_response* GetInclude(
+        const char* filename) = 0;
+    virtual void ReleaseInclude(shaderc_includer_response* data) = 0;
+  };
+
+  // Sets the includer instance for libshaderc to call on when resolving
```
AWoloszyn: It is at fairly obvious from the interface, but might be worth explicitly mentioning that the ownership of the includer is passed to the CompileOptions, and will be destroyed with the CompileOptions.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,106 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+                                 shaderc_glsl_vertex_shader, options_));
+
+  // Sets compiler to preprocessing only mode, so we can check the result after
+  // preprocessing. Expect to see the content of the fake file object in the
+  // output of preprocessing.
+  options_.SetPreprocessingOnlyMode();
+  std::string preprocessing_output = CompilationOutput(
+      kMainIncludingShader, shaderc_glsl_vertex_shader, options_);
+  EXPECT_THAT(preprocessing_output, HasSubstr("#line 0 \"path/to/file_0\"\n"
+                                              " void bar(){ }\n"
```
dneto0: It seems to me all the cases can just reuse the same test code, parameterized by the fake filesystem contents, the name of an initial file (or its contents), and  the expected preprocessor output.
It would be easier to read the tests and see if they are covering the entire space we want.

libshaderc/src/shaderc_test.cc
```diff
@@ -482,6 +483,190 @@ TEST_F(CompileStringWithOptionsTest, PreprocessingOnlyOption) {
...
+      kMainIncludingShader, shaderc_glsl_vertex_shader, options_.get());
+  EXPECT_THAT(preprocessing_output, HasSubstr("#line 0 \"path/to/file_0\"\n"
+                                              "#line 0 \"path/to/file_1\"\n"
+                                              " void bar(){ }\n"
+                                              "#line 1 \"path/to/file_0\"\n"
+                                              " void main(){ foo();bar();}\n"
+                                              "#line 2"));
+}
+
 TEST_F(CompileStringWithOptionsTest, WarningsOnLine) {
```
dneto0: Similarly, should implement the same suite of tests as the C++ API tests.  And use a single test implementation but parameterized by the fake filesystem contents, initial filename, and expected preprocessor output.

libshaderc/include/shaderc.hpp
```diff
@@ -125,6 +126,48 @@ class CompileOptions {
...
+  // CompileOptions and will be destroyed with CompileOptions.
+  // TODO: File inclusion needs to be context-aware.
+  // e.g.
+  //  In file: /path/to/main_shader.vert:
+  //  #include "include/a"
+  //  In file: /path/to/include/a":
+  //  #include "b"
+  //  When compiling /path/to/main_shader.vert, the compiler should be able to
+  //  go to /path/to/include/b to find the file b.
+  //  This needs context info from compiler to client includer, supporint needs
```
dneto0: supporint ?

AWoloszyn: support needs interface changes?

libshaderc/src/shaderc.cc
```diff
@@ -85,21 +85,56 @@ struct {
...
-    return std::make_pair<std::string, std::string>(
-        "", "unexpected include directive");
+    if (!AreValidCallbacks())
+      return std::make_pair<std::string, std::string>(
+          "", "unexpected include directive");
+    shaderc_includer_response* data =
+        GetIncluderResponse_(user_data_, filename);
+    std::pair<std::string, std::string> entry =
```
dneto0: could have used "auto", but this is fine.

libshaderc/src/shaderc.cc
```diff
@@ -85,21 +85,56 @@ struct {
...
+        std::make_pair<std::string, std::string>(
+            std::string(data->path, data->path_length),
+            std::string(data->content, data->content_length));
+    ReleaseIncluderResponse_(user_data_, data);
+    return entry;
   }
+
+  const shaderc_includer_response_get_fn GetIncluderResponse_;
```
dneto0: style: members should_use_underscores_and_end_with_underscore_

libshaderc/src/shaderc.cc
```diff
@@ -85,21 +85,56 @@ struct {
...
+  const shaderc_includer_response_release_fn ReleaseIncluderResponse_;
+  void* user_data_;
 };
 
 }  // anonymous namespace
 
 struct shaderc_compile_options {
+  shaderc_compile_options(){};
   shaderc_util::Compiler compiler;
+  shaderc_includer_response_get_fn GetIncluderResponse;
```
dneto0: style: struct member names are like class member names, but without trailing underscore.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,123 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+                        testing::ValuesIn(std::vector<IncluderTestCase>{
+                            IncluderTestCase(
+                                // Fake file system.
+                                {
+                                    {"root",
+                                     "void foo() {}\n"
+                                     "#include \"path/to/file_1\"\n"},
+                                    {"path/to/file_1", "content of file_1\n"},
+                                },
+                                // Expected outeput.
```
dneto0: "output"

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,123 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
               HasSubstr("void main(){ }"));
 }
 
+// To test file inclusion, use hashmap as a fake file system to store fake files
```
dneto0: "use a hashmap"

dneto0: Actually, it's an unordered map. :-)

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,123 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+// to be included.
+using FakeFS = std::unordered_map<std::string, std::string>;
+
+// A includer test case needs: 1) A fake file system which is actually an
+// unordered_map, so that we can resolve the content given a file path. 2) An
```
dneto0: "A string"

dneto0: Also, should describe the special role of the "root" file. That way you don't need to (and shouldn't) later in the test code.

libshaderc/src/shaderc_cpp_test.cc
dneto0: Should say how the unordered map describes a filesystem:    The unordered map represents a filesystem by  mapping  filename string to the contents of that file as a string.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,123 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+TEST_P(IncluderTests, SetIncluder) {
+  const IncluderTestCase& test_case = GetParam();
+  const FakeFS& fs = test_case.fake_fs();
+  // Compilation is always started on 'root' file.
+  const std::string& shader = fs.at("root");
+  shaderc::Compiler compiler;
+  CompileOptions options;
+  // Sets includer instance.
+  options.SetIncluder(std::unique_ptr<TestIncluder>(new TestIncluder(fs)));
+  // Sets the compiler to preprocessing only mode.
```
dneto0: These last two comments are redundant. Remove them.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -504,6 +505,123 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+  options.SetPreprocessingOnlyMode();
+  const shaderc::SpvModule module = compiler.CompileGlslToSpv(
+      shader.c_str(), shaderc_glsl_vertex_shader, options);
+  // Checks the existence of the expected string.
+  EXPECT_THAT(module.GetData(), HasSubstr(test_case.expected_substring()));
+}
+
+TEST_P(IncluderTests, SetIncluderClonedOptions) {
```
dneto0: This looks like it's a dead test: not referenced anywhere.

dneto0: Nope. I was wrong.  I've been misreading the googletest docs w.r.t. parameterized tests.

libshaderc/src/shaderc_test.cc
```diff
@@ -482,6 +483,149 @@ TEST_F(CompileStringWithOptionsTest, PreprocessingOnlyOption) {
...
+  // Get path and content from the fake file system.
+  shaderc_includer_response* GetInclude(const char* filename) {
+    responses_.emplace_back(shaderc_includer_response{
+        filename, strlen(filename), fake_fs_.at(std::string(filename)).c_str(),
+        fake_fs_.at(std::string(filename)).size()});
+    return &responses_.back();
+  }
+
+  // Response data is owned as private property, no need to release explicitly.
+  void ReleaseInclude(shaderc_includer_response* data) { (void)data; };
```
dneto0: just drop the parameter name.  Then you can just write this:
```
void ReleaseInclude(shaderc_includer_response*) { }
```

(Also, the trailing semicolon is redundant and should be removed.)

dneto0:
+2
I'll squash and merge

dneto0:
Merged into master with 8c3684eda

PR #89, opened 2016-01-18
Set deducer

Qining:
Have something I feel unsure.

### DIFF
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -24,12 +24,26 @@ extern "C" {
 #include <stdint.h>
 
 typedef enum {
+  // Forced shader kinds. These shader kinds force the compiler to compile the
+  // source code as the specified kind of shader.
   shaderc_glsl_vertex_shader,
   shaderc_glsl_fragment_shader,
   shaderc_glsl_compute_shader,
   shaderc_glsl_geometry_shader,
   shaderc_glsl_tess_control_shader,
-  shaderc_glsl_tess_evaluation_shader
+  shaderc_glsl_tess_evaluation_shader,
+  // Deduce the shader kind from #pragma annotation in the source code. Compiler
+  // will emit error if #pragma annotation is not found.
+  shaderc_glsl_infer_from_source,
+  // Default shader kinds. Compiler will fall back to compile the source code as
+  // the specified kind of shader when #pragma annotation is not found in the
+  // source code.
+  shaderc_glsl_default_vertex_shader,
+  shaderc_glsl_default_fragment_shader,
+  shaderc_glsl_default_compute_shader,
+  shaderc_glsl_default_geometry_shader,
+  shaderc_glsl_default_tess_control_shader,
+  shaderc_glsl_default_tess_evaluation_shader,
 } shaderc_shader_kind;
 
 typedef enum {
@@ -221,18 +235,24 @@ void shaderc_compile_options_set_warnings_as_errors(
 // An opaque handle to the results of a call to shaderc_compile_into_spv().
 typedef struct shaderc_spv_module* shaderc_spv_module_t;
 
-// Takes a GLSL source string and the associated shader type, compiles
-// it according to the given additional_options. By default the source
-// string will be compiled into SPIR-V binary and a shaderc_spv_module will
-// be returned to hold the results of the compilation. When disassembly
-// mode or preprocessing only mode is enabled in the additional_options,
-// the source string will be compiled into char strings and held by the
-// returned shaderc_spv_module.  The entry_point_name null-terminated
-// string defines the name of the entry point to associate with this
-// GLSL source. If the additional_options parameter is not NULL, then the
-// compilation is modified by any options present. May be safely called
-// from multiple threads without explicit synchronization. If there was
-// failure in allocating the compiler object NULL will be returned.
+// Takes a GLSL source string and the associated shader kind, compiles it
+// according to the given additional_options. If the shader kind is not set
+// to a specified kind, but shaderc_glslc_infer_from_source, the compiler
+// will try to deduce the shader kind from the source string and a failure
+// in deducing will generate an error. Currently only #pragma annotation is
+// supported. If the shader kind is set to one of the default shader kinds,
+// the compiler will fall back to the default shader kind in case it failed to
+// deduce the shader kind from source string.
+// By default the source string will be compiled into SPIR-V binary
+// and a shaderc_spv_module will be returned to hold the results of the
+// compilation. When disassembly mode or preprocessing only mode is enabled
+// in the additional_options, the source string will be compiled into char
+// strings and held by the returned shaderc_spv_module.  The entry_point_name
+// null-terminated string defines the name of the entry point to associate
+// with this GLSL source. If the additional_options parameter is not NULL,
+// then the compilation is modified by any options present. May be safely
+// called from multiple threads without explicit synchronization. If there
+// was failure in allocating the compiler object NULL will be returned.
 shaderc_spv_module_t shaderc_compile_into_spv(
     const shaderc_compiler_t compiler, const char* source_text,
     size_t source_text_size, shaderc_shader_kind shader_kind,
--- a/libshaderc/include/shaderc.hpp
+++ b/libshaderc/include/shaderc.hpp
@@ -243,6 +243,14 @@ class Compiler {
   // Compiles the given source GLSL into a SPIR-V module.
   // The source_text parameter must be a valid pointer.
   // The source_text_size parameter must be the length of the source text.
+  // The shader_kind parameter either forces the compilation to be done with a
+  // specified shader kind, or hint the compiler how to determine the exact
+  // shader kind. If the shader kind is set to shaderc_glslc_infer_from_source,
+  // the compiler will try to deduce the shader kind from the source string and
+  // a failure in this proess will generate an error. Currently only #pragma
+  // annotation is supported. If the shader kind is set to one of the default
+  // shader kinds, the compiler will fall back to the specified default shader
+  // kind in case it failed to deduce the shader kind from the source string.
   // The compilation is done with default compile options.
   // It is valid for the returned SpvModule object to outlive this compiler
   // object.
@@ -256,13 +264,21 @@ class Compiler {
   // Compiles the given source GLSL into a SPIR-V module.
   // The source_text parameter must be a valid pointer.
   // The source_text_size parameter must be the length of the source text.
+  // The shader_kind parameter either forces the compilation to be done with a
+  // specified shader kind, or hint the compiler how to determine the exact
+  // shader kind. If the shader kind is set to shaderc_glslc_infer_from_source,
+  // the compiler will try to deduce the shader kind from the source string and
+  // a failure in this proess will generate an error. Currently only #pragma
+  // annotation is supported. If the shader kind is set to one of the default
+  // shader kinds, the compiler will fall back to the specified default shader
+  // kind in case it failed to deduce the shader kind from the source string.
   // The compilation is passed any options specified in the CompileOptions
   // parameter.
   // It is valid for the returned SpvModule object to outlive this compiler
   // object.
   // Note when the options_ has disassembly mode or preprocessing only mode set
-  // on, the returned SpvModule will hold a text string, instead of a SPIR-V binary
-  // generated with default options.
+  // on, the returned SpvModule will hold a text string, instead of a SPIR-V
+  // binary generated with default options.
   SpvModule CompileGlslToSpv(const char* source_text, size_t source_text_size,
                              shaderc_shader_kind shader_kind,
                              const CompileOptions& options) const {
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -36,8 +36,11 @@
 
 namespace {
 
-// Returns shader stage (ie: vertex, fragment, etc.) corresponding to kind.
-EShLanguage GetStage(shaderc_shader_kind kind) {
+// Returns shader stage (ie: vertex, fragment, etc.) in response to forced
+// shader kinds. If the shader kind is not a forced kind, returns EshLangCount
+// to let #pragma annotation or shader stage deducer determine the stage to
+// use.
+EShLanguage GetForcedStage(shaderc_shader_kind kind) {
   switch (kind) {
     case shaderc_glsl_vertex_shader:
       return EShLangVertex;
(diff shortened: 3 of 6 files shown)


libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -71,6 +71,91 @@ const char kTwoWarningsShader[] =
...
+    "}";
+
+// Vertex only shader with invalid #pragma annotation.
+const char kVertexOnlyShaderWithInvalidPragma[] =
+    "#version 310 es\n"
+    "#pragma shader_stage(fragment)\n"
+    "void main() {\n"
+    "    gl_Position = vec4(1.);\n"
+    "}";
+
```
Qining: All of these shaders are the same in c and cpp tests, should we pull them out?

libshaderc/src/shaderc_test.cc
```diff
@@ -591,7 +781,7 @@ TEST_P(IncluderTests, SetIncluderCallbacksClonedOptions) {
               HasSubstr(test_case.expected_substring()));
 }
 
-INSTANTIATE_TEST_CASE_P(CppInterface, IncluderTests,
+INSTANTIATE_TEST_CASE_P(CompileStringTest, IncluderTests,
```
Qining: Fix naming mistake from set-includer commits: 8c3684eda012165b026ca4597ca76319a4c8ab31.

libshaderc/include/shaderc.hpp
```diff
@@ -219,6 +219,20 @@ class CompileOptions {
...
+  // Sets the weak shader kind to be used for following compilation. The
+  // shader kind set here has lowest priority compared with the
+  // forced_shader_kind passed through shaderc_compile_into_spv()
+  // and #pragma annotation in source text. The compiler will check
+  // the forced_shader_kind first. If the forced_shader_kind is
+  // shaderc_glsl_none_shader, the compiler will check the #pragma
+  // annotation in the source text. If the #pragma annotation is not set,
+  // the compiler will use the shader kind set by this function to compile
+  // the source text.
+  void SetWeakShaderKind(shaderc_shader_kind kind) {
```
AWoloszyn: I think this would be better as a parameter to CompilerGlslToSpv, (and likewise shaderc_compile_into_spv) rather than on the compile_options object. (Meaning the shaderc_compiler_into_spv could turn into 
```
shaderc_spv_module_t shaderc_compile_into_spv(
const shaderc_compiler_t compiler,
const char* source_text,
size_t source_text_size,
shaderc_shader_kind shader_kind,

bool override_pragmas, // <<<<<

const char* entry_point_name,
const shaderc_compile_options_t additional_options);
```

That was compile_options become much more easily re-used for (as an example) compiling all shaders in a pipeline with the same options.

dneto0: We discussed this in person, and decided to expand the shaderc_shader_kind so it has:
 - one section of values with plain names than mean "force vertex shader", for example
 - one value of "get shader kind from pragma annotation"
 - another section of values such as "default to vertex", for example.

That way you have *one* place/argument to specify all the interacting priorities.  (This avoids weird interactions like what does override_pragmas true mean with shader kind none?  Error?)
Also, the simplest-named ones are presented first and have the force semantics, so they are the easiest to use.

libshaderc/include/shaderc.h
```diff
@@ -24,12 +24,26 @@ extern "C" {
...
 typedef enum {
-  shaderc_glsl_vertex_shader,
-  shaderc_glsl_fragment_shader,
-  shaderc_glsl_compute_shader,
-  shaderc_glsl_geometry_shader,
-  shaderc_glsl_tess_control_shader,
-  shaderc_glsl_tess_evaluation_shader
+  // Forced shader kinds. These shader kinds force the compiler to compile the
+  // source code as the specified kind of shader.
+  shaderc_glsl_forced_vertex_shader,
```
dneto0: I thought we agreed these would keep the simple names like "shaderc_glsl_vertex_shader".
But have the forced semantics.  The caller is being explicit about what they want, so we should consider it forced.

libshaderc/include/shaderc.h
```diff
@@ -24,12 +24,26 @@ extern "C" {
...
+  shaderc_glsl_forced_fragment_shader,
+  shaderc_glsl_forced_compute_shader,
+  shaderc_glsl_forced_geometry_shader,
+  shaderc_glsl_forced_tess_control_shader,
+  shaderc_glsl_forced_tess_evaluation_shader,
+  // Deduce the shader kind from #pragma annotation in the source code. Compiler
+  // will emit error if #pragma annotation is not found.
+  shaderc_glsl_deduce_from_pragma,
```
dneto0: Alternately, could have called this "shaderc_glsl_infer_from_source" which is more vague about how it's done.
I'd like Andrew's opinion on this.

libshaderc/src/shaderc.cc
```diff
@@ -85,6 +96,60 @@ struct {
...
+      case shaderc_glsl_forced_vertex_shader:
+      case shaderc_glsl_forced_fragment_shader:
+      case shaderc_glsl_forced_compute_shader:
+      case shaderc_glsl_forced_geometry_shader:
+      case shaderc_glsl_forced_tess_control_shader:
+      case shaderc_glsl_forced_tess_evaluation_shader:
+        return EShLangCount;
+    }
+    assert(0 && "Unhandled shaderc_shader_kind");
+    return EShLangVertex;
```
dneto0: May as well make this dead one return EShLangCount

libshaderc/src/shaderc.cc
```diff
@@ -85,6 +96,60 @@ struct {
...
+                    << "' shader kind not determined" << std::endl;
+    }
+    return stage;
+  };
+  StageDeducer() : kind_(shaderc_glsl_deduce_from_pragma){};
+
+ private:
+  EShLanguage GetDefaultStage(shaderc_shader_kind kind) const {
+    switch (kind) {
+      case shaderc_glsl_default_vertex_shader:
```
dneto0: Might read more nicely if the cases were presented in the same order as the enum definitions.
You still get the nice fall-through behaviour.

libshaderc/include/shaderc.h
AWoloszyn: I would agree. I think 
shaderc_compile_into_spv(...., shaderc_glsl_vertex_shader, ..) reads fairly clearly as "I want this to be a vertex shader"

libshaderc/src/shaderc.cc
```diff
@@ -85,6 +96,60 @@ struct {
...
+  EShLanguage operator()(std::ostream* error_stream,
+                         const shaderc_util::string_piece& error_tag) const {
+    EShLanguage stage = GetDefaultStage(kind_);
+    if (stage == EShLangCount) {
+      *error_stream << "libshaderc error: '" << error_tag
+                    << "' shader kind not determined" << std::endl;
+    }
+    return stage;
+  };
+  StageDeducer() : kind_(shaderc_glsl_deduce_from_pragma){};
```
dneto0: Put all the constructors together.

libshaderc/include/shaderc.h
AWoloszyn: I would be ok with shaderc_glsl_deduce_from_pragma (or infer_from_pragma).
If we are too vague, then the user may expect it to deduce based on things that aren't a pragma.

For example from the use of gl_VertexID one could infer that this is a vertex shader, and gl_PatchVerticesIn could point to a tessc shader, but if we say "we only infer from pragmas" then we have a reasonable expectation that even in the presence of those other stage specific bits, we only look at the pragmas.

libshaderc/include/shaderc.h
Qining: But then how does the compiler know the shader kind, if the arg is set to shaderc_infer_from_source (or shaderc_glsl_deduce_from_pragma) and #pragma is not found?
We won't have any hints about which shader kind to use in this case, unless we make the deducer callback exposed to client again. Libshaderc doesn't do the job to infer the shader kind from file name, I think.

libshaderc/src/shaderc.cc
```diff
@@ -85,6 +96,60 @@ struct {
...
+// the compiler should emit error if #pragma annotation is not found in this
+// case.
+class StageDeducer {
+ public:
+  StageDeducer(shaderc_shader_kind kind) : kind_(kind){};
+  EShLanguage operator()(std::ostream* error_stream,
+                         const shaderc_util::string_piece& error_tag) const {
+    EShLanguage stage = GetDefaultStage(kind_);
+    if (stage == EShLangCount) {
+      *error_stream << "libshaderc error: '" << error_tag
```
AWoloszyn: I don't see any examples elsewhere, but we probably don't have to specify that it is a libshaderc error something like

*error_stream << "error: '" << error_tag << "' could not determine the shader stage"

libshaderc/src/shaderc_cpp_test.cc
AWoloszyn: I would be fine with a common header that declares all of these for the tests.

libshaderc/src/shaderc.cc
dneto0: I think the style guide allows a default argument in this case.  It might improve readability.
Coder's choice.

libshaderc/include/shaderc.h
Qining: Oops, I see, you mean to rename the forced ones to shaderc_glsl_<xxxx>_shader... I thought you want me to remove the default ones. Sorry, my mistake. So we should rename the forced ones to shaderc_glsl_<xxx>_shader, and keep the naming of the default ones?

libshaderc/src/shaderc_cpp_test.cc
dneto0: Duplicating them keeps it simple and isolated, less indirection.  
Coder's choice.  
But also, if you do commonize the data, avoid static-duration class objects.  Only POD allowed.

Others might disagree whether DRY would be better for this.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -499,12 +584,102 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
 }
 
+// A shader kind test cases needs: 1) A shader text with or without #pragma
+// annotation, 2) forced_shader_kind, 3) weak_shader_kind.
+struct ShaderKindTestCase {
+  const char* shader_;
+  shaderc_shader_kind shader_kind_;
+};
+
+// Test the shader kind deduction process. If the shader kind is one of the
```
dneto0: Much of this blurb belongs on the CompilesToValidSpv method.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -499,12 +584,102 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
...
+    ShaderKindTestCase test_case = GetParam();
+    return IsValidSpv(compiler_.CompileGlslToSpv(
+        test_case.shader_, test_case.shader_kind_, options_));
+  }
+
+  shaderc::Compiler compiler_;
+  CompileOptions options_;
+
+ private:
+  static bool IsValidSpv(const shaderc::SpvModule& result) {
```
dneto0: Maybe this should just be a free function in the module.   It's not really coupled to the internals of the class.

Also, the method needs a blurb.

libshaderc/include/shaderc.h
AWoloszyn: Sorry, I think the full list should be

shaderc_glsl_vertex_shader,
shaderc_glsl_fragment_shader,
shaderc_glsl_compute_shader,
shaderc_glsl_geometry_shader,
shaderc_glsl_tess_control_shader,
shaderc_glsl_tess_evaluation_shader,
shaderc_glsl_deduce_from_pragma,
shaderc_glsl_default_vertex_shader,
shaderc_glsl_default_fragment_shader,
shaderc_glsl_default_compute_shader,
shaderc_glsl_default_geometry_shader,
shaderc_glsl_default_tess_control_shader,
shaderc_glsl_default_tess_evaluation_shader,

libshaderc/src/shaderc_test.cc
```diff
@@ -479,10 +564,100 @@ TEST_F(CompileStringWithOptionsTest, PreprocessingOnlyOption) {
...
+// source code and emit error if none such annotation is found. When the shader
+// kind is one of the default ones, the compiler will fall back to use the
+// specified shader kind if and only if #pragma annoation is not found.
+class ShaderKindDeductionTests
+    : public testing::TestWithParam<ShaderKindTestCase> {
+ public:
+  ShaderKindDeductionTests() : options_(shaderc_compile_options_initialize()) {}
+
+ protected:
+  bool CompilesToValidSpv() {
```
dneto0: Blurb

libshaderc/src/shaderc_test.cc
dneto0: +1

libshaderc/include/shaderc.h
AWoloszyn: +1

dneto0: Yes, what Andrew said. That's more what I expect and want.

Separately:  I like *_infer_from_source better than *_deduce_from_pragma 
For example, in future we might want to support a different source language (non GLSL) that has a more graceful way of indicating shader type in the shader source text.

libshaderc/include/shaderc.h
dneto0: With *_infer_from_source you get extra flexibility in case you want to support other languages that have nicer ways of specifying shader stage than pragma.
Set expectations about what is inferrable via documentation.?

dneto0:
Done review.

libshaderc/include/shaderc.h
AWoloszyn: I do not have a strong opinion on this one, and yes this does allow better use of other source languages in the future, so +1 to infer_from_source.

libshaderc/include/shaderc.h
dneto0: Another thought.  Maybe we can have both enum names.
*_infer_from_source can internally be translated to whatever set of more specific options alllow you to infer from the source, whether it's pragmas or other language features.
This could be done in a subsequent CL.
So let's say Coder's Choice for using the enum name based on pragmas.

libshaderc/src/shaderc.cc
```diff
@@ -85,6 +96,60 @@ struct {
...
+// a forced shader kind, this functor won't be called and it simply returns
+// EShLangCount to make the syntax correct. When the shader kind is set to
+// shaderc_glsl_deduce_from_pragma, this functor also returns EShLangCount, but
+// the compiler should emit error if #pragma annotation is not found in this
+// case.
+class StageDeducer {
+ public:
+  StageDeducer() : kind_(shaderc_glsl_infer_from_source){};
+  StageDeducer(shaderc_shader_kind kind) : kind_(kind){};
+  EShLanguage operator()(std::ostream* error_stream,
```
dneto0: Need a blurb. :-)

libshaderc/src/shaderc.cc
```diff
@@ -85,6 +96,60 @@ struct {
...
+    EShLanguage stage = GetDefaultStage(kind_);
+    if (stage == EShLangCount) {
+      *error_stream << "error: '" << error_tag
+                    << "' could not determine the shader stage" << std::endl;
+    }
+    return stage;
+  };
+
+ private:
+  EShLanguage GetDefaultStage(shaderc_shader_kind kind) const {
```
dneto0: Need a blurb.

dneto0:
Just need 2 blurbs on methods.  Nearly there. :-)

libshaderc/src/shaderc.cc
Qining: Added in the commit 26e9f46. Will work on rebasing now.

Qining: Updated to 5dc7d71.

dneto0:
+2.  I'll rebase, squash and merge

dneto0:
Ok.  @Qining agreed to rebase.  It was more work than I expected. :-)

dneto0:
Squashed into one commit and pushed onto master as 3f48624006

PR #90, opened 2016-01-20
Fix mismatch of builtins in preprocess vs. compile

dneto0:
The fix is to:
 - Reinitialize Glslang for the preprocessing step used to
   deduce the shader stage.
 - When doing that preprocessing, combine the "message rules"
   with the just-preprocess rule.

Test profile-specific shaders vs. specified target environment.

### DIFF
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -71,7 +71,7 @@ const char kTwoWarningsShader[] =
 
 // A shader that compiles under OpenGL compatibility profile rules,
 // but not OpenGL core profile rules.
-const char kOpenGLCompatibilityShader[] =
+const char kOpenGLCompatibilityFragmentShader[] =
     R"(#version 100
      uniform highp sampler2D tex;
      void main() {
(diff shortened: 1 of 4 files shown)


libshaderc/src/shaderc_test.cc
```diff
@@ -69,6 +69,20 @@ const char kTwoWarningsShader[] =
...
+// but not OpenGL core profile rules.
+const char kOpenGLCompatibilityShader[] =
+    R"(#version 100
+     uniform highp sampler2D tex;
+     void main() {
+       gl_FragColor = texture2D(tex, vec2(0.0,0.0));
+     })";
+
+// A shader that compiles under OpenGL core profile rules.
+const char kOpenGLVertexShader[] =
```
dekimir: Nit: stage is a part of the variable's name here but not on line 74.

AWoloszyn:
+1 from me, good to go when @dekimir  is happy.

libshaderc/src/shaderc_test.cc
```diff
@@ -725,32 +739,74 @@ TEST_F(CompileStringWithOptionsTest, IfDefCompileOption) {
...
+}
+
+TEST_F(CompileStringWithOptionsTest, TargetEnvIgnoredWhenPreprocessing) {
+  shaderc_compile_options_set_preprocessing_only_mode(options_.get());
+
+  EXPECT_TRUE(CompilationSuccess(kOpenGLCompatibilityShader,
+                                 shaderc_glsl_fragment_shader, options_.get()));
+
+  shaderc_compile_options_set_target_env(options_.get(),
+                                         shaderc_target_env_opengl_compat, 0);
```
dekimir: Nit: weird indent.

dekimir:
+2 I'm already happy, minor nitpicks notwithstanding.

libshaderc/src/shaderc_test.cc
dneto0: It's what clang-format gives me.  \o/

dneto0:
All feedback addressed

dneto0:
I'll assume @dekimir 's +2 still stands as I've addressed his comments.
I'll rebase, squash and merge.

dneto0:
Merged into master as 602f4b32

PR #93, opened 2016-01-20
Add new glslang source files to Android.mk.

dekimir:


third_party/Android.mk
```diff
@@ -9,6 +9,7 @@ LOCAL_CXXFLAGS:=-std=c++11 $(GLSLANG_OS_FLAGS)
 LOCAL_EXPORT_C_INCLUDES:=$(GLSLANG_LOCAL_PATH)
 LOCAL_SRC_FILES:= \
 	SPIRV/GlslangToSpv.cpp \
+	SPIRV/InReadableOrder.cpp \
 	SPIRV/SpvBuilder.cpp \
 	SPIRV/SPVRemapper.cpp \
```
AWoloszyn: Unrelated to this change, so I won't ask you to fix it, but SPVRemapper should go above SpvBuilder, OR if we decide that case does not matter, doc.* should go at the top.

AWoloszyn:
+2 Looks good to me.

PR #96, opened 2016-01-21
Add error type field result module

Qining:
The StageDeducer in libshaderc now keeps a flag, set to true when it is called to deduce a shader stage but failed (the last chance to get the stage). We then check the flag after compilation, to determine the error type, and store it in the result module

Replace get_success() with get_compilation_result(). Instead of returning a boolean, it returns an enum type: shaderc_compilation_result. The enum value can indicate whether the compilation succeeded or failed due to some reasons.

### DIFF
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -63,12 +63,11 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
-// Used in result module (shaderc_spv_module) to tell the type of errors
+// Used in the result module (shaderc_spv_module) to tell the type of errors
 // generated during compilation.
 typedef enum {
   shaderc_none_error,
-  shaderc_shader_kind_error,  // errors due to failure in shader stage
-                              // deduction.
+  shaderc_shader_kind_error,  // error in the deduction of shader stage
   shaderc_compilation_error,
 } shaderc_error_type;
 
@@ -292,8 +291,8 @@ size_t shaderc_module_get_num_warnings(const shaderc_spv_module_t module);
 // Returns the number of errors generated during the compilation.
 size_t shaderc_module_get_num_errors(const shaderc_spv_module_t module);
 
-// Returns the type of the generated compilation errors. Returns
-// shaderc_none_error if there isn't error or result module doesn't exist.
+// Returns the type of the errors of one compilation. Returns shaderc_none_error
+// if there isn't any error.
 shaderc_error_type shaderc_module_get_error_type(const shaderc_spv_module_t);
 
 // Returns a pointer to the start of the SPIR-V bytes, either SPIR-V binary or
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -66,10 +66,11 @@ typedef enum {
 // Used in the result module (shaderc_spv_module) to tell the type of errors
 // generated during compilation.
 typedef enum {
-  shaderc_none_error,
-  shaderc_shader_kind_error,  // error in the deduction of shader stage
-  shaderc_compilation_error,
-} shaderc_error_type;
+  shaderc_compilation_result_success,
+  shaderc_compilation_result_failed_invalid_stage,  // error in the deduction of shader stage
+  shaderc_compilation_result_compilation_failed,
+  shaderc_compilation_result_null_result_module,
+} shaderc_compilation_result;
 
 // Usage examples:
 //
@@ -291,9 +292,11 @@ size_t shaderc_module_get_num_warnings(const shaderc_spv_module_t module);
 // Returns the number of errors generated during the compilation.
 size_t shaderc_module_get_num_errors(const shaderc_spv_module_t module);
 
-// Returns the type of the errors of one compilation. Returns shaderc_none_error
-// if there isn't any error.
-shaderc_error_type shaderc_module_get_error_type(const shaderc_spv_module_t);
+// Returns the compilation result, indicating whether the compilation succeeded,
+// or failed due to some reasons, like invalid shader stage or compilation
+// errors.
+shaderc_compilation_result shaderc_module_get_compilation_result(
+    const shaderc_spv_module_t);
 
 // Returns a pointer to the start of the SPIR-V bytes, either SPIR-V binary or
 // char string. When the source string is compiled into SPIR-V binary, this is
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -63,6 +63,15 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
+// Used in result module (shaderc_spv_module) to tell the type of errors
+// generated during compilation.
+typedef enum {
+  shaderc_none_error,
+  shaderc_shader_kind_error,  // errors due to failure in shader stage
+                              // deduction.
+  shaderc_compilation_error,
+} shaderc_error_type;
+
 // Usage examples:
 //
 // Aggressively release compiler resources, but spend time in initialization
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -292,6 +292,10 @@ size_t shaderc_module_get_num_warnings(const shaderc_spv_module_t module);
 // Returns the number of errors generated during the compilation.
 size_t shaderc_module_get_num_errors(const shaderc_spv_module_t module);
 
+// Returns the type of the generated compilation errors. Returns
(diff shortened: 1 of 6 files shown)


libshaderc/src/shaderc_test.cc
```diff
@@ -400,38 +447,6 @@ TEST_F(CompileStringWithOptionsTest, GenerateDebugInfoDisassembly) {
               HasSubstr("debug_info_sample"));
 }
 
-TEST_F(CompileStringWithOptionsTest, GetNumErrors) {
```
Qining: Move these tests to 'CompileStringTest' group, as they are not using 'options'

libshaderc/include/shaderc.h
```diff
@@ -63,6 +63,15 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
+// Used in result module (shaderc_spv_module) to tell the type of errors
+// generated during compilation.
+typedef enum {
+  shaderc_none_error,
+  shaderc_shader_kind_error,  // errors due to failure in shader stage
```
AWoloszyn: shaderc_stage_deduction_error
or maybe
shaderc_unknown_stage_error

libshaderc/include/shaderc.h
```diff
@@ -63,6 +63,15 @@ typedef enum {
...
+                              // deduction.
+  shaderc_compilation_error,
+} shaderc_error_type;
+
 // Usage examples:
 //
```
AWoloszyn: Might be best to re-name the enum, 
shaderc_none_error makes me sad.

typedef enum {
   shaderc_compilation_result_success,
   shaderc_compilation_result_failed_invalid_stage,
   shaderc_compilation_result_compilation_failed
} shaderc_compilation_result;

libshaderc/include/shaderc.h
```diff
@@ -283,6 +292,10 @@ size_t shaderc_module_get_num_warnings(const shaderc_spv_module_t module);
 // Returns the number of errors generated during the compilation.
 size_t shaderc_module_get_num_errors(const shaderc_spv_module_t module);
 
+// Returns the type of the generated compilation errors. Returns
```
AWoloszyn: This could be combined with shaderc_module_get_success into shaderc_module_get_compilation_result.

libshaderc/include/shaderc.h
Qining: Actually I was just typing to say we may need new names here. These looks much better.

libshaderc/include/shaderc.h
Qining: Agree, updated like you said in eb74082.

libshaderc/include/shaderc.h
```diff
@@ -63,6 +63,15 @@ typedef enum {
...
 } shaderc_profile;
 
+// Used in the result module (shaderc_spv_module) to tell the type of errors
+// generated during compilation.
+typedef enum {
+  shaderc_compilation_result_success,
+  shaderc_compilation_result_failed_invalid_stage,  // error in the deduction of shader stage
+  shaderc_compilation_result_compilation_failed,
+  shaderc_compilation_result_null_result_module,
+} shaderc_compilation_result;
```
Qining: Old GetSuccess() (the cpp api) checks the existence of a result module, and return false if there isn't. So need to add this one of the compilation results.

libshaderc/include/shaderc.h
```diff
@@ -63,6 +63,15 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
+// Used in the result module (shaderc_spv_module) to tell the type of errors
+// generated during compilation.
```
dneto0: Should be shader_compilation_status_* instead of shader_compilation_result_*

A "result" sounds like a more heavyweight thing, as if it included the resulting binary.

Precedent is OpenCL CL_PROGRAM_BUILD_STATUS query from clGetProgramBuildInfo.
Also, COMPILE_STATUS query from GL's GetShaderiv query.

libshaderc/include/shaderc.h
```diff
@@ -283,6 +292,12 @@ size_t shaderc_module_get_num_warnings(const shaderc_spv_module_t module);
 // Returns the number of errors generated during the compilation.
 size_t shaderc_module_get_num_errors(const shaderc_spv_module_t module);
 
+// Returns the compilation result, indicating whether the compilation succeeded,
```
dneto0: compilation *status*

libshaderc/include/shaderc.hpp
```diff
@@ -49,6 +44,16 @@ class SpvModule {
     return shaderc_module_get_error_message(module_);
   }
 
+  // Returns the compilation result, indicating whether the compilation
+  // succeeded, or failed due to some reasons, like invalid shader stage or
+  // compilation errors.
+  shaderc_compilation_result GetCompilationResult() const {
```
dneto0: rename to GetCompilationStatus

libshaderc/src/shaderc.cc
```diff
@@ -107,8 +107,8 @@ std::mutex compile_mutex;  // Guards shaderc_compile_*.
 // case.
 class StageDeducer {
  public:
-  StageDeducer() : kind_(shaderc_glsl_infer_from_source){};
-  StageDeducer(shaderc_shader_kind kind) : kind_(kind){};
+  StageDeducer() : kind_(shaderc_glsl_infer_from_source), error_(false){};
+  StageDeducer(shaderc_shader_kind kind) : kind_(kind), error_(false){};
```
dneto0: Using a default argument is looking better, as you can be more DRY.
But Coder's choice again.

libshaderc/src/shaderc.cc
```diff
@@ -116,16 +116,18 @@ class StageDeducer {
...
-                         const shaderc_util::string_piece& error_tag) const {
+  EShLanguage operator()(std::ostream* /*error_stream*/,
+                         const shaderc_util::string_piece& /*error_tag*/) {
     EShLanguage stage = GetDefaultStage(kind_);
     if (stage == EShLangCount) {
-      *error_stream << "error: '" << error_tag
-                    << "' could not determine the shader stage" << std::endl;
+      error_ = true;
     }
     return stage;
```
dneto0: Feels like we should set error_ = false in the ok case.
What is supposed to happen if glslang calls it twice with two different tags?   Ok, might not happen now, but we should make it well defined.

libshaderc/src/shaderc.cc
```diff
@@ -116,16 +116,18 @@ class StageDeducer {
   // 'default' shader kind to the corresponding shader stage. As this is the
   // last trial to determine the shader stage, failing to find the corresponding
```
dneto0: Update blurb to say " ... will record an error."

libshaderc/include/shaderc.h
```diff
@@ -63,6 +63,15 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
+// Used in the result module (shaderc_spv_module) to tell the type of errors
+// generated during compilation.
+typedef enum {
+  shaderc_compilation_result_success,
```
dneto0: explicitly set this to 0.  It's a useful pattern to have success always be 0, and error cases non-zero just like exit status from a process.
Declaring it explcitly to 0 means callers can rely on this behaviour.

libshaderc/src/shaderc_private.h
```diff
@@ -33,6 +33,8 @@ struct shaderc_spv_module {
   size_t num_errors = 0;
   // Number of warnings.
   size_t num_warnings = 0;
+  // Compilation result.
+  shaderc_compilation_result compilation_result;
```
dneto0: compilation_status
Also, explicitly set it to something.  Suggest shaderc_compilation_status_null_result_module.  :-)

dneto0:
-1 minor issues.
Naming, member initialization, one blurb tweak.

libshaderc/include/shaderc.h
```diff
@@ -63,6 +63,15 @@ typedef enum {
   shaderc_profile_es,
 } shaderc_profile;
 
+// Used in the result module (shaderc_spv_module) to indicate the status of an
```
dneto0: Strict stylists would say you should only describe the inherent meaning of the thing you're defining.  The definition should stand on its own.
Don't describe it by saying how it's used, since that will change over time. It gets out of date.
Ok for now.

dneto0:
+2
I'll squash and merge

dneto0:
Merged into master as 6e1cb55e9b86f29

PR #97, opened 2016-01-22
Parse version profile libshaderc

Qining:
Currently the test is not completed, I think we need more test cases.

Didn't add interface in CPP header, I think we don't need one for this method, do we?

### DIFF
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -461,6 +461,7 @@ bool shaderc_parse_version_profile(const char* str, int* version,
       return true;
     case ECompatibilityProfile:
       *profile = shaderc_profile_compatibility;
+      return true;
     case ENoProfile:
       *profile = shaderc_profile_none;
       return true;
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -468,4 +468,8 @@ bool shaderc_parse_version_profile(const char* str, int* version,
     case EBadProfile:
       return false;
   }
+
+  // Shouldn't reach here, all profile enum should be handled above.
+  // Be strict to return false.
+  return false;
 }
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -25,6 +25,7 @@
 
 #include "libshaderc_util/compiler.h"
 #include "libshaderc_util/resources.h"
+#include "libshaderc_util/version_profile.h"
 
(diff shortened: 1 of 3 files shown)


libshaderc/src/shaderc_test.cc
```diff
@@ -956,4 +956,48 @@ TEST_F(CompileKindsTest, TessEvaluation) {
...
+  shaderc_profile expected_profile;
+};
+
+// Test for a helper function to parse version and profile from string.
+using ParseVersionProfileTest =
+    testing::TestWithParam<ParseVersionProfileTestCase>;
+
+TEST_P(ParseVersionProfileTest, FromNullTerminatedString) {
+  const ParseVersionProfileTestCase& test_case = GetParam();
+  int version;
```
dneto0: good practice to initialize scalar non-class values. (version and profile in this case.)
Some compiler somewhere complain about later uses without initializations.  Just avoid that so we don't convert that warning to an error.

libshaderc/src/shaderc_test.cc
```diff
@@ -956,4 +956,48 @@ TEST_F(CompileKindsTest, TessEvaluation) {
...
+    EXPECT_EQ(test_case.expected_profile, profile);
+  }
+}
+
+INSTANTIATE_TEST_CASE_P(
+    HelperMethods, ParseVersionProfileTest,
+    testing::ValuesIn(std::vector<ParseVersionProfileTestCase>{
+        // Valid version profiles
+        {"450core", true, 450, shaderc_profile_core},
+        {"310es", true, 310, shaderc_profile_es},
```
dneto0: check a valid result with compatibility profile.

dneto0:
-1  Need another valid test case, and initialize scalars to avoid possible warning of use-before-def.

Qining:
Updated with a compatibility profile test and a empty string test.

dneto0:
+2
I'll squash and merge

libshaderc/src/shaderc.cc
```diff
@@ -443,3 +444,28 @@ void shaderc_get_spv_version(unsigned int* version, unsigned int* revision) {
...
+    case ECompatibilityProfile:
+      *profile = shaderc_profile_compatibility;
+      return true;
+    case ENoProfile:
+      *profile = shaderc_profile_none;
+      return true;
+    case EBadProfile:
+      return false;
+  }
+}
```
dneto0: GCC complains error: control reaches end of non-void function

dneto0:
Squashed and pushed into master as c73f444d05

PR #99, opened 2016-01-22
Return ICE instead of Compilation Failure if we catch and exception.

AWoloszyn:
Be more transparent about the cause of failure when we
catch and internal exception in the compiler.


dneto0:
Fixed a typo, then merged into master as b062432916

PR #100, opened 2016-01-22
Port glslc to use libshaderc' compilation interface

Qining:
1. Replace shader_stage.cc/.h with shader_kind.cc/.h, also replaced the tests for it
2. FileCompiler now inherits from shaderc::Compiler. It delegates an shaderc::CompilerOptions object to set modes, etc. 
3. Add disassembly_ and preprocessing_only_ flags to FileCompiler, as libshaderc is not supposed to maintain states.
4. Add a vector of string in main() to hold the macro definitions.
5. Updated the CMakeList.txt, glslc now depends on shaderc.
6. Modified FileIncluder uses libshaderc_util::FileFinder. But it only keeps the latest response, data needs to be copied to underlying glslang.
7. FileCompiler now is using std::cerr as the error stream.


glslc/src/file_compiler.cc
```diff
@@ -77,7 +91,29 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
+                   "specified ahead";
+    } else if (error_file_name == "<stdin>") {
+      std::cerr
+          << "glslc: error: '-': -fshader-stage required when input is from "
+             "standard "
+             "input \"-\"";
+    } else {
+      std::cerr << "glslc: error: "
+                << "'" << error_file_name << "': "
+                << "file not recognized: File format not recognized";
```
Qining: The error messages about 'invalid shader kind/stage' are moved to here.
Note that, at this point, we already transform '-' to '<stdin>' when the input is from standard input. The error message here is to meet the demand of current tests.

glslc/src/file_includer.cc
```diff
@@ -18,17 +18,37 @@
...
+        file_content_.size(),
+    };
   } else {
-    // TODO(deki): Emit error message.
-    return std::make_pair("", "");
+    char error_msg[] = "include file not found.";
+    file_content_.insert(file_content_.begin(), error_msg,
+                         error_msg + sizeof(error_msg) / sizeof(error_msg[0]));
+    response_ = {
+        filename, std::strlen(filename), file_content_.data(),
```
Qining: Passed the error message "include file not found" to glslang, and glslang will complain with this error message. But I think this may not be a good design.

glslc/src/file_compiler.h
```diff
@@ -40,7 +43,7 @@ class FileCompiler : public shaderc_util::Compiler {
   // Any errors/warnings found in the shader source will be output to std::cerr
```
dekimir: The blurb seems out of date now.

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,27 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
-      Compile(source_string, shader_stage, error_file_name.str(), deducer,
-              glslc::FileIncluder(&include_file_finder_), output_stream,
-              &std::cerr, &total_warnings_, &total_errors_);
 
-  if (!compilation_success && output_stream->fail()) {
+  std::unique_ptr<FileIncluder> includer(new FileIncluder(&include_file_finder_));
+  options_.SetIncluder(std::move(includer));
+
+  shaderc::SpvModule result =
+      CompileGlslToSpv(source_string.begin(), source_string.size(), shader_kind,
```
dekimir: It is not portable to pass `string::iterator` where a `char*` is expected.  Use `.c_str` instead of `.begin`.

dekimir: Never mind, this is `string_piece::begin()`, so that's OK.

glslc/src/main.cc
```diff
@@ -176,24 +185,43 @@ int main(int argc, char** argv) {
...
+
+        // We have to insert a null terminator at the end of name, as we are
+        // using libshaderc's API instead of libshaderc_util's. Libshaderc API
+        // assumes a const char* string, while libshaderc_util assume a
+        // string_piece one. The original implementation doesn't append '\0' but
+        // works fine with libshaderc_util because it has 'end_' property. But
+        // libshaderc api doesn't know that so the macro 'name' will include the
+        // 'value' if we don't insert '\0'.
+
+        macro_names.emplace_back(std::string(name_piece.data(), name_length));
```
dekimir: Delete `std::string`.

glslc/src/main.cc
```diff
@@ -176,24 +185,43 @@ int main(int argc, char** argv) {
...
+        macro_values.emplace_back(
+            std::string(name_length < argument_length
```
dekimir: Delete `std::string`.

glslc/src/main.cc
```diff
@@ -133,8 +138,12 @@ int main(int argc, char** argv) {
       compiler.SetWorkingDirectory(workdir.str());
     } else if (arg.starts_with("-fshader-stage=")) {
       const string_piece stage = arg.substr(std::strlen("-fshader-stage="));
-      force_shader_stage = glslc::GetShaderStageFromCmdLine(arg);
-      if (force_shader_stage == EShLangCount) {
+      // forced_shader_kind can be 1) one of the forced shader kinds, which
+      // means following files will be compiler with the specified shader
```
dekimir: Do you mean "compiled"?

glslc/src/shader_kind.h
```diff
@@ -12,34 +12,23 @@
...
-  // Any errors are written to error_stream.
-  EShLanguage operator()(std::ostream* error_stream,
-                         const shaderc_util::string_piece& error_tag) const;
-
- private:
-  shaderc_util::string_piece file_name_;
-};
-
-EShLanguage GetShaderStageFromCmdLine(
+// Parse the string piece from command line to get the force shader kind.
```
dekimir: Explain what happens under different conditions: parse errors, missing flags, etc.

glslc/src/shader_kind.h
```diff
@@ -12,34 +12,23 @@
...
+shaderc_shader_kind GetForcedShaderKindFromCmdLine(
     const shaderc_util::string_piece& f_shader_stage);
```
dekimir: Why is the argument called `f_shader_stage` if it's allegedly a command line?

glslc/src/main.cc
```diff
@@ -133,8 +138,12 @@ int main(int argc, char** argv) {
...
+      // kind; 2) shaderc_glsl_infer_from_source, which means 'arg' doesn't
```
dekimir: This is a weird result to get in case of an invalid arg.  Why not explicitly indicate an error?

glslc/src/shader_kind.h
dekimir: Upon closer inspection, the "command line" is a red herring -- this is about parsing text into `shaderc_shader_kind`.  Both the function and the argument need a better name.

glslc/src/shader_kind.cc
```diff
@@ -0,0 +1,76 @@
...
+}
+
+shaderc_shader_kind GetForcedShaderKindFromCmdLine(
+    const shaderc_util::string_piece& f_shader_stage) {
+  size_t equal_pos = f_shader_stage.find_first_of("=");
+  if (equal_pos == std::string::npos) return shaderc_glsl_infer_from_source;
+  return MapStageNameToForcedKind(f_shader_stage.substr(equal_pos + 1));
+}
+
+shaderc_shader_kind DeduceDefaultShaderKindFromFileName(string_piece file_name) {
```
dekimir: `const string_piece`?

glslc/src/file_compiler.h
```diff
@@ -69,12 +72,45 @@ class FileCompiler : public shaderc_util::Compiler {
   void SetIndividualCompilationMode();
 
   // Instead of outputting object files, output the disassembled textual output.
-  virtual void SetDisassemblyMode() override;
+  virtual void SetDisassemblyMode();
```
dekimir: Drop `virtual` and re-add `override`.

https://google.github.io/styleguide/cppguide.html#Inheritance

glslc/src/file_compiler.h
```diff
@@ -69,12 +72,45 @@ class FileCompiler : public shaderc_util::Compiler {
...
+
+  // Delegates CompilerOptions to add macro definiations.
```
dekimir: Typo "definitions".  Also not clear what is being delegated to whom.  Consider rewriting in simpler terms.

glslc/src/file_compiler.h
```diff
@@ -69,12 +72,45 @@ class FileCompiler : public shaderc_util::Compiler {
...
+  virtual void AddMacroDefinition(const char* name, const char* value) {
```
dekimir: What's wrong with the base class's `AddMacroDefinition`?

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,27 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
-  StageDeducer deducer(input_file);
-  bool compilation_success =
-      Compile(source_string, shader_stage, error_file_name.str(), deducer,
-              glslc::FileIncluder(&include_file_finder_), output_stream,
-              &std::cerr, &total_warnings_, &total_errors_);
 
-  if (!compilation_success && output_stream->fail()) {
+  std::unique_ptr<FileIncluder> includer(new FileIncluder(&include_file_finder_));
+  options_.SetIncluder(std::move(includer));
+
```
Qining: Use the new includer setter method in libshaderc. We have to create an includer for each file to be compiled, as we are passing unique_ptr.

glslc/src/file_compiler.cc
```diff
@@ -89,28 +125,29 @@ void FileCompiler::AddIncludeDirectory(const std::string& path) {
...
   if (!disassemble_) {
     needs_linking_ = false;
     file_extension_ = ".spv";
   }
 }
 
-void FileCompiler::SetDisassemblyMode() {
-  Compiler::SetDisassemblyMode();
+void FileCompiler::SetDisassemblyFlags() {
+  disassemble_ = true;
```
Qining: The original FileCompiler inherits from libshaderc_util::compiler, which has an disassemble_ flag, and it is set when libshaderc_util::compiler::SetDisassemblyMode() is called (that is why we can not see any code that sets disassemble_ in the original FileCompiler). But with libshaderc in between, we need to maintain the state flags here, we can not access libshaderc_util::Compiler anymore.

glslc/src/file_compiler.cc
```diff
@@ -89,28 +125,29 @@ void FileCompiler::AddIncludeDirectory(const std::string& path) {
...
   needs_linking_ = false;
   file_extension_ = ".s";
 }
 
-void FileCompiler::SetPreprocessingOnlyMode() {
-  Compiler::SetPreprocessingOnlyMode();
+void FileCompiler::SetPreprocessingOnlyFlags() {
+  preprocess_only_ = true;
```
Qining: Same as disassemble_ flag.

glslc/src/stage_test.cc
```diff
@@ -27,63 +28,40 @@ using shaderc_util::string_piece;
...
+            glslc::DeduceDefaultShaderKindFromFileName("a.comp"));
 }
 
-TEST(StageDeducer, InvalidStages) {
+TEST(DeduceDefaultShaderKindFromFileName, InvalidStage) {
   std::stringstream error_stream;
+  EXPECT_EQ(shaderc_glsl_infer_from_source,
+            glslc::DeduceDefaultShaderKindFromFileName("a.glsl"));
 
-  EXPECT_EQ(EShLangCount, glslc::StageDeducer("a.glsl")(&error_stream, "tag"));
```
Qining: Tests about StageDeducer are moved to libshaderc's tests. And by porting to libshaderc, we are not using stage deducer any more, so its tests are removed.

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,27 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
+      compiler_.CompileGlslToSpv(source_string.begin(), source_string.size(),
+                                 shader_stage, error_file_name.data(), options_);
+  total_errors_ += result.GetNumErrors();
+  total_warnings_ += result.GetNumWarnings();
+
+  bool compilation_success =
+      result.GetCompilationStatus() == shaderc_compilation_status_success
+          ? true
+          : false;
+
```
dekimir: Delete the above two lines; they're unnecessary.

glslc/src/file_compiler.h
```diff
@@ -57,24 +60,46 @@ class FileCompiler : public shaderc_util::Compiler {
...
   // represents the number of files that will be compiled.
   bool ValidateOptions(size_t num_files);
 
   // Outputs to std::cerr the number of warnings and errors if there are any.
   void OutputMessages();
 
-  // When any files are to be compiled, they are compiled individually and
-  // written to separate output files instead of linked together.
-  void SetIndividualCompilationMode();
+  // Sets the flag to indicate individual compilation mode. In this mode, all
```
dekimir: Use "flag" not "flags" here and elsewhere (including the function name).

glslc/src/file_compiler.h
```diff
@@ -57,24 +60,46 @@ class FileCompiler : public shaderc_util::Compiler {
...
+  // files are compiled individually and written to separate output files
+  // instead of linked together. This method also sets the needs_linking_ flags
```
dekimir: Do not use class privates in describing its public interface.

glslc/src/main.cc
```diff
@@ -97,11 +97,16 @@ bool GetOptionArgument(int argc, char** argv, int* index,
...
+  std::vector<std::pair<std::string, shaderc_shader_kind>> input_files;
+  shaderc_shader_kind forced_shader_stage = shaderc_glsl_infer_from_source;
   glslc::FileCompiler compiler;
   bool success = true;
   bool has_stdin_input = false;
+  // Contains all the macro names data. Current libshaderc interface requires
+  // null-terminated string, we need to 'pull out' and copy the macro names and
+  // hold their data here.
+  std::vector<std::string> macro_names;
+  std::vector<std::string> macro_values;
```
Qining: These two lines are the source of the failure on WIndows build. The addresses of the elements in vector is not fixed, as changing the capacity will copy/move what it contains. Will change to 'list' in the next CL. (after changing to 'list', Windows build has all tests passed. This time VS somehow saved me...)

dekimir: Sadly, this means that `shaderc_compile_options_add_macro_definition` is not fulfilling the contract described in its blurb.  It should be copying the macro strings.

glslc/src/file_compiler.h
```diff
@@ -16,20 +16,23 @@
 #define GLSLC_FILE_COMPILER_H
 
 #include <string>
-#include "libshaderc_util/compiler.h"
+
+#include "libshaderc_util/string_piece.h"
+#include "libshaderc_util/file_finder.h"
+#include "shaderc.hpp"
```
dekimir: How does this work without adding the `libshaderc/include` directory into CMakeLists.txt?

Qining: The 'target_link_libraries(glslc PRIVATE ... shaderc)' does the work. This brings shaderc's public include directories to glslc.

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,25 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
+  total_warnings_ += result.GetNumWarnings();
+
+  bool compilation_success =
+      result.GetCompilationStatus() == shaderc_compilation_status_success;
+
+  // Write output to output file.
+  output_stream->write(result.GetData(), result.GetLength());
+  // Write error message to std::cerr.
+  std::cerr << result.GetErrorMessage();
+  // Something wrong happen for output.
```
dekimir: "happened on" instead of "happen for"

Also move the comment to the next line.

glslc/src/file_compiler.cc
```diff
@@ -77,6 +89,28 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
                 << output_file_name_ << "'" << std::endl;
     }
   }
+
+  // Handle the error message for failing to deduce the shader kind.
+  if (result.GetCompilationStatus() ==
+      shaderc_compilation_status_invalid_stage) {
```
dekimir: In this situation, should we depend on `output_stream->write` succeeding on line 79?  If not, this check should happen before (and instead of) outputting.

glslc/src/file_compiler.h
```diff
@@ -57,24 +60,45 @@ class FileCompiler : public shaderc_util::Compiler {
...
+  // files are compiled individually and written to separate output files
+  // instead of linked together. This method also disables linking and sets the
+  // output file extension to ".spv". Disassembly mode and preprocessing only
+  // mode override this mode and flags.
+  void SetIndividualCompilationFlag();
+
+  // Sets the flag to indicate disassembly mode. In this mode, the compiler
+  // emits disassembled textual output, instead of outputting object files.
+  // This method also sets the output file extension to ".s" and disables
+  // linking. This mode overrides individual compilation mode, and preprocessing
```
dekimir: I find it clearer when preprocessing-only is hyphenated; I wonder if others agree?

Qining: I think preprocessing-only mode is better, just need to update the libshaderc comments to keep consistency. I will do that.

glslc/src/file_compiler.cc
Qining: This doesn't depends on any file writing code, I think you are right, I should move it ahead. 

Another thing is that, if we find error during writing output, should we return 'false' for to indicate the failure? I think we should return false.

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,46 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
+          << "glslc: error: '-': -fshader-stage required when input is from "
+             "standard "
+             "input \"-\"";
+    } else {
+      std::cerr << "glslc: error: "
+                << "'" << error_file_name << "': "
+                << "file not recognized: File format not recognized";
+    }
+    std::cerr << "\n";
+  }
```
dekimir: I'm still not sure we should fall through here to outputting the result.  Is the result even guaranteed to be valid when `shaderc_compilation_status_invalid_stage`?

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,46 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
 
-  if (!compilation_success && output_stream->fail()) {
+  // Write output to output file.
+  output_stream->write(result.GetData(), result.GetLength());
+  // Write error message to std::cerr.
+  std::cerr << result.GetErrorMessage();
+  // Something wrong happen on output.
```
dekimir: Replace "happen" with "happened" and move to the next line.

glslc/src/file_compiler.cc
Qining: I guess the output stream is empty in that case, but you are right, it should return false before emitting output.

glslc/src/file_includer.cc
dekimir: Captured in #105.

glslc/src/file_includer.h
```diff
@@ -17,28 +17,28 @@
...
-#include "libshaderc_util/counting_includer.h"
 #include "libshaderc_util/file_finder.h"
+#include "shaderc.hpp"
 
 namespace glslc {
 
 // A CountingIncluder for files.  Translates the #include argument into the
 // file's contents, based on a FileFinder.
-class FileIncluder : public shaderc_util::CountingIncluder {
+class FileIncluder : public shaderc::CompileOptions::IncluderInterface{
```
dekimir: Run `clang-format`.

glslc/src/file_includer.h
```diff
@@ -17,28 +17,28 @@
...
       : file_finder_(*file_finder) {}
+  shaderc_includer_response* GetInclude(const char* filename) override;
+  void ReleaseInclude(shaderc_includer_response* data) override;
 
  private:
-  // Finds filename in search path and returns its contents.
-  std::pair<std::string, std::string> include_delegate(
-      const char* filename) const override;
-
   // Used by include() to get the full filepath.
```
dekimir: Replace `include` with `GetInclude`.  Or eliminate stale-bait. ;)

glslc/src/file_includer.h
```diff
@@ -17,28 +17,28 @@
...
 
-#include "glslang/Public/ShaderLang.h"
-
-#include "libshaderc_util/counting_includer.h"
 #include "libshaderc_util/file_finder.h"
+#include "shaderc.hpp"
 
 namespace glslc {
 
 // A CountingIncluder for files.  Translates the #include argument into the
```
dekimir: Blurb out of date.

glslc/src/main.cc
```diff
@@ -97,11 +98,18 @@ bool GetOptionArgument(int argc, char** argv, int* index,
 }  // anonymous namespace
 
 int main(int argc, char** argv) {
-  std::vector<std::pair<std::string, EShLanguage>> input_files;
-  EShLanguage force_shader_stage = EShLangCount;
+  std::vector<std::pair<std::string, shaderc_shader_kind>> input_files;
+  shaderc_shader_kind forced_shader_stage = shaderc_glsl_infer_from_source;
```
dekimir: I wonder if there's a better name for `forced_shader_stage`.  It's just the value of the latest `-fshader-stage` flag, right?  So, perhaps something like `current_fshader_stage` might be better?

glslc/src/main.cc
```diff
@@ -133,8 +141,12 @@ int main(int argc, char** argv) {
       compiler.SetWorkingDirectory(workdir.str());
     } else if (arg.starts_with("-fshader-stage=")) {
       const string_piece stage = arg.substr(std::strlen("-fshader-stage="));
-      force_shader_stage = glslc::GetShaderStageFromCmdLine(arg);
-      if (force_shader_stage == EShLangCount) {
+      // forced_shader_stage can be 1) one of the forced shader kinds, which
+      // means following files will be compiled with the specified shader
+      // kind; 2) shaderc_glsl_infer_from_source, which means 'arg' doesn't
+      // specify a valid shader kind, an error should be emitted for this.
```
dekimir: The above really should be a part of `GetForcedShaderKindFromCmdLine`'s blurb, instead of existing at every call site.

glslc/src/main.cc
```diff
@@ -165,35 +177,54 @@ int main(int argc, char** argv) {
...
       } else {
         const string_piece argument = arg.substr(2);
-        size_t name_length = argument.find_first_of('=');
-        const string_piece name = argument.substr(0, name_length);
-        if (name.starts_with("GL_")) {
+        // Get the exact length of the macro string.
+        size_t equal_sign_loc = argument.find_first_of('=');
+        size_t name_length = equal_sign_loc != shaderc_util::string_piece::npos
+                                 ? equal_sign_loc
+                                 : (arg.size() - strlen("-D"));
```
dekimir: Nit: use `argument.size()` instead.

glslc/src/main.cc
```diff
@@ -165,35 +177,54 @@ int main(int argc, char** argv) {
...
+        const string_piece name_piece = argument.substr(0, name_length);
+        size_t argument_length = argument.size() - (argument.back() == '=' ? 1 : 0);
```
dekimir: This is confusing and complex.  I suggest removing it and checking for value absence in a different way -- see below.

glslc/src/main.cc
```diff
@@ -165,35 +177,54 @@ int main(int argc, char** argv) {
...
+        // We have to insert a null terminator at the end of name, as we are
+        // using libshaderc's API instead of libshaderc_util's. Libshaderc API
+        // assumes a const char* string, while libshaderc_util assume a
+        // string_piece one. The original implementation doesn't append '\0' but
+        // works fine with libshaderc_util because it has 'end_' property. But
+        // libshaderc api doesn't know that so the macro 'name' will include the
+        // 'value' if we don't insert '\0'.
+
+        macro_names.emplace_back(name_piece.data(), name_length);
+        macro_values.emplace_back(name_length < argument_length
```
dekimir: Use this condition instead: `equal_sign_loc == npos || equal_sign_loc == argument.size() - 1`.

glslc/src/file_includer.cc
```diff
@@ -18,17 +18,39 @@
...
+      shaderc_util::ReadFile(file_full_path_, &file_content_)) {
+    response_ = {
+        file_full_path_.c_str(), file_full_path_.length(), file_content_.data(),
+        file_content_.size(),
+    };
   } else {
-    // TODO(deki): Emit error message.
-    return std::make_pair("", "");
+    // The file to be included is not found or can not be opened. Response to be
+    // returned should have an empty string as path, and an error message as content.
```
dekimir: Redundant: this is a part of the includer interface and shouldn't be repeated everywhere an includer is used.

glslc/src/file_includer.cc
```diff
@@ -18,17 +18,39 @@
...
+    char error_msg[] = "Cannot open or find include file.";
+    file_content_.insert(file_content_.begin(), error_msg,
```
dekimir: Use `vector::assign` instead of `insert`.

Use `std::begin(error_msg)` and `std::end(error_msg)` for better readability.

glslc/src/file_includer.cc
```diff
@@ -18,17 +18,39 @@
...
+    response_ = {
+        filename, std::strlen(filename), file_content_.data(),
+        file_content_.size(),
+    };
   }
+  return &response_;
+}
+
+void FileIncluder::ReleaseInclude(shaderc_includer_response* data) {
+  // In this first draft, we don't need to use data argument, as we cache only
```
dekimir: Stale-bait.  I'd leave out the whole comment.

glslc/src/file_includer.cc
```diff
@@ -18,17 +18,39 @@
...
+  // the latest response data. All we need to do is resetting the cached
+  // response data to null.
+  (void)data;
```
dekimir: Instead of this, don't name `data` at all in the signature.

https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions

glslc/src/file_includer.h
```diff
@@ -17,28 +17,31 @@
...
 
  private:
-  // Finds filename in search path and returns its contents.
-  std::pair<std::string, std::string> include_delegate(
-      const char* filename) const override;
-
-  // Used by include() to get the full filepath.
+  // Used by GetInclude() to get the full filepath.
   const shaderc_util::FileFinder& file_finder_;
+  shaderc_includer_response response_;
```
dekimir: Add a blurb explaining that only one response needs to be kept alive at any time due to the `InternalFileIncluder::include_delegate` implementation.

glslc/src/main.cc
```diff
@@ -97,11 +98,18 @@ bool GetOptionArgument(int argc, char** argv, int* index,
 }  // anonymous namespace
 
 int main(int argc, char** argv) {
-  std::vector<std::pair<std::string, EShLanguage>> input_files;
-  EShLanguage force_shader_stage = EShLangCount;
+  std::vector<std::pair<std::string, shaderc_shader_kind>> input_files;
+  shaderc_shader_kind current_shader_stage = shaderc_glsl_infer_from_source;
```
dekimir: Nit: consider `current_fshader_stage` (note the "f"), because this keeps that flag's value.

glslc/src/shader_stage.h
```diff
@@ -17,29 +17,21 @@
...
-                         const shaderc_util::string_piece& error_tag) const;
-
- private:
-  shaderc_util::string_piece file_name_;
-};
-
-EShLanguage GetShaderStageFromCmdLine(
-    const shaderc_util::string_piece& f_shader_stage);
+// Parse the string piece from command line to get the force shader stage.
+// If the 'f_shader_stage_str' can not be parsed to an valid force shader stage,
```
dekimir: "cannot"

"a valid"

glslc/src/file_includer.h
```diff
@@ -17,28 +17,31 @@
...
-#include "libshaderc_util/counting_includer.h"
 #include "libshaderc_util/file_finder.h"
+#include "shaderc.hpp"
 
 namespace glslc {
 
-// A CountingIncluder for files.  Translates the #include argument into the
-// file's contents, based on a FileFinder.
-class FileIncluder : public shaderc_util::CountingIncluder {
+// An includer for files implementing shaderc's includer interface. It resposnes
```
dekimir: "responds"

Qining: Just updated and fixed the comments before this.

glslc/src/file_includer.cc
```diff
@@ -18,17 +18,31 @@
...
+        file_full_path_.c_str(), file_full_path_.length(), file_content_.data(),
+        file_content_.size(),
+    };
   } else {
-    // TODO(deki): Emit error message.
-    return std::make_pair("", "");
+    char error_msg[] = "Cannot open or find include file.";
+    file_content_.assign(std::begin(error_msg), std::end(error_msg));
+    response_ = {
+        filename, std::strlen(filename), file_content_.data(),
```
dekimir: User an empty string here, not `filename`.

dekimir: Also add a test for this; somehow we're lacking that coverage. :(

Qining: Updated with a test for 'file not found' case. But actually we still have another case: 'cannot open file' not covered.

glslc/src/file_compiler.cc
```diff
@@ -62,13 +62,48 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
     source_string = string_piece(&(*input_data.begin()),
                                  &(*input_data.begin()) + input_data.size());
   }
-  StageDeducer deducer(input_file);
+
+  std::unique_ptr<FileIncluder> includer(new FileIncluder(&include_file_finder_));
+  options_.SetIncluder(std::move(includer));
+
+  shaderc::SpvModule result =
+      compiler_.CompileGlslToSpv(source_string.begin(), source_string.size(),
```
Qining: I will change source_string.begin() to source_string.data().

Qining: I will change source_string.begin() to source_string.data().

dekimir:
+2 finally :smile:

Qining:
squash and merge to master 2b4eafba4dbc2d7229dc317aa3cae8819f25d00f

PR #102, opened 2016-01-26
Create a class wrapper for SPIRV-Tools API.

antiagainst:
Use singleton to defer creating SPIRV-Tools context object and
amortize the overhead.

### DIFF
--- a/libshaderc_util/src/compiler.cc
+++ b/libshaderc_util/src/compiler.cc
@@ -67,40 +67,61 @@ std::pair<int, string_piece> DecodeLineDirective(string_piece directive) {
   return std::make_pair(line, directive);
 }
 
-// Writes the message contained in the diagnostic parameter to *dest. Assumes
-// the diagnostic message is reported for a binary location.
-void OutputSpvToolsDiagnostic(spv_diagnostic diagnostic, std::string* dest) {
-  assert(!diagnostic->isTextSource);
+// A wrapper class for SPIRV-Tools API. Uses singleton design pattern here to
+// amortize the cost of creating the SPIRV-Tools context object.
+class SpvtoolsApi {
+ public:
+  ~SpvtoolsApi() { spvContextDestroy(context_); }
 
-  std::ostringstream os;
-  os << diagnostic->position.index << ": " << diagnostic->error;
-  *dest = os.str();
-}
-
-// Disassembles the given binary. Returns SPV_SUCCESS and writes the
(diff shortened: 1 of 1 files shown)


antiagainst:
I remembered that @dneto0 had concerns about this approach when commenting on #81, but didn't say why. So still push it out to seek opinions.

AWoloszyn:
So I think @dneto0 was trying to raise 2 points.
 1) We plan on removing the necessity of the spirv-tools context object by using precompiled tables, rather than run-time generated ones. So this makes the cost 0 for initialization and freeing.
 2) As a rule singletons are probably the wrong solution to any given problem.

My opinions on why we should avoid singletons at all costs are:
  - They are simply global objects that are initialized correctly, but that does not actually prevent a problem with freeing them. (Some other singleton that referenced this could cause all sorts of destruction order problems).
 - This singleton in particular is being returned as a reference, meaning that it is modifiable by any user
which does not play well in multi-threaded contexts. (SpvtoolsApi ::GetInstance()::~SpvtoolsApi() for example). We also do not have guarantees that Disassemble can be called in a multi-threaded context.
 - This removes a lot of control from the user of shaderc in terms of performance/memory. If we enabled copying of spirv-tools contexts, and put it in the compiler object, then the user of the library could control exactly when these things are created/destroyed, and pay the costs when they know they are willing to do so, rather than at some unspecified time in the future.

dekimir:
To agree with @AWoloszyn : http://www.object-oriented-security.org/lets-argue/singletons

libshaderc_util/src/compiler.cc
```diff
@@ -67,40 +67,61 @@ std::pair<int, string_piece> DecodeLineDirective(string_piece directive) {
...
+// amortize the cost of creating the SPIRV-Tools context object.
+class SpvtoolsApi {
+ public:
+  ~SpvtoolsApi() { spvContextDestroy(context_); }
+
+  // Returns the single API instance. The compiler guarantees that this instance
+  // is lazy initialized and will be destructed at program termination. The
+  // initialization is thread-safe under C++11 semantics.
+  static SpvtoolsApi& GetInstance() {
+    static SpvtoolsApi instance;
```
dekimir: I'm afraid this violates the style guide for the same reasons discussed here:

https://github.com/KhronosGroup/SPIRV-Tools/issues/53

antiagainst:
Thanks @AWoloszyn and @dekimir! Makes sense for me.

Lured some excellent technical insights by posting wrong solutions ;)

PR #103, opened 2016-01-26
Qualify CMake function names with "shaderc" to avoid conflictions.

antiagainst:
If a third-party project defines a CMake function with the same
name, by importing that project's CMake configuration, we may end
up overwriting our own copy. Qualify all defined functions to
reduce that probability.

Counterpart for https://github.com/KhronosGroup/SPIRV-Tools/pull/85


antiagainst:
It happens we have `default_compile_options` defined both in shaderc and SPIRV-Tools. So, things got wired and [lots of compilation warnings are given](https://travis-ci.org/google/shaderc/jobs/104183180).

dekimir:
+2

Issue #105, opened 2016-01-27
shaderc_includer_response doesn't allow for includer errors


dekimir:
I may be missing something, but it's not clear to me how to construct a shaderc_includer_response when the path is incorrect or content-reading fails.

dekimir:
Looks like `TShader::Includer::include()` has a dubious convention that an empty content string indicates an error.  That, of course, precludes a perfectly valid scenario in which the `#include`d file exists and happens to be empty.

Sadly, it was me who designed `TShader::Includer::include()` in the first place. :sob:

Qining:
In file_includer.cc of PR #100,

dekimir:
It appears that the convention is to return "" for the full path (and error message for the content) when an error occurs.  This should be mentioned in the blurb.

PR #106, opened 2016-01-27
Libshaderc rename kind to stage

Qining:
Rename 'kind' to 'stage'. Now we need to distinguish libhshaderc stage with glslang stage. 

Change 'preprocessing only mode' in the blurbs of libshaderc to 'preprocessing-only mode'

Add blurbs for the includer interface defined in shaderc.hpp. Hope this helps with #105.

### DIFF
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -31,7 +31,7 @@ using shaderc_util::string_piece;
 
 namespace glslc {
 bool FileCompiler::CompileShaderFile(const std::string& input_file,
-                                     shaderc_shader_kind shader_stage) {
+                                     shaderc_shader_stage shader_stage) {
   std::vector<char> input_data;
   std::string path = input_file;
   if (!workdir_.empty() && !shaderc_util::IsAbsolutePath(path)) {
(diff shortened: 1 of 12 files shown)


dneto0:
-1  Too many logically distinct things in the same pull request. Should be at least two different pull requests: one for comment tweaks, the other for the enum change.

But why is "stage" superior to "kind"?   Stages make more sense for a graphics pipeline (and lines up with the GL docs), but it sounds awkward with a compute shader since compute shaders are always alone.  Compute shaders are a late addition to OpenGL, which is probably why they *kept* "stage" in the docs.

Qining:
I got it. Stage does sound awkward when it comes to compute shader. I won't work on this PR for now and leave it here, or I can close it if you like.

dekimir:
> But why is "stage" superior to "kind"? 

Because "kind" isn't used anywhere: not in the `glslc` manual, not in glslang, not in Vulkan specs.  They all speak only of "stage".

dneto0:
Thanks for closing.
FWIW, the SPIR-V spec calls this concept "execution model"
https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#Execution_Model

So there's already official movement away from the "stage" concept.

PR #107, opened 2016-01-27
Improve includer blurbs and run clang-format.

dekimir:


dekimir:
@Qining @AWoloszyn @dneto0 PTAL

libshaderc/src/shaderc.cc
```diff
@@ -168,6 +168,7 @@ class StageDeducer {
   bool error_;
 };
 
+// A bridge between the libshaderc includer and libshaderc_util includer.
 class InternalFileIncluder : public shaderc_util::CountingIncluder {
```
Qining: I think we should change the name 'InternalFileIncluder' to something else. Anyway, we can do that later in another CL.

Qining:
lgtm

PR #108, opened 2016-01-28
add a test for unreadable include file

Qining:


glslc/test/environment.py
```diff
@@ -62,11 +63,16 @@ def write(self, parent):
...
 
     def write(self, directory):
         """Writes content to directory/name."""
-        with open(os.path.join(directory, self.name), 'w') as f:
+        full_path = os.path.join(directory, self.name)
+        with open(full_path, 'w') as f:
             f.write(self.content)
+        if not self.readable:
+          access_bits = stat.S_IMODE(os.stat(full_path).st_mode)
+          os.chmod(full_path, access_bits & ~stat.S_IREAD)
```
dekimir: How portable is this?  [Man page](https://docs.python.org/2/library/stat.html) says that this is "Unix V7 synonym for S_IRUSR", whatever that means.

glslc/test/include.py
```diff
@@ -51,6 +51,20 @@ class VerifyIncludeNotFound(expect.ErrorMessage):
...
+    """Tests #including an existing but not readable sibling file."""
+
+    environment = Directory('.', [
+        File('a.vert', 'content a\n#include "b"\n'),
+        File('b', 'content b\n', readable=False)])
+
+    glslc_args = ['-E', 'a.vert']
+    expected_error = [
+        "a.vert:2: error: '#include' : Cannot open or find include file.\n",
+        '1 error generated.\n'
```
Qining: The error message here is exactly the same as the include file 'NotFound' case. The error message in  libshaderc_util::ReadFile() doesn't show up here. This is because libshaderc_util::FileFinder's method: FindReadableFilePath() checks the access bits before calling ReadFile() so the call of ReadFile() in FileIncluder will never have an unreadable path to read.

glslc/test/environment.py
Qining: I think I should change it to S_IRUSR, gnu says:
"S_IREAD is an obsolete synonym provided for BSD compatibility."

dekimir:
+2

Qining:
It seems that Windows doesn't respect os.chmod() in Python (or may be some other problems). Windows Release build failed.

PR #109, opened 2016-01-28
change the contained class of predefined_macros_ from string_piece to…

### LINKED ISSUE #104: shaderc_compile_options_add_macro_definition doesn't copy macro name and value
The blurb for this function promises that name/value can be modified or deleted after it returns.  But that's currently not true: the underlying implementation merely keeps pointers to name/value.  The function should obey its contract.

Qining:
try to resolve #104

### DIFF
--- a/libshaderc/include/shaderc.h
+++ b/libshaderc/include/shaderc.h
@@ -136,16 +136,20 @@ shaderc_compile_options_t shaderc_compile_options_clone(
 // NULL to this function, and doing such will have no effect.
 void shaderc_compile_options_release(shaderc_compile_options_t options);
 
-// Adds a predefined macro to the compilation options. This has the
-// same effect as passing -Dname=value to the command-line compiler.
-// If value is NULL, it has the effect same as passing -Dname to the
-// command-line compiler. If a macro definition with the same name has
-// previously been added, the value is replaced with the new value.
-// The null-terminated strings that the name and value parameters point to
-// must remain valid for the duration of the call, but can be modified or
-// deleted after this function has returned.
+// Adds a predefined macro to the compilation options. This has the same
+// effect as passing -Dname=value to the command-line compiler.  If value
+// is NULL, it has the same effect as passing -Dname to the command-line
+// compiler. If a macro definition with the same name has previously been
+// added, the value is replaced with the new value. The macro name and
+// value are passed in with char pointers, which point to their data, and
+// the lengths of their data. The strings that the name and value pointers
+// point to must remain valid for the duration of the call, but can be
+// modified or deleted after this function has returned. In case of adding
+// a valueless macro, the value argument should be a null pointer or the
+// value_length should be 0u.
 void shaderc_compile_options_add_macro_definition(
-    shaderc_compile_options_t options, const char* name, const char* value);
+    shaderc_compile_options_t options, const char* name, size_t name_length,
+    const char* value, size_t value_length);
 
 // Sets the compiler mode to generate debug information in the output.
 void shaderc_compile_options_set_generate_debug_info(
--- a/libshaderc/include/shaderc.hpp
+++ b/libshaderc/include/shaderc.hpp
@@ -109,21 +109,22 @@ class CompileOptions {
     other.options_ = nullptr;
   }
 
-  // Adds a predefined macro to the compilation options.
-  // For further details refer to shaderc_compile_options_add_macro_definition
-  // in shaderc.h
-  void AddMacroDefinition(const char* name, const char* value) {
-    shaderc_compile_options_add_macro_definition(options_, name, value);
+  // Adds a predefined macro to the compilation options. It behaves the same as
+  // shaderc_compile_options_add_macro_definition in shaderc.h.
+  void AddMacroDefinition(const char* name, size_t name_length,
+                          const char* value, size_t value_length) {
+    shaderc_compile_options_add_macro_definition(options_, name, name_length,
+                                                 value, value_length);
   }
 
   // Adds a valueless predefined macro to the compilation options.
   void AddMacroDefinition(const std::string& name) {
-    AddMacroDefinition(name.c_str(), nullptr);
+    AddMacroDefinition(name.c_str(), name.size(), nullptr, 0u);
   }
 
   // Adds a predefined macro to the compilation options.
   void AddMacroDefinition(const std::string& name, const std::string& value) {
-    AddMacroDefinition(name.c_str(), value.c_str());
+    AddMacroDefinition(name.c_str(), name.size(), value.c_str(), value.size());
   }
 
   // Sets the compiler mode to generate debug information in the output.
--- a/libshaderc/src/shaderc_test.cc
+++ b/libshaderc/src/shaderc_test.cc
@@ -300,7 +300,8 @@ TEST_F(CompileStringWithOptionsTest, CloneCompilerOptions) {
 
 TEST_F(CompileStringWithOptionsTest, MacroCompileOptions) {
   ASSERT_NE(nullptr, compiler_.get_compiler_handle());
-  shaderc_compile_options_add_macro_definition(options_.get(), "E", "main");
+  shaderc_compile_options_add_macro_definition(options_.get(), "E", 1u, "main",
+                                               4u);
   const std::string kMinimalExpandedShader = "#version 140\nvoid E(){}";
   const std::string kMinimalDoubleExpandedShader = "#version 140\nF E(){}";
   EXPECT_TRUE(CompilesToValidSpv(compiler_, kMinimalExpandedShader,
@@ -315,8 +316,8 @@ TEST_F(CompileStringWithOptionsTest, MacroCompileOptions) {
                                   shaderc_glsl_vertex_shader,
                                   cloned_options.get()));
 
-  shaderc_compile_options_add_macro_definition(cloned_options.get(), "F",
-                                               "void");
+  shaderc_compile_options_add_macro_definition(cloned_options.get(), "F", 1u,
+                                               "void", 4u);
   // This should still not work with the original options.
   EXPECT_FALSE(CompilesToValidSpv(compiler_, kMinimalDoubleExpandedShader,
                                   shaderc_glsl_vertex_shader, options_.get()));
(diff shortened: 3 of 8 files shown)


libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -76,8 +75,8 @@ class Compiler {
 
   // Adds an implicit macro definition obeyed by subsequent CompileShader()
   // calls.
-  void AddMacroDefinition(const string_piece& macro,
```
dneto0: Should say the second argument can be nullptr, and how that's interpreted.
Add a test for that case.

dneto0:
-1 should update the blurb to describe nullability of the "definition" part of the value.

I feel we should be describing the const char* arguments as "null terminated strings".  The null terminated part matters.

Also, I'm getting uneasy with the plethora of interfaces that take a c-style string but no size argument.  Yes, it's less convenient, but forces the caller to think about sizing.

libshaderc_util/include/libshaderc_util/compiler.h
dneto0: This comment is still unaddressed.
Update blurb, and add tests for the nullptr case.

dneto0:
-1  Still unaddressed comments.

Qining:
Added 'size' arguments to libshaderc and libshaderc_util and updated the blurbs.

Should we update glslc's code in this CL?

libshaderc_util/src/compiler_test.cc
```diff
@@ -187,4 +187,21 @@ TEST_F(CompilerTest, DISABLED_RespectTargetEnvOnVulkanShader) {
   // TODO(dneto): Add test for a shader that should only compile for Vulkan.
 }
 
+TEST_F(CompilerTest, AddMacroDefinition) {
+  const std::string kMinimalExpandedShader = "void E(){}";
+  compiler_.AddMacroDefinition("E", 1u, "main", 4u);
```
dneto0: Also add a test where the length matters for both arguments, i.e. the passed-in length is *smaller* than the length of the char* argument.

E.g. 
  AddMacroDefinition("EFGH", 1u, "mainly", 4u);

dneto0:
-1
Need more test cases as described.

Qining:
Add tests for 1) macro and value are not null terminated strings, 2) macro and value are null terminated, but the lengths of them passed in are too large.

libshaderc/include/shaderc.h
```diff
@@ -136,16 +136,21 @@ shaderc_compile_options_t shaderc_compile_options_clone(
...
-// same effect as passing -Dname=value to the command-line compiler.
-// If value is NULL, it has the effect same as passing -Dname to the
-// command-line compiler. If a macro definition with the same name has
-// previously been added, the value is replaced with the new value.
-// The null-terminated strings that the name and value parameters point to
-// must remain valid for the duration of the call, but can be modified or
-// deleted after this function has returned.
+// Adds a predefined macro to the compilation options. This has the same
+// effect as passing -Dname=value to the command-line compiler.  If value
+// is NULL, it has the effect same as passing -Dname to the command-line
```
dekimir: "effect same" => "same effect"

libshaderc/include/shaderc.h
```diff
@@ -136,16 +136,21 @@ shaderc_compile_options_t shaderc_compile_options_clone(
...
+// compiler. If a macro definition with the same name has previously been
+// added, the value is replaced with the new value. The macro name and
+// value are passed in with char pointers, which point to their data, and
+// the lengths of their data. The strings that the name and value pointers
+// point to must remain valid for the duration of the call, but can be
+// modified or deleted after this function has returned. In case of adding
+// a valueless macro, the value argument can either be a null pointer or
+// point to an empty string. For valueless macro, the value_length parameter
+// will be ignored.
```
dekimir: I think this is trying to say that you can denote a valueless macro by making `value` null or `value_length` 0 (or both).  So, clearly, `value_length` isn't ignored.  And "point to an empty string" seems ambiguous.

libshaderc/include/shaderc.hpp
```diff
@@ -109,21 +109,29 @@ class CompileOptions {
     other.options_ = nullptr;
   }
 
-  // Adds a predefined macro to the compilation options.
+  // Adds a predefined macro to the compilation options. The macro name and
+  // value are passed in with their char* pointers of their strings and their
```
dekimir: Stale bait.  Just omit this and say it behaves the same as `shaderc_compile_options_add_macro_definition`.

libshaderc/include/shaderc.hpp
```diff
@@ -109,21 +109,29 @@ class CompileOptions {
...
+  // lengths. And the strings must remain valid for the duration of the call,
+  // but can be modified or deleted after this function has returned. The macro
+  // value argument can either be a nullptr or point to an empty string in case
+  // of adding a valueless macro. In case of valueless macro, the length of the
+  // value string will be ignored.
   // For further details refer to shaderc_compile_options_add_macro_definition
```
dekimir: In fact, all the details are (redundantly) spelled out above -- there are no "further details" in `shaderc.h` right now.

libshaderc/src/shaderc_test.cc
```diff
@@ -318,6 +319,25 @@ TEST_F(CompileStringWithOptionsTest, MacroCompileOptions) {
                                  cloned_options.get()));
 }
 
+TEST_F(CompileStringWithOptionsTest, MacroCompileOptionsNotNullTerminated) {
+  ASSERT_NE(nullptr, compiler_.get_compiler_handle());
+  shaderc_compile_options_add_macro_definition(options_.get(), "EFGH", 1u, "mainnnnnn",
+                                               4u);
+  const std::string kMinimalExpandedShader = "void E(){}";
+  const std::string kMinimalDoubleExpandedShader = "F E(){}";
```
dekimir: DRY

libshaderc_util/src/compiler_test.cc
```diff
@@ -187,4 +187,27 @@ TEST_F(CompilerTest, DISABLED_RespectTargetEnvOnVulkanShader) {
...
+}
+
+TEST_F(CompilerTest, AddValuelessMacroDefinition) {
+  const std::string kMinimalExpandedShader =
+      "#ifdef E\n"
+      "void main(){}\n"
+      "#else\n"
+      "#error\n"
+      "#endif";
+  compiler_.AddMacroDefinition("E", 1u, nullptr, 0u);
```
dekimir: I wouldn't set them both to zero.  We should be able to handle either of them being zero.

dneto0:
+1 
Would like @dekimir to weigh in.

libshaderc/include/shaderc.h
```diff
@@ -136,16 +136,20 @@ shaderc_compile_options_t shaderc_compile_options_clone(
...
+// Adds a predefined macro to the compilation options. This has the same
+// effect as passing -Dname=value to the command-line compiler.  If value
+// is NULL, it has the same effect as passing -Dname to the command-line
+// compiler. If a macro definition with the same name has previously been
+// added, the value is replaced with the new value. The macro name and
+// value are passed in with char pointers, which point to their data, and
+// the lengths of their data. The strings that the name and value pointers
+// point to must remain valid for the duration of the call, but can be
+// modified or deleted after this function has returned. In case of adding
+// a valueless macro, the value argument can be a null pointer or an empty
```
dekimir: I still think "an empty string" is ambiguous and should be dropped.

libshaderc/include/shaderc.hpp
```diff
@@ -109,21 +109,24 @@ class CompileOptions {
     other.options_ = nullptr;
   }
 
-  // Adds a predefined macro to the compilation options.
+  // Adds a predefined macro to the compilation options. It behaves the same as
+  // shaderc_compile_options_add_macro_definition in shaderc.h.
   // For further details refer to shaderc_compile_options_add_macro_definition
   // in shaderc.h
```
dekimir: Now the above two lines can be dropped.

libshaderc_util/src/compiler_test.cc
```diff
@@ -187,4 +196,28 @@ TEST_F(CompilerTest, DISABLED_RespectTargetEnvOnVulkanShader) {
   // TODO(dneto): Add test for a shader that should only compile for Vulkan.
 }
 
+TEST_F(CompilerTest, AddMacroDefinition) {
+  const std::string kMinimalExpandedShader = "void E(){}";
+  compiler_.AddMacroDefinition("E", 1u, "main", 4u);
+  EXPECT_TRUE(SimpleCompilationSucceeds(kMinimalExpandedShader, EShLangVertex));
+}
+
+TEST_F(CompilerTest, AddValuelessMacroDefinitionNullPointer) {
```
dekimir: I think this test and the next would be more valuable at the libshaderc level.  More coverage of the user-visible interface, less implementation-dependent.

Qining:
Tests about 'not null-terminated macro string' and 'zero length macro value string' added to libshaderc.

Blurbs updated.

libshaderc/include/shaderc.h
```diff
@@ -136,16 +136,20 @@ shaderc_compile_options_t shaderc_compile_options_clone(
...
+// Adds a predefined macro to the compilation options. This has the same
+// effect as passing -Dname=value to the command-line compiler.  If value
+// is NULL, it has the same effect as passing -Dname to the command-line
+// compiler. If a macro definition with the same name has previously been
+// added, the value is replaced with the new value. The macro name and
+// value are passed in with char pointers, which point to their data, and
+// the lengths of their data. The strings that the name and value pointers
+// point to must remain valid for the duration of the call, but can be
+// modified or deleted after this function has returned. In case of adding
+// a valueless macro, the value argument should be a null pointer and the
```
dekimir: "and" => "or"

Qining: Should has been fixed by now.

dekimir:
+2

Qining:
Rebase and merged to master: 3bf8a0c243479f25afc0c3fcbde221a51edf1abd

PR #110, opened 2016-01-29
Target env cmd

Qining:
Add --target-env= to the command line option list of glslc


glslc/src/main.cc
```diff
@@ -61,6 +61,10 @@ void PrintHelp(std::ostream* out) {
        << "310es,\n"
        << "                    450core, etc.\n"
        << "  -S                Only run preprocess and compilation steps.\n"
+       << "  --target-env=<environment>\n"
+       << "                    Set the target shader environment, and the"
+       << " semantics\n"
+       << "                    of warnings and errors.\n"
```
dneto0: Should say what the values are, and that the default is 'vulkan'.

dneto0:
- 1
Minor change requested:  Spell out more in the online help for glslc.

A higher level point, not to be addressed in this CL:  What is the difference between -std=* and --target-env=*  ?
The key is that -std= describes the dialect of the source code, whereas the --target-env describes what system we're compiling to.  It's a subtle difference.  I might make a followup CL to address this in the manual.

dneto0:
+2
I'll squash and merge

dneto0:
Squashed and pushed into master as f6d8374504f7d8

PR #111, opened 2016-01-29
Enabled caching of glsl state.

AWoloszyn:
We only Initialize/Finalize process every compile because there
is a subset of options that interfere with internal glslang
state. Until this is remedied in glslang, allow re-use of
state except in the cases where we know state will get broken.

### DIFF
--- a/libshaderc_util/include/libshaderc_util/compiler.h
+++ b/libshaderc_util/include/libshaderc_util/compiler.h
@@ -32,10 +32,10 @@
 namespace shaderc_util {
 
 // Initializes glslang on creation, and destroys it on completion.
-// There is expected to be precisely one of these objects active within
-// the current process at a time.
+// This object is expected to be a singleton, so that internal
+// glslang state can be correctly handled.
 // TODO(awoloszyn): Once glslang no longer has static global mutable state
-//                  remove this object.
+//                  remove this class.
 class GlslInitializer {
  public:
   GlslInitializer() : last_messages_(EShMsgDefault) {
@@ -71,8 +71,8 @@ class GlslInitializer {
 
   // Obtains exclusive access to the glslang state. The state remains
   // exclusive until the Initialization Token has been destroyed.
-  // Re-initializes glsl state iff the previous rules and the current rules
-  // are incompatible.
+  // Re-initializes glsl state iff the previous messages and the current
+  // messages are incompatible.
   InitializationToken Acquire(EShMessages new_messages) {
     state_lock_.lock();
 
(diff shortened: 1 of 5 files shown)


libshaderc_util/src/compiler.cc
```diff
@@ -171,7 +173,6 @@ bool Compiler::Compile(
   }
 
   // Parsing requires its own Glslang symbol tables.
```
dneto0: This comment is now out of place.  
It was talking about the use of the GlslInitializer object.

libshaderc_util/src/compiler.cc
```diff
@@ -251,7 +254,7 @@ std::tuple<bool, std::string, std::string> Compiler::PreprocessShader(
   // or Vulkan is turned on.  Since preprocessing uses different
```
dneto0: Similarly, this comment is now no longer required

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -31,10 +32,68 @@
 namespace shaderc_util {
 
 // Initializes glslang on creation, and destroys it on completion.
+// There is expected to be precisely one of these objects active within
```
dekimir: Actually, there can be zero. :)

Also, "active" could be ambiguous.  Perhaps it's better to say that this is thread-hostile and must be a singleton.

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -31,10 +32,68 @@
...
+// the current process at a time.
+// TODO(awoloszyn): Once glslang no longer has static global mutable state
+//                  remove this object.
```
dekimir: "class" not "object"

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -31,10 +32,68 @@
...
+    InitializationToken(GlslInitializer* initializer)
+        : initializer_(initializer) {}
+
+    friend class GlslInitializer;
+    GlslInitializer* initializer_;
+  };
+
+  // Obtains exclusive access to the glslang state. The state remains
+  // exclusive until the Initialization Token has been destroyed.
+  // Re-initializes glsl state iff the previous rules and the current rules
```
dekimir: "messages" not "rules"

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -97,6 +159,10 @@ class Compiler {
   // from the shader text. Any #include directives are parsed with the given
   // includer.
   //
+  // The initializer parameter must be a valid GlslInitializer object.
+  // Acquire will be called on the initializer, and cleaned up
```
dekimir: "cleaned up" is ambiguous: better to say that `Release` will be called when compilation is finished.

dekimir:
Looks good, just needs some comment clean-up.

AWoloszyn:
@dekimir Updated the comments.

dneto0:
+2

AWoloszyn:
Squashed and merged in https://github.com/google/shaderc/commit/d080537cad6d354f7996b91cd95a2e434fc767e5

PR #115, opened 2016-02-01
Input file with absolute path should not disable -working-directory option

Qining:
Previous glslc ignore -working-directory when the input files are passed with absolute paths, the output files are placed with the input files. But clang actually respects working-directory in the same cases.

1. Whenever 'linking' is disabled, -working-directory should have effects.
2. The output file location should refer to -working-directory, except when linking is enabled.

### DIFF
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -177,22 +177,98 @@ void FileCompiler::OutputMessages() {
   shaderc_util::OutputMessages(&std::cerr, total_warnings_, total_errors_);
 }
 
+namespace {
+// An util function to get the file name (base) from a given absolute path.
+// It checks the begining of the path string to differentiate Windows and Linux.
+std::string GetBaseNameFromAbsolutePath(const std::string& path) {
+  if (path.empty()) return "";
+  // Unix-like OS: /path/to/file
+  if (path.front() == '/') return path.substr(path.find_last_of('/'));
+  // Windows: \\server\user\file
+  if (path.size() > 1 && path[0] == '\\' && path[1] == '\\') {
+    return path.substr(path.find_last_of('\\'));
+  }
+  // Windows: X:\path\to\file
+  if (path.size() > 2 && ::isalpha(path[0]) && path[1] == ':' &&
+      path[2] == '\\') {
+    return path.substr(path.find_last_of('\\'));
+  }
+  return "";
+}
+}  // anonymous namespace
+
 std::string FileCompiler::GetOutputFileName(std::string input_filename) {
-  std::string output_file = "a.spv";
-  if (!needs_linking_) {
-    if (IsStageFile(input_filename)) {
-      output_file = input_filename + file_extension_;
+  std::string default_output_file = "a.spv";
+
+  if(needs_linking_) {
+    // When linking enabled, ignore -working-directory for output, And don't
+    // induce output file name from input file name.
+    if (output_file_name_.empty()) {
+      // no '-o' argument specified, return the default output file name.
+      // e.g. glslc shader.vert
+      //      => a.spv in <curdir>/
+      return default_output_file;
     } else {
-      output_file = input_filename.substr(0, input_filename.find_last_of('.')) +
-                    file_extension_;
+      // If '-o' specified, use the argument as output file name, no matter
+      // relative path or absolute path it is.
+      // e.g. glslc shader.vert -o output.spv
+      //      => output.spv in <curdir>/
+      return output_file_name_.str();
+    }
+  } else {
+    // When linking is disable, should refer to -working-directory when the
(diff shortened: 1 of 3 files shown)


glslc/test/working_directory.py
```diff
@@ -148,15 +149,50 @@ class TestWorkDirEqInArg(expect.ValidNamedObjectFile):
...
+class TestWorkDirCompileFileAbsolutePath(WorkDirAffectsNotLinkedFile):
     """Tests -working-directory=<dir> when compiling input file with absolute
-    path."""
+    path. The output location should refer to the working directory"""
 
     shader = FileShader('void main() {}', '.vert')
+    environment = Directory('.', [
+        Directory('subdir', []),
+    ])
     glslc_args = ['-c', '-working-directory=subdir', shader]
```
Qining: The previous test expect to find the output file in the same directory with the input shader file. But we should expect the output file in 'subdir'.

glslc/src/file_compiler.cc
```diff
@@ -177,6 +177,24 @@ void FileCompiler::OutputMessages() {
   shaderc_util::OutputMessages(&std::cerr, total_warnings_, total_errors_);
 }
 
+namespace {
+std::string GetBaseNameFromAbsolutePath(const std::string& path) {
```
Qining: Should we add such a util function to libshaderc_util?

glslc/src/file_compiler.cc
```diff
@@ -177,6 +177,24 @@ void FileCompiler::OutputMessages() {
...
+    return path.substr(path.find_last_of('\\'));
+  }
+  // Windows: X:\path\to\file
+  if (path.size() > 2 && ::isalpha(path[0]) && path[1] == ':' &&
+      path[2] == '\\') {
+    return path.substr(path.find_last_of('\\'));
+  }
+  return "";
+}
+} // anonymous namespace
```
dneto0: Looks like you didn't run clang-format?
That normally puts two spaces before the namespace comment.

glslc/src/file_compiler.cc
dneto0: If we don't need it in multiple places, then let's just keep it here.

dneto0: Blurb!

dneto0: Hmm. I see now that libshaderc_util/io.cpp has IsAbsolutePath, and that repeats some of the same logic here to determine which kind of path is being used.

So it would make sense to put this method into libshaderc_util.

glslc/src/file_compiler.cc
```diff
@@ -188,8 +206,11 @@ std::string FileCompiler::GetOutputFileName(std::string input_filename) {
     }
   }
   if (!output_file_name_.empty()) output_file = output_file_name_.str();
-  if (!needs_linking_ && !workdir_.empty() &&
-      !shaderc_util::IsAbsolutePath(input_filename)) {
+  if (!needs_linking_ && !workdir_.empty()) {
+    if (shaderc_util::IsAbsolutePath(output_file)) {
+      // Get the file name from the absolute path of the input file.
+      output_file = GetBaseNameFromAbsolutePath(output_file);
```
dneto0: This seems really confusing.

dneto0:
Seems the blurb at GetOutputFileName is very underspecified.  The glslc/README.asciidoc also leaves some holes.

dneto0:
Here's my attempt at spelling out exactly how to compute the output file name.
I'd like @dekimir and @AWoloszyn to weigh in on it.

https://github.com/dneto0/shaderc/commit/92a7d3d8f765215df2c79d3a958aeb4ee7cdce00

glslc/src/file_compiler.cc
```diff
@@ -188,8 +208,13 @@ std::string FileCompiler::GetOutputFileName(std::string input_filename) {
     }
   }
   if (!output_file_name_.empty()) output_file = output_file_name_.str();
-  if (!needs_linking_ && !workdir_.empty() &&
-      !shaderc_util::IsAbsolutePath(input_filename)) {
+  if (!needs_linking_ && !workdir_.empty()) {
+    if (shaderc_util::IsAbsolutePath(output_file)) {
```
dneto0: This seems wrong.  Shouldn't it be ! ... IsAbsolutePath?

dneto0:
Please see proposed comment in #119 for how I'd like it to all work.
And yes, there are missing cases in our test suite. So those should be added too.

dneto0:
We've removed -working-directory, so this PR should be reworked or dropped. :-)

Qining:
Working directory has been removed from glslc in commit: 7b94f85f789f4739df8216ef5016e11ddd467efc. 

So this PR is closed.

PR #117, opened 2016-02-02
Don't run spirv-tools-testsuite.

dekimir:
When spirv-tools is added outside the Shaderc project, let that outside
location decide how and if to run the spirv-tools tests.


dneto0:
Wouldn't it just be better to do add_test in SPIRV-Tools itself?

Then delete this end of things altogether.

dekimir:
@dneto0 Good point, let me try that.

dekimir:
Updated to skip spirv-tools-testsuite at this level altogether.

See also https://github.com/KhronosGroup/SPIRV-Tools/pull/95.

AWoloszyn:
+2

PR #118, opened 2016-02-02
Add Error message substring comparison and use it for --target-env tests

Qining:
Only test that the error message has what we expected as a substring, doesn't require an exact match. This fix the --target-env test for glslc, which have different error messages with different glslang.

### DIFF
--- a/glslc/test/expect.py
+++ b/glslc/test/expect.py
@@ -274,6 +274,35 @@ class ErrorMessage(GlslCTest):
         return True, ''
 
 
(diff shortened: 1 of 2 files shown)


dneto0:
+2

dneto0:
-1.  It's failing for me with different glslang

dneto0:
Superceded by https://github.com/google/shaderc/pull/121 
which also adjusts tests for newer glslang versions.

PR #121, opened 2016-02-03
Glslang spv rejects SPIR-V compiles for old versions of GLSL code

dneto0:
Adjust tests accordingly.


Qining:
lgtm. I see all the --target-env stuffs are included.

glslc/test/include.py
```diff
@@ -125,12 +132,14 @@ class VerifyCompileIncludeTwoSiblings(expect.ValidObjectFile):
...
         File('b.vert',
-             """#include "a"
+             """#version 140
+             #include "a"
              void bfun(){afun();}
              #include "c"
              """),
         File('a',
-             """void afun(){}
+             """
```
dekimir: Why was this newline necessary?  Is there a another latent bug in the `#include`d content scanner?

glslc/test/include.py
```diff
@@ -142,21 +151,23 @@ class VerifyNestedIncludeAmongSiblings(expect.StdoutMatch):
...
         File('b', 'content b\n#include "c"\n'),
         File('c', 'content c\n')])
 
     glslc_args = ['-E', 'a.vert']
 
     expected_stdout = \
-"""#extension GL_GOOGLE_include_directive : enable
+"""#version 140
+#extension GL_GOOGLE_include_directive : enable
 #line 0 "a.vert"
```
dekimir: How is this still correct?

dekimir:
Looks very good overall.  Apparently, some of the added newlines aren't strictly necessary; I'd avoid those in `#include` tests, which were crafted to target specific parser bugs.  Also I'm not sure about the relationship between `#version` and the first `#line` in preprocessor output.

glslc/test/include.py
dneto0: Turns out it's not necessary.

glslc/test/include.py
dneto0: I haven't modified any of the glslang #line processing or #include processing.
Are you objecting to the line number still being 0?  
I suspect that the #extension line is injected as a preamble that doesn't count toward line numbers in a.vert.  Also, #line number handling semantics changed in some version, I can't remember which.

glslc/test/include.py
dneto0: Later commit has reverted to the previous text in this case.

glslc/test/include.py
dekimir: I was referring to the fact that a line was inserted into `a.vert` so all `#line`s in the output should have shifted, yet not all did.  But I withdraw my comment on the basis that there's no spec for this, and we can declare `#version` special.

dekimir:
+2

dneto0:
Thanks. I'll squash my 2 commits together, then push to master.

dneto0:
Pushed into master as 2 commits ending with fab641131

PR #123, opened 2016-02-05
Add dependency info dumping handler to glslc to dump dependency info

Qining:
Enable -M, -MD, -MF, -MT.

Still a work in progress.

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -236,6 +236,80 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
 a non-zero exit code from `glslc`. If `-w` is specified the warnings
 generated are suppressed before they are converted to errors.
 
+=== Dependency Generation Options
+
+==== `-M` or `-MM`
+
+`-M` generates *make* dependencies, outputs a rule suitable for *make*
+describing the dependencies of the main source file. Instead of outputting
+the result of preprocessing, the preprocessor outputs one *make* rule
+containing the SPIR-V object file name for that source file, a colon, and the
+names of all the included files.
+
+Unless specified explicitly (with `-MT`), the SPIR-V object file name in the
+generated *make* rules follows the rules of  <<output-file-naming,Output File
+Naming>> as in `-c` compilation stage.
+
+The dependency info file name can be specified by `-o` and `-MF` options. When
+both are specified, `-o` option is ignored.
+
+Specifying `-M` implies `-E`, and suppresses warnings with an implicit `-w`.
+
+`-MM` is an alias for `-M`.
+
+E.g., `glslc -M main.vert` will dump `main.vert.spv: main.vert <other included
+files>` to stdout. More examples are listed in
+<<dependency-generation-examples,Dependency Generation Examples>>
+
+==== `-MD`
+
+`-MD` tells the glslc compiler to both compile the source and generate *make*
+dependencies. Dependencies are written to a file whose name is determined as
+follows: If option `-MF` is specified, use its argument. Otherwise, use the
+filename formed by append *.d* to the name of the file containing compilation
+results.
+
+E.g., `glslc -c -MD main.vert` will generate `main.vert.spv` as the SPIR-V
+object file and `main.vert.spv.d` as the dependency info file. More examples
+are listed in <<dependency-generation-examples,Dependency Generation Examples>>
+
+==== `-MF`
+
+`-MF` lets you specify the dependency info file name when used with `-M` or
+`-MD`. This option is invalid when used with multiple input files.
+
+E.g., `glslc -c -MD main.vert -MF dep_info` will generate `main.vert.spv` as
+the SPIR-V object file and `dep_info` as the dependency info file.
+
+==== `-MT`
+
+`-MT` lets you specify the target of the rule emitted by dependency generation
+when used with `-M` or `-MD`.  This option is invalid when used with multiple
+input files.
+
+E.g., `glslc -M main.vert -MT target` will dump following dependency info to
+stdout: `target: main.vert <other dependent files>`.
+
+
+[[dependency-generation-examples]]
+.Dependency Generation Examples
+|===
+|Command Line Input|Compilation Output File|Dependency Output File|Dependency Info
+
+|glslc -M main.vert             | <NA> | <Stdout> | main.vert.spv: main.vert
+|glslc -M main.vert -o dep_info | <NA> | dep_info | main.vert.spv: main.vert
+|glslc -M main.vert -MF dep_info| <NA> | dep_info | main.vert.spv: main.vert
+|glslc -M main.vert -MT target  | <NA> | <Stdout> | target: main.vert
+|glslc -MD main.vert    |a.spv  |main.vert.spv.d|main.vert.spv: main.vert
+|glslc -c -MD main.vert |main.vert.spv|main.vert.spv.d|main.vert.spv: main.vert
+|glslc -S -MD main.vert |main.vert.s  |main.vert.s.d  |main.vert.s: main.vert
+|glslc -c -MD main.vert -MF dep_info  |main.vert.spv|dep_info|main.vert.spv: main.vert
+|glslc -c -MD main.vert -o obj  |obj  |obj.d  |obj: main.vert
+|glslc -c -MD main.vert -o obj -MF dep_info -MT target|obj|dep_info|target: main.vert
+|===
+
+
+
 == Divergence from and extensions to GLSL specifications
 
 === Source-filename-based `#line` and `\\__FILE__`
--- a/glslc/src/dependency_info.cc
+++ b/glslc/src/dependency_info.cc
@@ -1,0 +1,102 @@
+// Copyright 2015 The Shaderc Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "dependency_info.h"
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
+#include "file.h"
+#include "libshaderc_util/io.h"
+
+namespace glslc {
+
+bool DependencyInfoDumpingHandler::DumpDependencyInfo(
+    std::string compilation_output_file_name, std::string source_file_name,
+    std::string* compilation_output_ptr,
+    const std::unordered_set<std::string>& dependent_files) {
+  std::string dep_target_label = GetTarget(compilation_output_file_name);
+  std::string dep_file_name =
+      GetDependencyFileName(compilation_output_file_name);
+
+  // Dump everything to a string stream first, then dump its content to either
+  // a file or compilation output string, depends on current dumping mode.
+  std::stringstream dep_string_stream;
+  // dump target label and the source_file_name.
+  dep_string_stream << dep_target_label << ": " << source_file_name;
+  // dump the dependent file names.
+  for (auto& dependent_file_name : dependent_files) {
+    dep_string_stream << " " << dependent_file_name;
+  }
+  dep_string_stream << std::endl;
+
+  if (mode_ == dump_as_compilation_output) {
+    compilation_output_ptr->assign(dep_string_stream.str());
+  } else if (mode_ == dump_as_extra_file) {
+    std::ofstream potential_file_stream_for_dep_info_dump;
+    std::ostream* dep_file_stream = shaderc_util::GetOutputStream(
+        dep_file_name, &potential_file_stream_for_dep_info_dump);
+    *dep_file_stream << dep_string_stream.str();
+    if (dep_file_stream->fail()) {
+      std::cerr << "glslc: error: error writing dependent_files info to output "
+                   "file: '"
+                << dep_file_name << "'" << std::endl;
+      return false;
+    }
+  } else {
+    // mode_ should not be 'not_set', we should never be here.
+    return false;
+  }
+  return true;
+}
+
+std::string DependencyInfoDumpingHandler::GetTarget(
+    const std::string& compilation_output_file_name) {
+  if (!user_specified_dep_target_label_.empty()) {
+    return user_specified_dep_target_label_;
+  }
+
+  return compilation_output_file_name;
+}
+
+std::string DependencyInfoDumpingHandler::GetDependencyFileName(
+    const std::string& compilation_output_file_name) {
+  if (!user_specified_dep_file_name_.empty()) {
+    return user_specified_dep_file_name_;
+  }
+
+  return compilation_output_file_name + ".d";
+}
+
+bool DependencyInfoDumpingHandler::IsValid(std::string* error_msg_ptr,
+                                           size_t num_files) {
+  if (DumpingModeNotSet()) {
+    *error_msg_ptr =
+        "to generate dependencies you must specify either -M (-MM) or -MD";
+    return false;
+  }
+
+  if (!user_specified_dep_file_name_.empty() ||
+      !user_specified_dep_target_label_.empty()) {
+    if (num_files > 1) {
+      *error_msg_ptr =
+          "to specify dependency info file name or dependency info target, "
+          "only one input file is allowed.";
+      return false;
+    }
+  }
+  return true;
+}
+}
--- a/glslc/src/dependency_info.h
+++ b/glslc/src/dependency_info.h
@@ -1,0 +1,134 @@
+// Copyright 2015 The Shaderc Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef GLSLC_DEPENDENCY_INFO_H
+#define GLSLC_DEPENDENCY_INFO_H
+
+#include <unordered_set>
+#include <string>
+#include <string>
+
+namespace glslc {
+
+// An object to handle everything about dumping dependency info. Internally it
+// has two valid dumping mode: 1) Dump to extra dependency info files, such as
+// *.d files. This mode is used when we want to generate dependency info and
+// also compile. 2) Overwrite the original compilation output and dump
+// dependency info as compilation output. This mode is used when we do not want
+// to compile the source code and want the dependency info only.
+class DependencyInfoDumpingHandler {
+ public:
+  // Sets the dependency target explicitly. It's the same as the argument to
+  // -MT.
+  void SetTarget(const std::string& target_label) {
+    user_specified_dep_target_label_ = target_label;
+  }
+
+  // Sets the name of the file where dependency info will be written.
+  void SetDependencyFileName(const std::string& dep_file_name) {
+    user_specified_dep_file_name_ = dep_file_name;
+  };
+
+  // Dump depdendency info to a) an extra dependency info file, b) an string
+  // which holds the compilation output. The choice depends on the dump
+  // mode of the handler. Returns true if dumping is succeeded, false otherwise.
+  //
+  // The dependency file name and target are deduced based on 1) user
+  // specified dependency file name and target name, 2) the output filename when
+  // the compiler is in 'does not need linking' and 'not preprocessing-only'
+  // mode. It is passed through compilation_output_file_name.
+  //
+  // When the handler is set to dump dependency info as extra dependency info
+  // files, this method will open a file with the dependency file name and write
+  // the dependency info to it. Error messages caused by writing to the file are
+  // emitted to stderr.
+  //
+  // When the handler is set to dump dependency info as compilation output, the
+  // compilation output string, which is passed through compilation_output_ptr,
+  // will be cleared and this method will write dependency info to it. Then the
+  // dependency info should be emitted as normal compilation output.
+  //
+  // If the dump mode is not set when this method is called, return false.
+  bool DumpDependencyInfo(std::string compilation_output_file_name,
+                          std::string source_file_name,
+                          std::string* compilation_output_ptr,
+                          const std::unordered_set<std::string>& dependent_files);
+
+  // Sets to always dump dependency info as an extra file, instead of the normal
+  // compilation output. This means the output name specified by -o options
+  // won't be used for the dependency info file.
+  void SetDumpToExtraDependencyInfoFiles() { mode_ = dump_as_extra_file; };
+
+  // Sets to dump dependency info as normal compilation output. The dependency
+  // info will be either saved in a file with -o option specified file, or, if
+  // no output file name specified, to stdout.
+  void SetDumpAsNormalCompilationOutput() {
+    mode_ = dump_as_compilation_output;
+  }
+
+  // Returns true if the handler's dumping mode is set to dump dependency info
+  // as extra dependency info files.
+  bool DumpingToExtraDependencyInfoFiles() {
+    return mode_ == dump_as_extra_file;
+  }
+
+  // Returns true if the handler's dumping mode is set to dump dependency info
+  // as normal compilation output.
+  bool DumpingAsCompilationOutput() {
+    return mode_ == dump_as_compilation_output;
+  }
+
+  // Returns true if the handler's dumping mode is not set.
+  bool DumpingModeNotSet() { return mode_ == not_set; }
+
+  // Returns true if the handler is at valid state for dumping dependency info.
+  bool IsValid(std::string* error_msg_ptr, size_t num_files);
+
+ private:
+  typedef enum {
+    // not_set mode tells that the dumping mode is not set yet, so the handler
+    // is not ready for dumping dependency info. Calling DumpDependencyInfo when
+    // the handler is in this mode will cause failure.
+    not_set = 0,
+    // Dumping dependency info as normal compilation output mode.  In this mode,
+    // the dependency info will be dumped as compilation output by overwriting
+    // the string which holds the compilation output.
+    dump_as_compilation_output,
+    // Dumping dependency info as extra dependency info files mode. In this
+    // mode, dependency info will be dumped to a user specified dependency info
+    // file or a *.d file. Compilation output will still be generated along with
+    // the dependency info.
+    dump_as_extra_file,
+  } dump_mode;
+
+  // Returns the target file label to be used in depdendency info file. If -MT
+  // defined a label, use that string as the label. Otherwise returns the
+  // compilation output filename deduced in 'doesn't need linking' and 'not
+  // preprocessing-only' mode.
+  std::string GetTarget(const std::string& compilation_output_file_name);
+
+  // Returns the dependency file name to be used. If -MF defined a file name
+  // before, use it. Othwise, returns a filename formed by appending .d to the
+  // output filename deduced in 'doesn't need linking' and 'no
+  // preprocessing-only' mode.
+  std::string GetDependencyFileName(
+      const std::string& compilation_output_file_name);
+
+  std::string user_specified_dep_file_name_;
+  std::string user_specified_dep_target_label_;
+  dump_mode mode_;
+};
+}
+
+#endif  // GLSLC_DEPENDENCY_INFO_H
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -62,6 +62,9 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
 
   std::unique_ptr<FileIncluder> includer(
       new FileIncluder(&include_file_finder_));
+  // Get a reference to the dependency trace before we pass the ownership to
+  // shaderc::CompileOptions.
+  const auto& included_files = includer->file_path_trace();
   options_.SetIncluder(std::move(includer));
 
   shaderc::SpvModule result = compiler_.CompileGlslToSpv(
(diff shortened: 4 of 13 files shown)


glslc/src/dependency_info.h
```diff
@@ -0,0 +1,70 @@
...
+#define GLSLC_DEPENDENCY_INFO_H
+
+#include <set>
+#include <string>
+
+namespace glslc {
+class DependencyInfoDumpingHandler {
+ public:
+  // Sets -MT argument.
+  void SetDepTargetFileLabel(const std::string& target_label) {
```
dneto0: I'd just call this "SetDepTarget".  

It's simpler still accurate, and more in line with the first sentence of the the GCC doc for -MT
"Change the target of the rule emitted by dependency generation. "
https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#Preprocessor-Options 

Also, the comment would be more clear if you described what it's doing in English rather than just referencing "-MT".
You can say both. E.g.
"Sets the dependency target explicitly.  It's the same as the argument to -MT"

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,70 @@
...
+    dep_target_label_ = std::string(target_label);
```
dneto0: You don't need the  explicit constructor.

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  // Sets -MF argument.
+  void SetDepFileName(const std::string& dep_file_name) {
```
dneto0: Say "Sets the name of the file where dependency info will be written."

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,70 @@
...
+    dep_file_name_ = dep_file_name;
+  };
+
+  // Collects the dependency info and dump it to dependency info file.
```
dneto0: Does it do the collection, or does it just write the info.
Why do you need to set the target file if there's a separate method to do it?
What's the trace argument for?  Why is it called "trace"?

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,70 @@
...
+  bool DumpDependencyInfo(std::string target_file, std::string source_file_name,
+                          const std::set<std::string>& trace);
+
+  // -MM or -M will prevent emitting the output of compilation.
+  bool dump_dependency_info_only_ = false;
```
dneto0: These should be private members.  If you need a setter, add one with a proper blurb.

dneto0: Do you really need two booleans?
If dependency info is being generated, either
 (a) it's the main output of the compiler and we also set preprocess-only (the -M and -MM cases)
or  (b) it's written out separately, and the other parts of the compiler behave normally (the -MD and -MMD cases).

I have to see the higher level algorithm to see if it makes sense.

Qining: Hmm, I was thinking about this, but there is a corner case: when both -M and -MD are specified.

In Clang++, it makes -M overrides -MD and emits an warning to say -MD is not used.
In GCC, it generates dependency info files with name: *.d, and blocks compilation output.

In either way, we may need another flag, as we actually have three states, instead of two.
Clang: we need to know if both -MD and -M are specified, so we need one flag for each of them.
GCC: As you said, -M writes dependency info to main output and -MD writes out separately. But this means, in '-M -MD' cases, if we have -MD overrides -M, we will have compilation output, but if we let -M overrides -MD, we will have dependency info dumped in a wrong stream.

In general I think there are three states instead of two. I was wondering how do we want glslc to behave in such cases.

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+names of all the included files.
+
+Unless specified explicitly (with `-MT`), the SPIR-V object file name in the
+generated *make* rules follows the rules of  <<output-file-naming,Output File
+Naming>> as in `-c` compilation stage.
+
+The dependency info file name can be specified by `-o` and `-MF` options. When
+both are specified, `-o` option is ignored.
+
+Passing `-M` implies `-E`, and suppresses warnings with an implicit `-w`.
```
dneto0: Replace "Passing" with "Specifying"

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+
+`-MM` is an alias of `-M`.
```
dneto0: I think it reads better as "alias for" instead of "alias of"  ??
Not too worried about this, though.

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+
+E.g., `glslc -M main.vert` will dump `main.vert.spv: main.vert <other included
```
dneto0: The trouble with these examples is that we don't support compilation without -c.
Please update them to use -c.

Also, it will read better if you make it an explicit list, or a table.

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+files>` to stdout. `glslc -M main.vert -o dep_info` will generate a file
+`dep_info` to store the *make* rule of `main.vert`. `glslc -M main.vert -MT
+target` will have *make* rule with `target` as the object file name: `target:
+main.vert <other included files>`.
+
+==== `-MD`
+
+`-MD` generates *make* dependencies and compiles. Unlike `-M`, `-MD` does not
+implies `-E` and it outputs both compilation results and dependencies. If the
```
dneto0: "implies" -> "imply"

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+user specifies an dependency info file name by `-MF`, the compiler will take
```
dneto0: "take" -> "use"

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+that name, otherwise, the compiler will take the `-o` option if one is given,
```
dneto0: This sentence is too long, and I'm not sure it parses.

Suggest:
Dependencies are written to a file whose name is determined as follows:  If option -MF is specified, use its argument.  Otherwise, use the filename formed by append *.d* to the name of the file containing compilation results.

glslc/README.asciidoc
dneto0: Never mind.  @Qining pointed out it actually does work. :-)

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+as the dependency info file.
+
+==== `-MF`
+
+`-MF` lets you specify the dependency info file name when used with `-M` or
+`-MD`.
+
+==== `-MT`
+
+`-MT` let you specify the target of the rule emitted by dependency generation
```
dneto0: let -> lets

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+            "only one input file is allowed.";
+        return false;
+      }
+    }
+    return true;
+  }
+
+ private:
+  typedef enum {
+    // By default, dump mode should be set to not_set, which means the
```
dneto0: This is too much information for the definition of this enum value.
Youd can just say:  not_set means don't generate dependency info.
How and when it's *USED* is another matter, dealt with elsewhere.

dneto0: Maybe more clearly: not_set means capture the info but don't write it out?  I'm not sure because I haven't read the rest of the code yet.

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  // dependency info to them. Error messages caused by writing the files are
+  // emitted to stderr.
+  //
+  // When the handler is set to dump dependency info as compilation output, the
+  // compilation output string, which is passed through compilation_output_ptr,
+  // will be cleared and this method will write dependency info to it. Then the
+  // dependency info should be emitted as normal compilation output.
+  //
+  // If the dump mode is not set when this method is called, return false.
+  bool DumpDependencyInfo(std::string compilation_output_file_name,
```
dneto0: Always put the constructor first.

dneto0: Never mind.  That isn't the constructor.  :-)

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  } dump_mode;
+
+  // Deduce the target file label to be used in depdendency info file.
+  // If -MT defined a label, use that string as the label. Otherwise,
+  // if compilation_output_file_name is empty, "-" or "a.spv", the label
+  // should be deduced from source_file_name, otherwise, it should be
+  // compilation_output_file_name. If using the source_file_name as the
+  // label but source_file_name is an absolute path, we need to strip the
+  // directory name for the label.
+  std::string GetFinalTargetLabel(
```
dneto0: Can just be GetTarget  ?
The "Final" is just misleading.  The "Label" seems redundant.

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+
+  // Deduce the dependency file name to be used.
+  // If -MF defined a file name before, use it. Othwise, the dependency
+  // info file should be named compilation_output_file_name+".d", unless
+  // compilation_output_file_name is "-", "a.spv", in which case we should
+  // deduce dependency info file name from the source_file_name. When
+  // source_file_name is used for the deduection, ignore absolute path,
+  // relative paths and working directory, the file is always placed under
+  // current directory.
+  std::string GetFinalDepdencyFileName(
```
dneto0: Spelling:  GetDependencyFileName
Also, you don't need "Final"

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  // if compilation_output_file_name is empty, "-" or "a.spv", the label
+  // should be deduced from source_file_name, otherwise, it should be
+  // compilation_output_file_name. If using the source_file_name as the
+  // label but source_file_name is an absolute path, we need to strip the
+  // directory name for the label.
+  std::string GetFinalTargetLabel(
+      const std::string& compilation_output_file_name,
+      const std::string& source_file_name);
+
+  // Deduce the dependency file name to be used.
```
dneto0: Deduce -> Returns

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+    dump_as_compilation_output,
+    // Dumping dependency info as extra dependency info files mode. In this
+    // mode, dependency info will be dumped to user specified dependency
+    // info file or a *.d file whose name is deduced from main output file
+    // name or input file name. Compilation output will still be generated
+    // along with the dependency info.
+    dump_as_extra_file,
+  } dump_mode;
+
+  // Deduce the target file label to be used in depdendency info file.
```
dneto0: Deduce -> Returns

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+#include <set>
+#include <string>
+#include <string>
+
+namespace glslc {
+class DependencyInfoDumpingHandler {
+ public:
+  // Sets the dependency target explicitly. It's the same as the argument to
+  // -MT.
+  void SetDepTarget(const std::string& target_label) {
```
dneto0: SetTarget

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+    user_specified_dep_target_label_ = std::string(target_label);
+  }
+
+  // Sets the name of the file where dependency info will be written.
+  void SetDepFileName(const std::string& dep_file_name) {
```
dneto0: Be consistent between "Dep" and "Dependency" in your methods.
SetDependencyFileName

glslc/README.asciidoc
```diff
@@ -236,6 +236,66 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+main.vert -o shader_obj` will generate `shader_obj` as the SPIR-V object file
+and `shader_obj.d` as the dependency info file. `glslc -c -MD main.vert -MF
+dep_info` will generate `main.vert.spv` as the SPIR-V object file and `dep_info`
+as the dependency info file.
+
+==== `-MF`
+
+`-MF` lets you specify the dependency info file name when used with `-M` or
+`-MD`.
+
```
dneto0: What happens if I have multiple source files, and -c, and use -MF.
Should be an error.

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+#include <string>
+#include <string>
+
+namespace glslc {
+class DependencyInfoDumpingHandler {
+ public:
+  // Sets the dependency target explicitly. It's the same as the argument to
+  // -MT.
+  void SetDepTarget(const std::string& target_label) {
+    user_specified_dep_target_label_ = std::string(target_label);
```
dneto0: don't need the constuctor for std::string

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,129 @@
...
+#include "dependency_info.h"
+
+#include <fstream>
+#include <iostream>
+#include <set>
+#include <sstream>
+
+#include "file.h"
+#include "libshaderc_util/io.h"
+#include "libshaderc_util/string_piece.h"
```
dneto0: Remove this include if you didn't use string_piece

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,129 @@
...
+
+std::string DependencyInfoDumpingHandler::GetFinalTargetLabel(
+    const std::string& compilation_output_file_name,
+    const std::string& source_file_name) {
+  if (!user_specified_dep_target_label_.empty()) {
+    return user_specified_dep_target_label_;
+  }
+
+  if (mode_ == dump_as_extra_file && compilation_output_file_name != "-" &&
+      compilation_output_file_name != "a.spv") {
```
dneto0: What if someone did -o a.spv  
?

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,129 @@
...
+    // In dump_as_extra_file mode, the target follows the
+    // compilation_output_file_name if it is not '-' (means stdout) or 'a.spv'.
+    return compilation_output_file_name;
+  }
+
+  // In dump_as_compilation_output mode, the target name is always induced from
+  // the source file name, unless there is an user specified one.
+  // In dump_as_extra_file mode, only when a compilation_output_file_name is set
+  // to '-' or 'a.spv', we need to deduce the target name from source file name.
+  std::string source_file_base_name =
```
dneto0: This seems to duplicate logic elsewhere in glslc.  Why do we need to do that?

dneto0: If we needed this, then it should have been passed in as an extra argument from the outside.  We should do this complex computation only once in glslc.

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  void SetDumpToExtraDependencyInfoFiles() { mode_ = dump_as_extra_file; };
+
+  // Sets to dump dependency info as normal compilation output. The dependency
+  // info will be either saved in a file with -o option specified file, or, if
+  // no output file name specified, to stdout.
+  void SetDumpAsNormalCompilationOutput() {
+    mode_ = dump_as_compilation_output;
+  }
+
+  // Checks if the handler's dumping mode is set to dump dependency info as
```
dneto0: Returns true if...

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  // extra dependency info files.
+  bool DumpingToExtraDependencyInfoFiles() {
+    return mode_ == dump_as_extra_file;
+  }
+
+  // Checks if the handler's dumping mode is set to dump dependency info as
```
dneto0: Returns true if...

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  // normal compilation output.
+  bool DumpingAsCompilationOutput() {
+    return mode_ == dump_as_compilation_output;
+  }
+
+  // Checks if the handler's dumping mode is not set.
```
dneto0: Returns true if...

glslc/src/dependency_info.h
```diff
@@ -0,0 +1,163 @@
...
+  bool DumpingModeNotSet() { return mode_ == not_set; }
+
+  // Checks if the handler is at valid state for dumping dependency info.
```
dneto0: Returns true if ...

dneto0: Also, this body is big enough that it doesn't belong in the header.

glslc/src/file_includer.h
```diff
@@ -45,6 +49,7 @@ class FileIncluder : public shaderc::CompileOptions::IncluderInterface {
   shaderc_includer_response response_;
   std::vector<char> file_content_;
   std::string file_full_path_;
+  std::set<std::string> file_path_trace_;
```
dneto0: Say what this is.  Does order matter, or not?
I suspect it should be called source_files_used_

dneto0: Are paths absolute?

glslc/src/main.cc
```diff
@@ -59,6 +59,12 @@ void PrintHelp(std::ostream* out) {
        << "                    are concatenations of version and profile, e.g. "
        << "310es,\n"
        << "                    450core, etc.\n"
+       << "  -M                Generate make dependencies.\n"
```
dneto0: Could also say.  "Implies -E."

glslc/src/main.cc
```diff
@@ -59,6 +59,12 @@ void PrintHelp(std::ostream* out) {
...
+       << "  -MD               Generate make dependencies and compile.\n"
+       << "  -MF <file>        Write dependency output to the given file.\n"
+       << "  -MT <target>      Change the target of the rule emitted by "
```
dneto0: Say "Specify" instead of "Change"

glslc/src/dependency_info.cc
Qining: The problem is when 'need_linking_' is true, the output file name passed in here is "a.spv". In this case, the output file name is not computed out by the complex logic, but assigned directly. So here we need to append ".spv.d " to the source file name to get the dependency file name.

If we want to pass the name from outside, I think we will need to split the FileCompiler::GetOutputFile() method, to always generate a "Default output file name in case of -c (w/o -o)" and then pass that name to dependency info handler. FileCompiler will need to check the 'need_linking_' flag and '-o' option out of GetOutputFile() to determine if "a.spv" should overwrite the output file name. Inside dependency handler, it will be as simple as "if -MF set, use it; else output_file_name+".d". 

BTW, the substr() line should be replaced by GetBaseFileName(), this is not related to the dep file extension.

glslc/src/file_includer.h
Qining: Current paths are not absolute, so in some cases we can not behave as gcc/clang. When input file name is represented in absolute path, gcc/clang always use absolute for the dependent file names in the .d files. If the input files are relative paths, gcc/clang uses relative paths too.

Because we don't have context aware includer yet, I think the handler can not support this function right now as the dependency info dumping handler just use the filenames "as is" and has no idea about files.

Currently, when we an input file which does not include other files, and is represented in absolute path, we can have its absolute path in the .d file, because the handler uses the input source file name "as is", and it is not from the record of includer.

Qining: I think the order doesn't matter, so can we use 'unordered_set' instead? I used 'set' as it can ease a little bit on the testing side as we are using string comparison and order does matter in such case. But we can always change this.

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,160 @@
...
+std::string DependencyInfoDumpingHandler::GetDependencyFileName(
+    const std::string& compilation_output_file_name
+    /* const std::string& source_file_name */) {
+  if (!user_specified_dep_file_name_.empty()) {
+    return user_specified_dep_file_name_;
+  }
+
+  return compilation_output_file_name + ".d";
+
+  // if (mode_ == dump_as_compilation_output) {
```
dneto0: Delete dead code.

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,160 @@
...
+  // ? source_file_base_name + ".spv"
+  // : source_file_base_name.substr(
+  // 0, source_file_base_name.find_last_of('.')) +
+  // ".spv";
+  // return label;
+}
+
+std::string DependencyInfoDumpingHandler::GetDependencyFileName(
+    const std::string& compilation_output_file_name
+    /* const std::string& source_file_name */) {
```
dneto0: Delete dead arguments.

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,160 @@
...
+std::string DependencyInfoDumpingHandler::GetTarget(
+    const std::string& compilation_output_file_name
+    /* const std::string& source_file_name */) {
+  if (!user_specified_dep_target_label_.empty()) {
+    return user_specified_dep_target_label_;
+  }
+
+  return compilation_output_file_name;
+
+  // if (mode_ == dump_as_extra_file && compilation_output_file_name != "-" &&
```
dneto0: Delete dead code.

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,160 @@
...
+  } else {
+    // mode_ should not be 'not_set', we should never be here.
+    return false;
+  }
+  return true;
+}
+
+std::string DependencyInfoDumpingHandler::GetTarget(
+    const std::string& compilation_output_file_name
+    /* const std::string& source_file_name */) {
```
dneto0: Delete dead arguments.

glslc/src/file_compiler.cc
```diff
@@ -62,6 +62,9 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
 
   std::unique_ptr<FileIncluder> includer(
       new FileIncluder(&include_file_finder_));
+  // Get an reference to the dependency trace before we pass the ownership to
```
dneto0: "a reference"

glslc/src/file_compiler.cc
```diff
@@ -96,8 +99,24 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
     return false;
   }
 
+  // Cache the compilation output, its content might be overwritten by
```
dneto0: This always makes a copy of the compilation results, even in the case there is no need.

you could avoid the copy as follows.
compilation_output could be a stringpiece that, for now points to the result.GetData(), result.GetLength.
The return value of DumpDependencyInfo could be a std::pair<bool, std::string>,
where the first part is the current return value (pass/fail), and the second part is the dependency output string if it's generated as compilation output, and an empty string otherwise.
Then if that second parameter is non-empty, then set the string-piece to point to that string.  (But make sure the storage stays around long enough to e used when dumping the output. That is, declare the std::pair result variable before checking if the dependency info handler exists.)

That way you never make a copy of the compilation output when you don't need to.

glslc/test/option_dash_M.py
```diff
@@ -0,0 +1,669 @@
...
+    generated dependency info.
+    e.g. glslc -M shader.vert -MT target
+      => target: shader.vert
+    """
+    environment = EMPTY_SHADER_IN_CURDIR
+    glslc_args = ['-M', 'shader.vert', '-MT', 'target']
+    dependency_rules_expected = [{'target': 'target',
+                                  'dependency': {'shader.vert'}}]
+
+# TODO:(qining) Add tests for absolute input path with #include directives, the
```
dneto0: Good TODO

glslc/test/option_dash_M.py
```diff
@@ -0,0 +1,669 @@
...
+    def check_stdout_dependency_info(self, status):
+        # Add the absolute path of include/b.vert to the dependency set
+        self.dependency_rules_expected[0]['dependency'].add(os.path.dirname(
+            self.shader_main.filename) + '/include/b.vert')
+        return DependencyInfoStdoutMatch.check_stdout_dependency_info(self,
+                                                                      status)
+
+
+@inside_glslc_testsuite('OptionsCapM')
+class TestDashCapMOverridesOtherModes(DependencyInfoStdoutMatch):
```
dneto0: Good!

glslc/test/option_dash_M.py
```diff
@@ -0,0 +1,669 @@
...
+    e.g. glslc -MD shader.vert
+      => <a.spv: valid SPIR-V object file>
+      => <shader.vert.spv.d: dependency info>
+    """
+    environment = EMPTY_SHADER_IN_CURDIR
+    glslc_args = ['-MD', 'shader.vert']
+    expected_object_filenames = ('a.spv', )
+    target_filename = 'shader.vert.spv.d'
+    expected_file_contents = ['shader.vert.spv: shader.vert\n']
+
```
dneto0: While I think of it: Are there tests that check that -M implies -E and -w ?

glslc/test/option_dash_M.py
```diff
@@ -0,0 +1,669 @@
...
+    """
+    environment = EMPTY_SHADER_IN_CURDIR
+    glslc_args = ['-M', '-MD', 'shader.vert']
+    expected_stderr = ['glslc: error: both -M (or -MM) and -MD are specified. '
+                       'Only one should be used at one time.\n']
+
+
+@inside_glslc_testsuite('OptionsCapM')
+class TestErrorDashCapMFWithMultipleInputFiles(expect.StderrMatch):
+    """Tests that when -MF option is specified, only one input file should be
```
dneto0: Comment is wrong.  Say it's an error to use -MF when you have two source files.

dneto0: Nope. I read it wrong. What you have is right.

glslc/test/parameter_tests.py
```diff
@@ -68,6 +68,11 @@ class HelpParameters(
   -std=<value>      Version and profile for input files. Possible values
                     are concatenations of version and profile, e.g. 310es,
                     450core, etc.
+  -M                Generate make dependencies. Implies -E and -w.
```
dneto0: Hmnm. Should mention that -MM is an alias for -M

dneto0:
Lots of progress!
Nearly there.  Could use a simplification of the output_filename computation.
And seems to be missing a test that ensures that -M (and -MM) implies -E and -w.

You've done a great job at testing: Lots of case combinations, and clearly written tests.  You did a good job abstracting away the mechanics with helper functions, so each case is almost all clearly-spelled out data!

glslc/src/file_compiler.cc
Qining: I have updated it with using string_piece to refer to the compilation_output to be emitted, and use another string to hold the dump info, passing the string's address and then check if the string is empty or not to determine whether we should overwrite the string_piece compilation_output. This is roughly the same as returning a pair, just didn't change the signature of DumpDependencyInfo(). 

Now I sort of understand having two functions to 'get the data to dump' and 'dump it' might be better than having only one function.

dneto0: Yes, we call that "separation of concerns".  :-)

glslc/src/dependency_info.cc
```diff
@@ -0,0 +1,102 @@
...
+std::string DependencyInfoDumpingHandler::GetTarget(
+    const std::string& compilation_output_file_name) {
+  if (!user_specified_dep_target_label_.empty()) {
+    return user_specified_dep_target_label_;
+  }
+
+  return compilation_output_file_name;
+}
+
+std::string DependencyInfoDumpingHandler::GetDependencyFileName(
```
dneto0: Soooo much nicer now.

glslc/src/file_compiler.cc
```diff
@@ -96,8 +99,34 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
     return false;
   }
 
-  // Write output to output file.
-  output_stream->write(result.GetData(), result.GetLength());
+  // Get a string_piece which refers to the normal compilation output for now.
+  // This string_piece might be overwritten by dependency info dumping handler
+  // later, if the handler is instantiated to dump as normal compilation output,
+  // and the original compilation output should be blocked. Otherwise it won't
+  // be touched. The main output stream dump this string_piece later.
```
dneto0: dump -> dumps

glslc/src/file_compiler.cc
```diff
@@ -96,8 +99,34 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
...
+  string_piece compilation_output(result.GetData(),
+                                  result.GetData() + result.GetLength());
+
+  // If we have dependency info dumping handler instantiated, we should dump
+  // dependency info first, this may overwrite the compilation output
```
dneto0: End the sentence after "first"  "This may..." is a new sentence.

dneto0: Just a minor wording fix to make it more clear:
Isn't it that the compilation_output might be redirected to reference the contents of this string?  That's not quite saying that compilation output string piece will be overwritten (The way you wrote it sounds like the *contents* of the string piece will be overwritten, i.e. the storage pointed to by the string piece.)

glslc/src/main.cc
```diff
@@ -184,6 +192,49 @@ int main(int argc, char** argv) {
     } else if (arg == "-E") {
       compiler.options().SetPreprocessingOnlyMode();
       compiler.SetPreprocessingOnlyFlag();
+    } else if (arg == "-M" || arg == "-MM") {
+      // -M implies -E and -w
+      compiler.options().SetPreprocessingOnlyMode();
```
dneto0: It seems like an API problem that you'd have to call both options().SetPreprocessingOnlyMode *and* SetPreprocessingOnlyFlag.
It seems that you should only have to call SetPreprocessingOnlyFlag and it should take care of updating its internal options object.

But not for this CL.  It's a pre-existing problem.

glslc/test/option_dash_M.py
```diff
@@ -0,0 +1,738 @@
...
+    """Tests that -M implies -E, a .glsl file without an explict stage should
+    not generate an error.
+    e.g. glslc -M shader.glsl
+      => shader.spv: shader.glsl
+         <no error message should be generated>
+    """
+    environment = Directory('.', [File('shader.glsl', MINIMAL_SHADER)])
+    glslc_args = ['-M', 'shader.glsl']
+    dependency_rules_expected = [{'target': 'shader.spv',
+                                  'dependency': {'shader.glsl'}}]
```
dneto0: I think you need a expected_stdout = (some string) so that you actually ensure the output of the compiler is preprocessed output.

dneto0: Woops.  I was wrong.  You explained it in the comment, and I forgot how the dependency_rules_expected mixin variable worked. (It's been a long day)

dneto0:
Looks good to me.
Please squash one more time and remove the WIP from the commit.

glslc/README.asciidoc
```diff
@@ -236,6 +236,80 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
 a non-zero exit code from `glslc`. If `-w` is specified the warnings
 generated are suppressed before they are converted to errors.
 
+=== Dependency Generation Options
+
+==== `-M` or `-MM`
+
+`-M` generates *make* dependencies, outputs a rule suitable for *make*
```
dekimir: "," => "."
"outputs" => "It outputs"

Qining: Just created a PR #128 to fix this.

glslc/README.asciidoc
```diff
@@ -236,6 +236,80 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
...
+E.g., `glslc -M main.vert` will dump `main.vert.spv: main.vert <other included
+files>` to stdout. More examples are listed in
+<<dependency-generation-examples,Dependency Generation Examples>>
+
+==== `-MD`
+
+`-MD` tells the glslc compiler to both compile the source and generate *make*
+dependencies. Dependencies are written to a file whose name is determined as
+follows: If option `-MF` is specified, use its argument. Otherwise, use the
+filename formed by append *.d* to the name of the file containing compilation
```
dekimir: "appending"

glslc/README.asciidoc
```diff
@@ -236,6 +236,80 @@ that all `warning:` messages are shown as `error:` and any warnings will cause
 a non-zero exit code from `glslc`. If `-w` is specified the warnings
 generated are suppressed before they are converted to errors.
 
+=== Dependency Generation Options
+
+==== `-M` or `-MM`
+
+`-M` generates *make* dependencies, outputs a rule suitable for *make*
+describing the dependencies of the main source file. Instead of outputting
```
dekimir: "the main source file" is not defined.

PR #124, opened 2016-02-06
Drop -working-directory command line option

Qining:
Disable working directory tests; Modified --help messages;

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -96,24 +96,6 @@ Otherwise,
 `-o` lets you specify the output file's name. It cannot be used when there are
 multiple files generated. A filename of `-` represents standard output.
 
-==== `-working-directory`
-
-`-working-directory[=]` sets the working directory for the compilation.
-Paths/filenames in the following categories will be resolved against the
-argument of `-working-directory`:
-
-* Input files under *relative* paths.
-* Intermediate output files (files generated via the `-c` or `-S` option).
-** Including the filename under *relative* paths given by the `-o` option.
-* Non-system include files (files ``#include``d using double-quotation marks)
-  under *relative* paths.
-
-Output files will be generated into the current working directory if no
-<<compilation-stage-selection-options,compilation stage selection option>> is
-given.
-
-WARNING: Interaction with `#include` and `-I` is not implemented yet.
-
 === Language and Mode Selection Options
 
 [[option-f-shader-stage]]
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -34,9 +34,6 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
                                      shaderc_shader_kind shader_stage) {
   std::vector<char> input_data;
   std::string path = input_file;
-  if (!workdir_.empty() && !shaderc_util::IsAbsolutePath(path)) {
-    path = workdir_ + path;
-  }
   if (!shaderc_util::ReadFile(path, &input_data)) {
     return false;
   }
@@ -63,12 +60,13 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
                                  &(*input_data.begin()) + input_data.size());
   }
 
-  std::unique_ptr<FileIncluder> includer(new FileIncluder(&include_file_finder_));
+  std::unique_ptr<FileIncluder> includer(
+      new FileIncluder(&include_file_finder_));
   options_.SetIncluder(std::move(includer));
 
-  shaderc::SpvModule result =
-      compiler_.CompileGlslToSpv(source_string.data(), source_string.size(),
-                                 shader_stage, error_file_name.data(), options_);
+  shaderc::SpvModule result = compiler_.CompileGlslToSpv(
+      source_string.data(), source_string.size(), shader_stage,
+      error_file_name.data(), options_);
   total_errors_ += result.GetNumErrors();
   total_warnings_ += result.GetNumWarnings();
 
@@ -115,11 +113,6 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
   }
 
   return compilation_success;
-}
-
-void FileCompiler::SetWorkingDirectory(const std::string& dir) {
-  workdir_ = dir;
-  if (!dir.empty() && dir.back() != '/') workdir_.push_back('/');
 }
 
 void FileCompiler::AddIncludeDirectory(const std::string& path) {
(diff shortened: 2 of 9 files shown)


glslc/README.asciidoc
```diff
@@ -96,25 +96,7 @@ Otherwise,
...
-  under *relative* paths.
-
-Output files will be generated into the current working directory if no
-<<compilation-stage-selection-options,compilation stage selection option>> is
-given.
-
-WARNING: Interaction with `#include` and `-I` is not implemented yet.
-
-=== Language and Mode Selection Options
+:== Language and Mode Selection Options
```
dekimir: Revert this line.  "Language and Mode Selection Options" should be the same level as "Overall Options".

glslc/src/file_compiler.cc
```diff
@@ -177,22 +170,28 @@ void FileCompiler::OutputMessages() {
   shaderc_util::OutputMessages(&std::cerr, total_warnings_, total_errors_);
 }
 
+namespace {
+// A helper function to return the base file name from either absolute path or
+// relative path representation of a file.
+// e.g.: dir_a/dir_b/file_c.vert => file_c.vert
+std::string GetBaseFileName(const std::string& file_path) {
+  return file_path.substr(file_path.find_last_of("/\\") + 1);
```
dekimir: Move this to `libshaderc_util::io` and test properly.  This version doesn't work for `file_path` with no backslashes or ending in backslash.

dekimir: Actually, it probably works fine, because `1` is signed, so a potential `npos` is converted to `-1` before the addition.  Still, this belongs in `io`, and we should have tests for it.

dekimir: **Actually** `npos` will be converted to `-1` only if `string::size_type` bit width equals `int` bit width, which isn't guaranteed by the standard.

glslc/src/file_compiler.cc
```diff
@@ -177,22 +170,28 @@ void FileCompiler::OutputMessages() {
...
+      needs_linking_
+          ? "a.spv"
+          : (IsStageFile(input_filename)
+                 ? GetBaseFileName(input_filename) + file_extension_
+                 : GetBaseFileName(input_filename.substr(
+                                       0, input_filename.find_last_of('.')) +
+                                   file_extension_));
+  if (!output_file_name_.empty()) {
+    final_output_file_name =
+        std::string(output_file_name_.data(), output_file_name_.size());
```
dekimir: Why not `output_file_name_.str()`?

glslc/src/file_compiler.h
```diff
@@ -99,13 +96,26 @@ class FileCompiler {
...
+  //
+  //  If linking is not required, and the input file name does not have a
+  //  standard stage extension, then also returns the directory-stripped input
+  //  filename, but replace its extension with the result extension. (If the
+  //  resolved input filename does not have an extension, then append the result
+  //  extension.)
+  //
+  //  If linking is required and output filename is not specified, returns
+  //  "a.spv".
   std::string GetOutputFileName(std::string input_filename);
```
dekimir: Methods come before members: https://google.github.io/styleguide/cppguide.html#Declaration_Order

glslc/src/file_compiler.h
```diff
@@ -99,13 +96,26 @@ class FileCompiler {
   // compiler_.CompileGlslToSpv().
   shaderc::CompileOptions options_;
 
-  // Returns the name of the output file, given the input_filename string.
+  // Returns the final file name to be used for the output file.
+  // 
+  // If an output file name is specified by the argument of -o option, use that
```
dekimir: `FileCompiler` knows nothing about `-o`.  I think you meant `SetOutputFileName()`?

glslc/src/file_compiler.h
```diff
@@ -99,13 +96,26 @@ class FileCompiler {
...
+  // argument as the final output file name. 
+  //
+  // If the user did not specified an output filename:
```
dekimir: "specified" => "specify"

glslc/src/file_compiler.h
```diff
@@ -99,13 +96,26 @@ class FileCompiler {
...
+  //  If linking is not required, and the input filename has a
+  //  standard stage extension (e.g. .vert) then returns the input filename
+  //  withoug directory names but add the result extenstion (e.g. .spv or .s).
```
dekimir: Consider instead: "but with the result extension (eg, .spv or .s) appended".

glslc/src/file_compiler.h
```diff
@@ -99,13 +96,26 @@ class FileCompiler {
...
+  //
+  //  If linking is not required, and the input file name does not have a
+  //  standard stage extension, then also returns the directory-stripped input
+  //  filename, but replace its extension with the result extension. (If the
```
dekimir: "replace" => "replaces"

glslc/src/file_compiler.h
```diff
@@ -99,13 +96,26 @@ class FileCompiler {
...
+  //  resolved input filename does not have an extension, then append the result
```
dekimir: "append" => "appends"

glslc/README.asciidoc
```diff
@@ -96,25 +96,7 @@ Otherwise,
...
-  under *relative* paths.
-
-Output files will be generated into the current working directory if no
-<<compilation-stage-selection-options,compilation stage selection option>> is
-given.
-
-WARNING: Interaction with `#include` and `-I` is not implemented yet.
-
-=== Language and Mode Selection Options
+:=== Language and Mode Selection Options
```
dneto0: Why does this start with a colon now?

libshaderc_util/src/io.cc
```diff
@@ -56,6 +56,16 @@ bool IsAbsolutePath(const std::string& path) {
   return false;
 }
 
+std::string GetBaseFileName(const std::string& file_path) {
+  size_t loc_slash = file_path.find_last_of("/\\");
+  std::string base_name =
+      file_path.substr((loc_slash == std::string::npos ? -1 : loc_slash) + 1);
+  if (base_name == ".." || base_name == ".") {
```
dneto0: This is not in the spec.

libshaderc_util/src/io_test.cc
```diff
@@ -59,6 +60,32 @@ TEST(IsAbsolutePathTest, Windows) {
   EXPECT_FALSE(IsAbsolutePath(R"(C:..\File.txt)"));
```
dneto0: Did you mean \\F here?

dneto0: No. My mistake. Raw strings don't need escapes.

dneto0:
-1:

- Why the funny business with . and .. in GetBaseFileName?
- Why the colon at the front of the line in the .asciidoc?

dneto0:
+2

Issue #129, opened 2016-02-14
shaderc_module_get_success has no implementation


pixeljetstream:
shaderc_module_get_success only exists in header

dneto0:
Whoops.  That was supposed to be removed when we moved to 
shaderc_module_get_compilation_status.  Check for resulting value shaderc_compilation_status_success.

antiagainst:
Done via #131.

PR #131, opened 2016-02-14
Remove shaderc_module_get_success

dneto0:
Instead use shaderc_module_get_status and check for return value
shaderc_compilation_status_success.

Addresses https://github.com/google/shaderc/issues/129


dekimir:
+2

PR #133, opened 2016-02-16
WIP: Fix output data alignment and avoid copy

Qining:
Change the output data container of libshaderc compilation result from
std::string to std::vector<uint32_t> to guarantee the output data alignment.

Change the libshaderc_util::Compiler::Compile() signature and
implementation to avoid extra copies of SPIR-V binary. Also store the
string data of text output in vector<uint32_t>.

### DIFF
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -315,9 +315,10 @@ shaderc_spv_module_t shaderc_compile_into_spv(
   }
   result->compilation_status = shaderc_compilation_status_invalid_stage;
   bool compilation_succeeded = false;  // In case we exit early.
+  std::vector<uint32_t> compilation_output_data;
+  size_t compilation_output_data_size_in_bytes = 0u;
   if (!compiler->initializer) return result;
   TRY_IF_EXCEPTIONS_ENABLED {
-    std::stringstream output;
     std::stringstream errors;
     size_t total_warnings = 0;
     size_t total_errors = 0;
@@ -327,27 +328,33 @@ shaderc_spv_module_t shaderc_compile_into_spv(
         shaderc_util::string_piece(source_text, source_text + source_text_size);
     StageDeducer stage_deducer(shader_kind);
     if (additional_options) {
-      compilation_succeeded = additional_options->compiler.Compile(
-          source_string, forced_stage, input_file_name_str,
-          // stage_deducer has a flag: error_, which we need to check later.
-          // We need to make this a reference wrapper, so that std::function
-          // won't make a copy for this callable object.
-          std::ref(stage_deducer),
-          InternalFileIncluder(additional_options->get_includer_response,
-                               additional_options->release_includer_response,
-                               additional_options->includer_user_data),
-          &output, &errors, &total_warnings, &total_errors,
-          compiler->initializer);
+      // Depends on return value optimization to avoid extra copy.
+      std::tie(compilation_succeeded, compilation_output_data,
+               compilation_output_data_size_in_bytes) =
+          additional_options->compiler.Compile(
+              source_string, forced_stage, input_file_name_str,
+              // stage_deducer has a flag: error_, which we need to check later.
+              // We need to make this a reference wrapper, so that std::function
+              // won't make a copy for this callable object.
+              std::ref(stage_deducer),
+              InternalFileIncluder(
+                  additional_options->get_includer_response,
+                  additional_options->release_includer_response,
+                  additional_options->includer_user_data),
+              &errors, &total_warnings, &total_errors, compiler->initializer);
     } else {
       // Compile with default options.
-      compilation_succeeded = shaderc_util::Compiler().Compile(
-          source_string, forced_stage, input_file_name_str,
-          std::ref(stage_deducer), InternalFileIncluder(), &output, &errors,
-          &total_warnings, &total_errors, compiler->initializer);
+      std::tie(compilation_succeeded, compilation_output_data,
+               compilation_output_data_size_in_bytes) =
+          shaderc_util::Compiler().Compile(
+              source_string, forced_stage, input_file_name_str,
+              std::ref(stage_deducer), InternalFileIncluder(), &errors,
+              &total_warnings, &total_errors, compiler->initializer);
     }
 
     result->messages = errors.str();
-    result->spirv = output.str();
+    result->output_data = std::move(compilation_output_data);
+    result->output_data_size = compilation_output_data_size_in_bytes;
     result->num_warnings = total_warnings;
     result->num_errors = total_errors;
     if (compilation_succeeded) {
@@ -368,7 +375,7 @@ shaderc_spv_module_t shaderc_compile_into_spv(
 }
 
 size_t shaderc_module_get_length(const shaderc_spv_module_t module) {
-  return module->spirv.size();
+  return module->output_data_size;
 }
 
 size_t shaderc_module_get_num_warnings(const shaderc_spv_module_t module) {
@@ -380,7 +387,7 @@ size_t shaderc_module_get_num_errors(const shaderc_spv_module_t module) {
 }
 
 const char* shaderc_module_get_bytes(const shaderc_spv_module_t module) {
-  return module->spirv.c_str();
+  return reinterpret_cast<const char*>(module->output_data.data());
 }
 
 void shaderc_module_release(shaderc_spv_module_t module) { delete module; }
--- a/libshaderc/src/shaderc_cpp_test.cc
+++ b/libshaderc/src/shaderc_cpp_test.cc
@@ -530,7 +530,7 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
               HasSubstr("void main(){ }"));
 }
 
-// A shader kind test cases needs: 1) A shader text with or without #pragma
+// A shader kind test case needs: 1) A shader text with or without #pragma
 // annotation, 2) shader_kind.
 struct ShaderKindTestCase {
   const char* shader_;
--- a/libshaderc/src/shaderc_private.h
+++ b/libshaderc/src/shaderc_private.h
@@ -23,8 +23,14 @@
 
 // Described in shaderc.h.
 struct shaderc_spv_module {
-  // SPIR-V binary.
-  std::string spirv;
+  // Compilation output data. In normal compilation mode, it contains the
+  // compiled SPIR-V binary code. In disassembly and preprocessing-only mode, it
+  // contains a null-terminated string which is the text output. For text
+  // output, extra bytes with value 0x00 might be appended to complete the last
+  // uint32_t element.
+  std::vector<uint32_t> output_data;
+  // The size of the output data in term of bytes.
+  size_t output_data_size;
   // Compilation messages.
   std::string messages;
   // Number of errors.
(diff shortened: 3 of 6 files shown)


libshaderc_util/src/compiler.cc
```diff
@@ -103,20 +103,31 @@ spv_result_t DisassembleBinary(const std::vector<uint32_t>& binary,
   return result;
 }
 
+// Coverts a string to a vector of uint32_t by coping the content of a given
+// string to a vector<uint32_t> and return it. Appends '\0' at the end if extra
+// bytes are required to complete the last element.
+std::vector<uint32_t> ConvertStringToVector(std::string& str) {
```
Qining: Should I pull it out from here and place it with some sort of 'util' functions? I think this function is only used here and it is not relevant any external code, except STL, so it might be fine to live here?

libshaderc/src/shaderc.cc
```diff
@@ -315,9 +315,10 @@ shaderc_spv_module_t shaderc_compile_into_spv(
   }
   result->compilation_status = shaderc_compilation_status_invalid_stage;
   bool compilation_succeeded = false;  // In case we exit early.
+  std::vector<uint32_t> compilation_output_data;
+  size_t compilation_output_data_size = 0u;
```
dneto0: Should say this is in bytes.

libshaderc/src/shaderc_private.h
```diff
@@ -23,8 +23,10 @@
 
 // Described in shaderc.h.
 struct shaderc_spv_module {
-  // SPIR-V binary.
-  std::string spirv;
+  // Compilation output data.
```
dneto0: Should say more about what kinds of things this could contain. 
It can be a vector of SPIR-V binary module words, or a string (null terminated or not?).
And when it is a string, do we guarantee what the bytes are after the "size" bytes but within the last elements of the vector?

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -168,17 +167,23 @@ class Compiler {
...
-               EShLanguage forced_shader_stage, const std::string& error_tag,
-               const std::function<EShLanguage(std::ostream* error_stream,
-                                               const shaderc_util::string_piece&
-                                                   error_tag)>& stage_callback,
-               const CountingIncluder& includer, std::ostream* output_stream,
-               std::ostream* error_stream, size_t* total_warnings,
-               size_t* total_errors,
-               GlslInitializer* initializer) const;
+  //
+  // Returns a tuple consists of three fields. 1) a boolean which is true when
```
dneto0: "consisting"

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -168,17 +167,23 @@ class Compiler {
...
+  // the compilation succeeded, and false otherwise; 2) a vector of 32-bit words
+  // which contains the compilation output data, either compiled SPIR-V binary
+  // code, or the text string generated in preprocessing-only or disassembly
+  // mode; 3) the size of the output data. When the output is SPIR-V binary
+  // code, the size will always be N times of four. If the output is a text
```
dneto0: For (3) say whether the size is in bytes or not.  And when it's a string, say whether the size includes a null terminator, or not.

libshaderc_util/src/compiler.cc
```diff
@@ -103,20 +103,31 @@ spv_result_t DisassembleBinary(const std::vector<uint32_t>& binary,
...
+  size_t num_bytes_str = str.size() + 1u;
+  size_t vector_length = num_bytes_str / sizeof(uint32_t) +
```
dneto0: I'm ok with this. 
Alternate simpler formulation is:
```
  (num_bytes_str + sizeof(uint32_t)-1) / sizeof(uint32_t);
```

libshaderc_util/src/compiler.cc
```diff
@@ -103,20 +103,31 @@ spv_result_t DisassembleBinary(const std::vector<uint32_t>& binary,
...
+                         ((num_bytes_str % sizeof(uint32_t) == 0) ? 0 : 1);
+  std::vector<uint32_t> ret(vector_length, 0);
```
dneto0: Would be clearer if you used a real word for the variable, e.g. "result"

libshaderc_util/src/compiler.cc
```diff
@@ -140,7 +151,7 @@ bool Compiler::Compile(
                                    /* suppress_warnings = */ true,
                                    glslang_errors.c_str(), total_warnings,
                                    total_errors);
-    if (!success) return false;
+    if (!success) return std::make_tuple(false, std::vector<uint32_t>(), 0u);
```
dneto0: This std::make_tuple expression is repeated a lot.
You could make a const auto error_result = std::tuple(false, std::vector<uint32_t>(), 0u) and then use that all over the place.

libshaderc_util/src/compiler.cc
```diff
@@ -209,15 +220,13 @@ bool Compiler::Compile(
...
-    return shaderc_util::WriteFile(output_stream, text_or_error);
+    return std::make_tuple(true, ConvertStringToVector(text_or_error),
+                           text_or_error.size());
   } else {
-    return shaderc_util::WriteFile(
-        output_stream,
-        string_piece(
-            reinterpret_cast<const char*>(spirv.data()),
-            reinterpret_cast<const char*>(&spirv.back()) + sizeof(uint32_t)));
+    return std::make_tuple(true, std::move(spirv),
```
dneto0: Sorry to raise this now.  But does this really save the copy for the case of a binary SPIR-V module?  I'm concerned because you've populated the spirv vector, but the only copy-elision that is done is on the tuple, not the spirv vector.
It seems that to get copy elision you'd have to have made a std::tuple variable in the first place where you currently have the spirv variable definition.

dneto0: See http://en.cppreference.com/w/cpp/language/copy_elision

libshaderc_util/src/compiler.cc
```diff
@@ -209,15 +220,13 @@ bool Compiler::Compile(
...
+                           spirv.size() * sizeof(uint32_t));
```
dneto0: Sligthly prefer sizeof(spirv[0]) here.

libshaderc_util/src/compiler_test.cc
```diff
@@ -92,25 +92,36 @@ class DummyCountingIncluder : public shaderc_util::CountingIncluder {
...
-  // source for the given shader stage.  No includes are permitted, and shader
-  // stage deduction falls back to an invalid shader stage.
-  bool SimpleCompilationSucceeds(std::string source, EShLanguage stage) {
+  // Returns a tuple consists of two fields. 1) a boolean whose value is true if
+  // the compilation of the given shader source for the given shader stage
+  // succeeded, false otherwise; 2) the compilation output data in a vector of
+  // 4-byte words.
+  // No includes are permitted, and shader stage deduction falls back to an
+  // invalid shader stage.
+  std::tuple<bool, std::vector<uint32_t> > SimpleCompilationSucceeds(
```
dneto0: There was no reason to change the signature of SimpleCompilationSucceeds.  You only ever want the boolean answer.  In the changes below, you only ever use std::ignore for the second part of the result.

dneto0:
-1

libshaderc_util/src/compiler.cc
Qining: According to my experiment results, with 'std::move' => save a copy, without 'std::move' => extra copy.

Qining: It seems that once we use make_tuple (or pair) for the return value, we lost the NRVO. I think here I should use std::move for all the vector returned in the tuple.

libshaderc_util/src/compiler.cc
```diff
@@ -103,20 +103,33 @@ spv_result_t DisassembleBinary(const std::vector<uint32_t>& binary,
   return result;
 }
 
+// Coverts a string to a vector of uint32_t by coping the content of a given
```
dneto0: "Converts" ... "copying". .. "returns it".

libshaderc_util/src/compiler.cc
dneto0: Pulling it out has the advantage that you could unit test this function.
In fact that's a strong reason to do so.
You should cover cases of string lengths 0 through at least 9.

libshaderc_util/src/compiler.cc
dneto0: I think it should work if, where you have the definition of spirv, you make a tuple variable.  Then  populate its components in the code, then return that variable here.

dneto0:
-1 for:
- some typos
- should unit test the convert-string-to-vector function (and hence expose it to the libshaderc_util api)
- make a tuple variable *early* to avoid the copy on return from the compilation function.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -530,7 +530,7 @@ TEST_F(CppInterface, PreprocessingOnlyModeSecondOverridesDisassemblyMode) {
               HasSubstr("void main(){ }"));
 }
 
-// A shader kind test cases needs: 1) A shader text with or without #pragma
```
Qining: Found a typo in previous CL.

libshaderc_util/src/compiler.cc
```diff
@@ -199,25 +213,31 @@ bool Compiler::Compile(
   success &= PrintFilteredErrors(error_tag, error_stream, warnings_as_errors_,
                                  suppress_warnings_, program.getInfoLog(),
                                  total_warnings, total_errors);
-  if (!success) return false;
+  if (!success) return result_tuple;
 
-  std::vector<uint32_t> spirv;
+  // 'spirv' is an alias for the compilation_output_data. This alias is added to
+  // serve as an input for the call to DissassemblyBinary.
+  std::vector<uint32_t>& spirv = compilation_output_data;
```
dneto0: This is ok, but you could have just used compilation_output_data in the call to DisassembleBinary instead of making the new variable "spirv".

libshaderc_util/src/compiler.cc
```diff
@@ -199,25 +213,31 @@ bool Compiler::Compile(
...
+    compilation_output_data_size_in_bytes = text_or_error.size();
+    return result_tuple;
   } else {
-    return shaderc_util::WriteFile(
-        output_stream,
-        string_piece(
-            reinterpret_cast<const char*>(spirv.data()),
-            reinterpret_cast<const char*>(&spirv.back()) + sizeof(uint32_t)));
+    succeeded = true;
+    // Note compilation_output_data is already set in GlslangToSpv().
```
dneto0: Good comment.

libshaderc_util/src/compiler_test.cc
```diff
@@ -221,4 +221,40 @@ TEST_F(CompilerTest, AddMacroDefinitionNotNullTerminated) {
   EXPECT_TRUE(SimpleCompilationSucceeds(kMinimalExpandedShader, EShLangVertex));
 }
 
+// A convert-string-to-vector test case consists of 1) an input string; 2) an
+// expected vector after the conversion.
+struct ConvertStringToVectorTestCase {
+  const std::string str_;
```
dneto0: The style guide says that struct member names don't end with underscore.
Only class member names do.

dneto0:
-1
Almost there!  Just remove the _ from the struct member names in the test code.

Qining:
Changed 'str_' to 'input_str', 'vec_' to 'expected_output_vec'.

Also changed the word 'set' to 'populate' in the comments.

dneto0:
+2
Removed "WIP" from commit message, rebased, and pushed into master as 455c7cd71

PR #136, opened 2016-02-17
Allow spirv-tools and glslang to live elsewhere.

AWoloszyn:
This allows both spirv-tools and glslang to live in separate source
directories when being invoked with ndk-build.

ndk-build THIRD_PARTY_PATH=foo  specifies that foo/glslang and
foo/spirv-tools exist.

ndk-build SPVTOOLS_LOCAL_PATH=/some/absolute/path specifies that glslang
exists in third_party/glslang but spirv-tools exists at
/some/absolute/path.

ndk-build GLSLANG_LOCAL_PATH=/some/absolute/path is the same but for
glslang.

### DIFF
--- a/third_party/Android.mk
+++ b/third_party/Android.mk
@@ -1,7 +1,7 @@
-THIRD_PARTY_PATH ?= $(call my-dir)
+THIRD_PARTY_PATH := $(call my-dir)
 
-LOCAL_PATH := $(THIRD_PARTY_PATH)/glslang
 GLSLANG_LOCAL_PATH := $(THIRD_PARTY_PATH)/glslang
+LOCAL_PATH := $(GLSLANG_LOCAL_PATH)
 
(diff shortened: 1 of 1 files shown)


dneto0:
+2

third_party/Android.mk
```diff
@@ -47,7 +47,7 @@ LOCAL_MODULE:=glslang
 LOCAL_CXXFLAGS:=-std=c++11 -fno-exceptions -fno-rtti $(GLSLANG_OS_FLAGS)
 LOCAL_EXPORT_C_INCLUDES:=$(GLSLANG_LOCAL_PATH)
 define gen_glslang_tab
-$(call generate-file-dir,$(TO)/glslang_tab.cpp)
```
dekimir: What's this?  I don't have that in a fresh pull of master...

PR #138, opened 2016-02-17
Add target build-version.

dekimir:


dneto0:
This will always overwrite the file.  Any other target that depends on the resulting file will be rebuilt every time we remake the project. That's suboptimal.

Maybe you were going to work around this at a higher level?

-1 for now.

utils/create_build_version.py
```diff
@@ -0,0 +1,47 @@
...
+import subprocess
+import sys
+
+def describe(dir):
+    """Runs 'git describe' in dir.  If successful, returns the output; otherwise,
+returns 'unknown: git-describe error'."""
+    try:
+        return subprocess.check_output(["git", "describe"], cwd=dir).rstrip()
+    except subprocess.CalledProcessError:
+        return 'unknown: git-describe error'
```
AWoloszyn: If git-describe fails, it would be weird to see:
glslc --version
unknown: git-describe error.

That moves the fact that there was a git error into the version string, which is very confusing.
What I think would make more sense is something like
"unknown git hash: build date: " + datetime.datetime.now().time().isoformat().

dekimir:
Addressed feedback and changed the file name to `build-version.inc`.

AWoloszyn:
+1 from me
Will let @dneto0  double check.

CMakeLists.txt
```diff
@@ -17,6 +17,12 @@ add_custom_target(add-copyright
   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   COMMENT "Add copyright")
 
+add_custom_target(build-version
```
AWoloszyn: Just a reminder that you will have to make any target that uses build-version depend on this target.

dekimir: Yes, but that's after I implement `--version`.  This here is enough to stamp binary builds manually when uploading them to a server, for instance.

utils/update_build_version.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python
+
+# Copyright 2016 The Shaderc Authors. All rights reserved.
```
dneto0: This line causes the copyright check to fail, and so my default build fails.
You already filed https://github.com/google/shaderc/issues/139 for that issue.

dneto0:
The logic is good, and works in my test environment.
But -1 since it causes a plain ninja build to fail the copyright check.

dekimir:
Created #142 to fix the copyright check.

dneto0:
+2.
I'll squash and push into master

dneto0:
Pushed into master as e9752f27

PR #142, opened 2016-02-18
Handle arbitrary year in Copyright.  Fixes #139.

### LINKED ISSUE #139: add_copyright.py --check doesn't handle "Copyright 2016"
It currently only admits "Copyright 2015".  It should allow any year.

dekimir:
Also bump the newly inserted copyright messages to 2016.

### DIFF
--- a/utils/add_copyright.py
+++ b/utils/add_copyright.py
@@ -23,9 +23,11 @@ with status 1 if any such files are found, 0 if none.
 import fileinput
 import fnmatch
 import os
+import re
 import sys
 
-COPYRIGHT = 'Copyright 2015 The Shaderc Authors. All rights reserved.'
+COPYRIGHTRE = re.compile('Copyright \d+ The Shaderc Authors. All rights reserved.')
+COPYRIGHT = 'Copyright 2016 The Shaderc Authors. All rights reserved.'
 LICENSED = """
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


dekimir:
Also handle empty files and files with nothing but empty lines or shebangs.

utils/add_copyright.py
```diff
@@ -88,13 +90,14 @@ def insert_copyright(glob, comment_prefix):
...
         for line in fileinput.input(file, inplace=1):
-            if line == copyright:
-                has_copyright = True
+            has_copyright = has_copyright or COPYRIGHTRE.search(line)
             if not has_copyright and not skip(line):
                 sys.stdout.write(copyright)
                 sys.stdout.write(licensed)
                 has_copyright = True
             sys.stdout.write(line)
+        if not has_copyright:
```
dneto0: I guess this covers the case of an initially empty file?
Ok by me.

utils/add_copyright.py
```diff
@@ -23,9 +23,11 @@
 import fileinput
 import fnmatch
 import os
+import re
 import sys
 
-COPYRIGHT = 'Copyright 2015 The Shaderc Authors. All rights reserved.'
+COPYRIGHTRE = re.compile('Copyright \d+ The Shaderc Authors. All rights reserved.')
+COPYRIGHT = 'Copyright 2016 The Shaderc Authors. All rights reserved.'
```
dneto0: So if we modify the same file in several years, we add a new copyright line?
This disallows things like
Copyright 2015-2016 ...

dneto0:
+2

dneto0:
Rebased and pushed into master as  79ccda13ff7

utils/add_copyright.py
dekimir: Good point.  I don't know exactly how we handle that situation, but once we spell out the rules, it will be trivial to modify this RE.

PR #143, opened 2016-02-18
Use statically typed compilation results in C++ API

dneto0:
Class shaderc::CompilationResult is a template over the element type for
the compilation output. It has three specializations: for a SPIR-V
binary module (with uint32_t element type), SPIR-V assembly text (char
element type), and preprocessed source (char element type).

TODO: Should probably also change the C API to make different entry points
for different output types.

### DIFF
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -180,16 +180,15 @@ void FileCompiler::AddIncludeDirectory(const std::string& path) {
 }
 
 void FileCompiler::SetIndividualCompilationFlag() {
-  if (!disassemble_) {
+  if (output_type_ != OutputType::SpirvAssemblyText) {
     needs_linking_ = false;
     file_extension_ = ".spv";
   }
 }
 
 void FileCompiler::SetDisassemblyFlag() {
-  if (!preprocess_only_) {
+  if (!PreprocessingOnly()) {
     output_type_ = OutputType::SpirvAssemblyText;
-    disassemble_ = true;
     needs_linking_ = false;
     file_extension_ = ".spvasm";
   }
@@ -197,7 +196,6 @@ void FileCompiler::SetDisassemblyFlag() {
 
 void FileCompiler::SetPreprocessingOnlyFlag() {
   output_type_ = OutputType::PreprocessedText;
-  preprocess_only_ = true;
   needs_linking_ = false;
   if (output_file_name_.empty()) {
     output_file_name_ = "-";
@@ -220,8 +218,8 @@ bool FileCompiler::ValidateOptions(size_t num_files) {
   // If we are outputting many object files, we cannot specify -o. Also
   // if we are preprocessing multiple files they must be to stdout.
   if (num_files > 1 &&
-      ((!preprocess_only_ && !needs_linking_ && !output_file_name_.empty()) ||
-       (preprocess_only_ && output_file_name_ != "-"))) {
+      ((!PreprocessingOnly() && !needs_linking_ && !output_file_name_.empty()) ||
+       (PreprocessingOnly() && output_file_name_ != "-"))) {
     std::cerr << "glslc: error: cannot specify -o when generating multiple"
                  " output files"
               << std::endl;
@@ -258,12 +256,12 @@ std::string FileCompiler::GetOutputFileName(std::string input_filename) {
 
 std::string FileCompiler::GetCandidateOutputFileName(
     std::string input_filename) {
-  if (!output_file_name_.empty() && !preprocess_only_) {
+  if (!output_file_name_.empty() && !PreprocessingOnly()) {
     return output_file_name_.str();
   }
 
   std::string extension = file_extension_;
-  if (preprocess_only_ || needs_linking_) {
+  if (PreprocessingOnly() || needs_linking_) {
     extension = ".spv";
   }
 
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -64,12 +64,41 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
       new FileIncluder(&include_file_finder_));
   // Get a reference to the dependency trace before we pass the ownership to
   // shaderc::CompileOptions.
-  const auto& included_files = includer->file_path_trace();
+  const auto& used_source_files = includer->file_path_trace();
   options_.SetIncluder(std::move(includer));
 
-  shaderc::CompilationResult result = compiler_.CompileGlslToSpv(
-      source_string.data(), source_string.size(), shader_stage,
-      error_file_name.data(), options_);
+  switch (output_type_) {
+    case OutputType::SpirvBinary: {
+      const auto result = compiler_.CompileGlslToSpv(
+          source_string.data(), source_string.size(), shader_stage,
+          error_file_name.data(), options_);
+      return EmitCompiledResult(result, input_file, error_file_name,
+                                used_source_files, output_stream);
+    }
+    case OutputType::SpirvAssemblyText: {
+      const auto result = compiler_.CompileGlslToSpvAssembly(
+          source_string.data(), source_string.size(), shader_stage,
+          error_file_name.data(), options_);
+      return EmitCompiledResult(result, input_file, error_file_name,
+                                used_source_files, output_stream);
+    }
+    case OutputType::PreprocessedText: {
+      const auto result = compiler_.PreprocessGlsl(
+          source_string.data(), source_string.size(), shader_stage,
+          error_file_name.data(), options_);
+      return EmitCompiledResult(result, input_file, error_file_name,
+                                used_source_files, output_stream);
+    }
+  }
+  return false;
+}
+
+template <typename CompilationResultType>
+bool FileCompiler::EmitCompiledResult(
+    const CompilationResultType& result, const std::string& input_file,
+    string_piece error_file_name,
+    const std::unordered_set<std::string>& used_source_files,
+    std::ostream* out) {
   total_errors_ += result.GetNumErrors();
   total_warnings_ += result.GetNumWarnings();
 
@@ -104,8 +133,9 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
   // later, if the handler is instantiated to dump as normal compilation output,
   // and the original compilation output should be blocked. Otherwise it won't
   // be touched. The main output stream dumps this string_piece later.
-  string_piece compilation_output(result.GetData(),
-                                  result.GetData() + result.GetLength());
+  string_piece compilation_output(
+      reinterpret_cast<const char*>(result.cbegin()),
+      reinterpret_cast<const char*>(result.cend()));
 
   // If we have dependency info dumping handler instantiated, we should dump
   // dependency info first. This may redirect the compilation output
@@ -114,7 +144,7 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
   if (dependency_info_dumping_handler_) {
     if (!dependency_info_dumping_handler_->DumpDependencyInfo(
             GetCandidateOutputFileName(input_file), error_file_name.data(),
-            &potential_dependency_info_output, included_files)) {
+            &potential_dependency_info_output, used_source_files)) {
       return false;
     }
     if (!potential_dependency_info_output.empty()) {
@@ -127,12 +157,12 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
   }
 
   // Write compilation output to output file.
-  output_stream->write(compilation_output.data(), compilation_output.size());
+  out->write(compilation_output.data(), compilation_output.size());
   // Write error message to std::cerr.
   std::cerr << result.GetErrorMessage();
-  if (output_stream->fail()) {
+  if (out->fail()) {
     // Something wrong happened on output.
-    if (output_stream == &std::cout) {
+    if (out == &std::cout) {
       std::cerr << "glslc: error: error writing to standard output"
                 << std::endl;
     } else {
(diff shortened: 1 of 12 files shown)


glslc/src/file_compiler.h
```diff
@@ -101,6 +105,23 @@ class FileCompiler {
   };
 
  private:
+  enum class OutputType {
+    SpirvBinary,  // A binary module, as defined by the SPIR-V specification.
+    SpirvAssemblyText,  // Assembly syntax defined by the SPIRV-Tools project.
+    PreprocessedText,   // Preprocessed source code.
+  };
+
+  // Compiles the given string a the specified shader kind and using stored
```
dekimir: This blurb seems incorrect: `EmitCompiledResult()` doesn't compile anything, it merely emits the result of a prior compilation.

glslc/src/file_compiler.h
```diff
@@ -101,6 +105,23 @@ class FileCompiler {
...
+  // options. On success, writes the result to the given output stream and
+  // returns true.  On failure, returns false, and possibly emit messages to
+  // the standard error stream.
+  template <typename CompilationResultType>
+  bool EmitCompiledResult(
+      const CompilationResultType& result, const std::string& input_file_name,
+      shaderc_util::string_piece error_file_name,
+      const std::unordered_set<std::string>& used_source_files,
+      std::ostream* out);
```
dekimir: Can this method be `const`?

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -39,12 +41,12 @@ bool CompilationResultIsSuccess(const shaderc::CompilationResult& result) {
 // magic number in the fixed postion of the byte array in the result object.
 // Returns true if the magic number is found at the correct postion, otherwise
 // returns false.
-bool IsValidSpv(const shaderc::CompilationResult& result) {
+template <typename T>
+bool IsValidSpv(const shaderc::CompilationResult<T>& result) {
```
Qining: I think we can use "SpvCompilationResult" instead of "CompilationResult<T>" here. Otherwise the "uint32_t*" below may look a little bit weird.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -492,22 +495,22 @@ TEST_F(CppInterface, ErrorTagIsInputFileName) {
 
 TEST_F(CppInterface, PreprocessingOnlyOption) {
   options_.SetPreprocessingOnlyMode();
-  shaderc::CompilationResult result = compiler_.CompileGlslToSpv(
+  const shaderc::AssemblyCompilationResult result = compiler_.PreprocessGlsl(
```
Qining: I think the type should be "PreprocessedSourceCompilationResult", not "AssemblyCompilationResult".

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -25,13 +25,15 @@
 
 namespace {
 
+using shaderc::AssemblyCompilationResult;
```
Qining: I guess we can have all three result types here?

libshaderc/include/shaderc/shaderc.hpp
```diff
@@ -97,6 +111,14 @@ class CompilationResult {
...
+// A compilation result for a SPIR-V binary module, which is an array
+// of uint32_t words.
+using SpvCompilationResult = CompilationResult<uint32_t>;
+// A compilation result in SPIR-V assembly syntax.
+using AssemblyCompilationResult = CompilationResult<char>;
+// Preprocessed source text.
+using PreprocessedSourceCompilationResult = CompilationResult<char>;
+
 // Contains any options that can have default values for a compilation.
 class CompileOptions {
```
dekimir: Wouldn't it make sense to also remove `SetPreprocessingOnlyMode()` and `SetDisassemblyMode()` from `CompileOptions`?

libshaderc/include/shaderc/shaderc.hpp
```diff
@@ -55,23 +69,23 @@ class CompilationResult {
...
-  // This pointer remains valid only until the CompilationResult is destroyed.
-  const char* GetData() const {
-    if (!compilation_result_) {
-      return "";
-    }
-    return shaderc_result_get_bytes(compilation_result_);
+  // Returns an iterator pointing to the start of the compilation
+  // output.  It is valid for the lifetime of this object.
+  // If there is no compilation result, then returns nullptr.
+  const_iterator cbegin() const {
```
AWoloszyn: It is completely valid to also call these begin() and end() (and have them return const_iterators).
This way range-based for loops can/will still work correctly.

glslc/src/file_compiler.h
dneto0: Turns out, no since the method also accumulates warning and error counts.  That will be mentioned in the blurb.

glslc/src/file_compiler.h
dneto0: Agreed.  Rewritten.

libshaderc/src/shaderc_cpp_test.cc
dneto0: Agreed.

libshaderc/src/shaderc_cpp_test.cc
dneto0: Good catch.

libshaderc/src/shaderc_cpp_test.cc
dneto0: Done.

libshaderc/include/shaderc/shaderc.hpp
dneto0: Yes. After next iteration in this PR.

libshaderc/include/shaderc/shaderc.hpp
dneto0: Done in next iteration.

libshaderc/include/shaderc/shaderc.hpp
```diff
@@ -55,25 +72,30 @@ class CompilationResult {
...
-  // Returns a pointer to the start of the compiled SPIR-V.
-  // It is guaranteed that static_cast<uint32_t> is valid to call on this
-  // pointer.
-  // This pointer remains valid only until the CompilationResult is destroyed.
-  const char* GetData() const {
-    if (!compilation_result_) {
-      return "";
-    }
-    return shaderc_result_get_bytes(compilation_result_);
+  // Returns an iterator pointing to the start of the compilation
```
AWoloszyn: Can we specify that this is a random_access_iterator.

We should also mention that the iterator returned iterates through contiguous memory, which is an important distinction if we plan on being able to put this in a string_piece.

glslc/src/file_compiler.cc
```diff
@@ -104,8 +133,9 @@ bool FileCompiler::CompileShaderFile(const std::string& input_file,
   // later, if the handler is instantiated to dump as normal compilation output,
   // and the original compilation output should be blocked. Otherwise it won't
   // be touched. The main output stream dumps this string_piece later.
-  string_piece compilation_output(result.GetData(),
-                                  result.GetData() + result.GetLength());
+  string_piece compilation_output(
```
AWoloszyn: Do we really need the reinterpret_casts here?
result::const_iterator is defined as  const OutputElementType*

libshaderc/include/shaderc/shaderc.hpp
dekimir: Mentioned up on line 29.

dneto0: There's a comment at the top of the class describing the kinds of iterators being handed out.  But next CL will expand this comment to say
"Returns a random access (contiguous) iterator ..."

glslc/src/file_compiler.cc
dneto0: Yes we do because OutputElementType can be a uint32_t and you need the reinterpret cast to change the pointer type.

AWoloszyn: Right, I see now, sorry about the confusion.

dneto0:
All feedback addressed.   No longer WIP, aside from needing a rebase.

AWoloszyn:
Looks good to me.

Qining:
lgtm

glslc/src/file_compiler.cc
```diff
@@ -158,6 +188,7 @@ void FileCompiler::SetIndividualCompilationFlag() {
 
 void FileCompiler::SetDisassemblyFlag() {
   if (!preprocess_only_) {
+    output_type_ = OutputType::SpirvAssemblyText;
     disassemble_ = true;
```
dekimir: `disassemble_` and `preprocess_only_` seem redundant now that we have `output_type_`, no?

libshaderc/include/shaderc/shaderc.h
```diff
@@ -251,22 +237,37 @@ typedef struct shaderc_compilation_result* shaderc_compilation_result_t;
...
-// the additional_options, the source string will be compiled into char strings
-// and held by the returned shaderc_compilation_result. The entry_point_name
-// null-terminated string defines the name of the entry point to associate with
-// this GLSL source. If the additional_options parameter is not NULL, then the
-// compilation is modified by any options present. May be safely called from
-// multiple threads without explicit synchronization. If there was failure in
-// allocating the compiler object NULL will be returned.
+// The source string will be compiled into SPIR-V binary and a
+// shaderc_compilation_result will be returned to hold the results.
+// compilation.  The entry_point_name null-terminated string defines the name
```
dekimir: Copy pasta: "compilation."

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -126,12 +138,21 @@ class CppInterface : public testing::Test {
...
         compiler_.CompileGlslToSpv(shader, kind, "shader", options);
     EXPECT_TRUE(CompilationResultIsSuccess(compilation_result)) << kind << '\n';
-    // Use string(const char* s, size_t n) constructor instead of
-    // string(const char* s) to make sure the string has complete binary data.
-    // string(const char* s) assumes a null-terminated C-string, which will cut
-    // the binary data when it sees a '\0' byte.
-    return std::string(compilation_result.GetData(),
-                       compilation_result.GetLength());
+    // Need to make sure you get complete binary data, including embedded nulls.
+    return CompilerOutputAsString(compilation_result);
```
dekimir: I don't understand why this returns SPIR-V binary as a string?

dekimir: In fact, I don't see where this function is used at all (in this file)?

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -491,51 +508,22 @@ TEST_F(CppInterface, ErrorTagIsInputFileName) {
...
   EXPECT_TRUE(CompilationResultIsSuccess(result_from_cloned_options));
-  EXPECT_THAT(result_from_cloned_options.GetData(),
-              HasSubstr("void main(){ }"));
-}
-
-TEST_F(CppInterface, PreprocessingOnlyModeFirstOverridesDisassemblyMode) {
-  // Sets preprocessing only mode first, then sets disassembly mode.
-  // Preprocessing only mode should override disassembly mode.
-  options_.SetPreprocessingOnlyMode();
-  options_.SetDisassemblyMode();
```
dekimir: Optional: I wonder if it's possible to create an equivalent test on `FileCompiler::SetDisassemblyFlag()`?

libshaderc/src/shaderc_test.cc
```diff
@@ -355,29 +396,20 @@ TEST_F(CompileStringWithOptionsTest, ValuelessMacroCompileOptionsNullPointer) {
...
   EXPECT_THAT(disassembly_text, HasSubstr("Capability Shader"));
   EXPECT_THAT(disassembly_text, HasSubstr("MemoryModel"));
-
-  // The mode should be carried into any clone of the original option object.
-  compile_options_ptr cloned_options(
-      shaderc_compile_options_clone(options_.get()));
-  const std::string disassembly_text_cloned_options = CompilationOutput(
-      kMinimalShader, shaderc_glsl_vertex_shader, cloned_options.get());
-  EXPECT_THAT(disassembly_text_cloned_options, HasSubstr("Capability Shader"));
-  EXPECT_THAT(disassembly_text_cloned_options, HasSubstr("MemoryModel"));
```
dekimir: Why was the above part removed?

dekimir:
-1 for copy-pasta, dead code in `shaderc_cpp_test`, and a removed `disassembly_text_cloned_options`.

libshaderc/src/shaderc_cpp_test.cc
dneto0: There are tests later in the file for the presence of debug info.  Those tests use HasSubstr looking for the name of a variable from the GLSL source.  That's why we return the SPIR-V binary as a string.  We don't want to pattern-match against an OpName instruction since we don't want to overspecify the named ID.

libshaderc/src/shaderc_cpp_test.cc
dneto0: Well yes our course.  So far we've chosen to write tests for glslc as an executable rather than gtest-based tests for the internals of glslc.

libshaderc/src/shaderc_test.cc
dneto0: The old test checked two things: 
- that setting the disassembly option on the options object cause the compiler output to be SPIR-V assembly text.
- that such setting was carried over when that options object was cloned.

However, in the new design disassembling is no longer an attribute on the options object. So it doesn't make sense to check the behaviour on a cloned object.  Cloning the options is orthogonal to calling the shaderc_compile_into_spv_assembly, way up in the Compilation helper class.

glslc/src/file_compiler.cc
dneto0: Removed in next commit.
For preprocess_only_ I made a small helper method PreprocessOnly that checks the value of output_type_ appropriately.

libshaderc/include/shaderc/shaderc.h
dneto0: Fixed in next commit.

dneto0:
All feedback addressed.

dekimir:
+2

dneto0:
Rebased, squashed, then pushed into master as 92d05ce5

PR #144, opened 2016-02-18
Added missing include directory to android build scripts.

AWoloszyn:
When glslang_tab.cpp and glslang_tab.cpp.h moved, the include directory
was not updated.

This was missed because the old files had not been removed from the
source directory.


dneto0:
+2

PR #146, opened 2016-02-19
Minor enhancements involving finder and includer.

antiagainst:


libshaderc_util/src/file_finder.cc
```diff
@@ -21,11 +21,8 @@
 namespace {
 
 // Returns "" if path is empty or ends in '/'.  Otherwise, returns "/".
-std::string MaybeSlash(const std::string path) {
-  if (path.empty() || path.back() == '/')
-    return "";
-  else
-    return "/";
+std::string MaybeSlash(const std::string& path) {
+  return (path.empty() || path.back() == '/') ? "" : "/";
```
dneto0: The ? :  is arguably less readable.

But the reference on the argument is useful.

dneto0:
+1

dekimir:
Rebased and pushed via 1f84b05e8efa02e018844997153a2478ed8e322c.

PR #149, opened 2016-02-22
Add an example program for simple C++ API usage

dneto0:
The program demonstrates basic shader compilation using the Shaderc
C++ API.  For clarity, each method is deliberately self-contained.

Techniques demonstrated:
- Preprocessing GLSL source text
- Compiling a shader to SPIR-V assembly text
- Compliing a shader to a SPIR-V binary module
- Setting basic options: setting a preprocessor symbol.
- Checking compilation status and extracting an error message.

### DIFF
--- a/examples/online-compile/CMakeLists.txt
+++ b/examples/online-compile/CMakeLists.txt
@@ -1,0 +1,4 @@
+add_executable(shaderc-online-compile main.cc)
+shaderc_default_compile_options(shaderc-online-compile)
+target_include_directories(shaderc-online-compile PRIVATE ${libshaderc_SOURCE_DIR}/include)
+target_link_libraries(shaderc-online-compile PRIVATE shaderc)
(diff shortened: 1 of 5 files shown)


AWoloszyn:
Looks nice and clear to me.
+1 Will let someone else comment as well.

examples/online-compile/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_executable(shaderc-online-compile main.cc)
+shaderc_default_compile_options(shaderc-online-compile)
+target_include_directories(shaderc-online-compile PRIVATE ${libshaderc_SOURCE_DIR}/include)
```
Qining: Just a minor thought, I guess we don't need to add libshaderc's include directory explicitly. The include directory should be carried with the library.

Qining:
lgtm. Very clear to me.

examples/online-compile/CMakeLists.txt
dneto0: You're right.  Fixed in the next commit.

dneto0:
Retained +2.  Squashed and pushed into master as  66d409cc6

PR #150, opened 2016-02-23
Export the includes in the android.mk file

AWoloszyn:
When we generate libshaderc_combined in Android.mk, also
create a libs/ and include/ directory that contains the actual
libraries.

This means a library can be created containing the libraries and
include files with the following style command.

ndk-build APP_PROJECT_PATH=<path/to/shaderc>
APP_BUILD_SCRIPT=<path/to/shaderc/Android.mk> APP_STL=<stl_version>
APP_ABI=<abis_to_build> NDK_APP_LIBS_OUT=<target_dir>

This will generate a target_dir with the following format
<target_dir> /
  include /
    shaderc /
      shaderc.h
      shaderc.hpp
  libs /
    <abis>
      libshaderc.a

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -26,7 +26,33 @@ $(1)/libshaderc_combined.a: $(addprefix $(1)/, $(ALL_LIBS)) $(1)/combine.ar
 	@cd $(1) && $(2)ar -M < combine.ar && cd -
 	@$(2)objcopy --strip-debug $(1)/libshaderc_combined.a
 
-libshaderc_combined:$(1)/libshaderc_combined.a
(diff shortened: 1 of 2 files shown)


dneto0:
The code looks fine but the commit message looks off.

- In the code the APP_STL is a directory above the target API directory, but is unmentioned in the output tree in the commit message.
-  NDK_APP_LIBS_OUT doesn't line up between the example in the commit message and the code.

Should the commit message say:

```
    This will generate a target_dir with the following format
    <target_dir>/ .. /
      include /
        shaderc /
          shaderc.h
          shaderc.hpp
     <target_dir> /
            <stl> /
               <abis> /
                   libshaderc.a
```

-1 just to fix up the commit message

AWoloszyn:
Updated the commit message.

dneto0:
+2

PR #153, opened 2016-02-24
Use raw string literal to store help messages.

antiagainst:


dekimir:
Rebased and pushed into master via 1ccede1528cb33533ff196b44a1ccbcd5dd79a90.

PR #154, opened 2016-02-24
Fixed CMake to use PYTHON_EXECUTABLE everywhere.

AWoloszyn:
It looks like some verisons of cmake set PYTHON_EXE but others
do not. According to the documentation PYTHON_EXECUTABLE
is the correct way to do it.


antiagainst:
LGTM

BTW, don't forget to change the instance in spirv-tools. ;)

AWoloszyn:
Thanks for the reminder, I will put up a CL there momentarily.

PR #155, opened 2016-02-24
If no tag in the repo, then show the commit hash

dneto0:
Also, use subprocess.Popen and p.communicate to get
better portability for Windows.

### DIFF
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -30,13 +30,37 @@ import sys
 
 OUTFILE = 'build-version.inc'
 
+
+def command_output(cmd, dir):
+    """Runs a command in a directory and returns its standard output stream.
+
+    Captures the standard error stream.
+
+    Raises a RuntimeError if the command fails to launch or otherwise fails.
+    """
+    p = subprocess.Popen(cmd,
+                         cwd=dir,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    (stdout, _) = p.communicate()
+    if p.returncode !=0:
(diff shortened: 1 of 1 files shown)


utils/update_build_version.py
```diff
@@ -30,13 +30,32 @@
 
 OUTFILE = 'build-version.inc'
 
+
+def command_output(cmd, dir):
+    """Runs a command in a directory and returns its standard output stream.
+
+    Raises a RuntimeError if the command fails to launch or otherwise fails.
+    """
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, cwd=dir)
```
AWoloszyn: To avoid spamming stderr if git-describe fails, you could stderr=subprocess.PIPE as well.

dneto0:
All feedback addressed.

AWoloszyn:
+2

utils/update_build_version.py
```diff
@@ -30,13 +30,37 @@
...
+
+    Raises a RuntimeError if the command fails to launch or otherwise fails.
+    """
+    p = subprocess.Popen(cmd,
+                         cwd=dir,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE)
+    (stdout, _) = p.communicate()
+    if p.returncode !=0:
+        raise RuntimeError("Failed to run %s in %s" % (cmd, dir))
```
dekimir: Optional: might be useful to include stderr content in the message.

dneto0:
Rebased and pushed into master as 8f7357ce

PR #156, opened 2016-02-24
Fixed the windows build.

AWoloszyn:
On MSVC gtest copies the struct passed into a parameterized test.
It does this via assignment, so we have to make it assignable.


dneto0:
+2

PR #159, opened 2016-02-26
Link executables statically on Windows.

dekimir:


CMakeLists.txt
```diff
@@ -18,6 +18,16 @@ add_custom_target(add-copyright
   WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
   COMMENT "Add copyright")
 
+if(WIN32)
```
dneto0: I think this should be conditional on MSVC since it's Microsoft compiler flags that are modified.
Someone might want to compile using MinGW on Windows, and hence see GCC flags.

E.g. see the    if (MSVC) conditional in cmake/utils.cmak

dneto0: Also, this segment merits a comment.  Say why you're doing this.

dneto0:
-1

dekimir:
Feedback addressed.

dneto0:
+2

PR #162, opened 2016-03-02
Added an option to use the shared CRT.

AWoloszyn:
This can be set by an external project or also in the configuration
itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,14 +19,19 @@ add_custom_target(add-copyright
   COMMENT "Add copyright")
 
 if(MSVC)
-  # Link executables statically by replacing /MD with /MT everywhere.
-  foreach(flag_var
-      CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-      CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-    if(${flag_var} MATCHES "/MD")
-      string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-    endif(${flag_var} MATCHES "/MD")
-  endforeach(flag_var)
+  option(SHADERC_ENABLE_SHARED_CRT
+          "Use the shared CRT instead of the static CRT"
+          ${SHADERC_ENABLE_SHARED_CRT})
+  if (NOT SHADERC_ENABLE_SHARED_CRT)
+    # Link executables statically by replacing /MD with /MT everywhere.
+    foreach(flag_var
+        CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
+        CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
+      if(${flag_var} MATCHES "/MD")
+        string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
+      endif(${flag_var} MATCHES "/MD")
+    endforeach(flag_var)
+  endif(SHADERC_ENABLE_SHARED_CRT)
 endif(MSVC)
 
 # Configure subdirectories.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -19,14 +19,18 @@ add_custom_target(add-copyright
...
 if(MSVC)
-  # Link executables statically by replacing /MD with /MT everywhere.
-  foreach(flag_var
-      CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-      CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-    if(${flag_var} MATCHES "/MD")
-      string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-    endif(${flag_var} MATCHES "/MD")
-  endforeach(flag_var)
+  option(ENABLE_SHARED_CRT "Use the shared CRT instead of the static CRT"
```
dneto0: Suggest we use a SHADERC_ prefix so someone could control this even when Shaderc is embedded in a larger project.

dneto0:
-1   Nit: Suggest renaming the option.
Also, should probably update the README to describe the linking behaviour, at least for Windows.

AWoloszyn:
Addressed all comments.

README.md
```diff
@@ -93,6 +93,10 @@ After a successful build, you should have a `glslc` executable somewhere under
 the `$BUILD_DIR/glslc/` directory, as well as a `libshaderc` library somewhere
 under the `$BUILD_DIR/libshaderc/` directory.
 
+The default behavior on MSVC is to link with the static CRT. If you would like
```
dekimir: Consider something like "...MSVC is to statically link all libraries and executables, obviating the need for Visual Studio redistributables on the target system".

CMakeLists.txt
```diff
@@ -19,14 +19,19 @@ add_custom_target(add-copyright
...
-  foreach(flag_var
-      CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-      CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO)
-    if(${flag_var} MATCHES "/MD")
-      string(REGEX REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-    endif(${flag_var} MATCHES "/MD")
-  endforeach(flag_var)
+  option(SHADERC_ENABLE_SHARED_CRT
+          "Use the shared CRT instead of the static CRT"
+          ${ENABLE_SHARED_CRT})
```
dekimir: Does this mean that the command line flag is `-DENABLE_SHARED_CRT` (not what the README says)?  If not, how can this be leveraged?

README.md
dneto0: I'd say "avoiding" instead of "obviating".  Simpler language is better.

CMakeLists.txt
dneto0: I think it's just a typo.  Should be SHADERC_ENABLED_SHARED_CRT in both places?

AWoloszyn: This was a failing to change all places.

README.md
AWoloszyn: I think that that statement is not sufficient. What really matters to the user is that they link the static CRT since this is something that has to match between a user's program and libshaderc. The statement "we statically link all libraries and executables" does not necessarily convey this information in a clear way. This is especially true since the flag is -DSHADERC_ENABLE_SHARED_CRT.

dekimir: I just wasn't sure that "static CRT" is synonymous with "CRT statically linked into an executable/library/DLL".  But it [sounds](https://msdn.microsoft.com/en-us/library/abx4dbyh(v=vs.120).aspx) like it is.

AWoloszyn: That is correct. Specifically it refers to which library your executable/library/dll will statically link to, the one that is actually the static library or the one that exports the symbols for the shared CRT. I find the nomenclature of "static CRT" to be much more clear.

dekimir:
LGTM, but make sure all checks pass.  Especially AppVeyor.

nkyriazis:
SHADERC_ENABLE_SHARED_CRT is exposed as a cmake option. However, changing it does not affect build. Its effect is established during the first cmake configure/generate. Afterwards, changing it has no  effect.

PR #164, opened 2016-03-06
Improve Python3 Compatibility

antiagainst:
* Use print_function from __future__ to tackle the print syntax
  change.
* In Python3 str and bytes are no longer the same thing. Need
  proper decode.

Counterpart for https://github.com/KhronosGroup/SPIRV-Tools/issues/144

### DIFF
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -73,6 +73,11 @@ def main():
 
     projects = ['shaderc', 'spirv-tools', 'glslang']
     tags = [
+        # decode() is needed here for Python3 compatibility. In Python2,
+        # str and bytes are the same type, but not in Python3.
+        # Popen.communicate() returns a bytes instance, which needs to be
+        # decoded into text data first in Python3. And this decode() won't
+        # hurt Python2.
         describe(p).decode('ascii').replace('"', '\\"')
         for p in sys.argv[1:]]
     new_content = ''.join([
--- a/utils/update_build_version.py
+++ b/utils/update_build_version.py
@@ -23,6 +23,8 @@
 # directory's "git describe" output enclosed in double quotes and appropriately
 # escaped.
 
+from __future__ import print_function
+
 import datetime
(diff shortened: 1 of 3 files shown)


antiagainst:
Now it should build cleanly with Python3. But sadly, when testing, `glslc_test_framework.py` and related things are still not Python3 compatible. Will be the next thing to fix.

utils/update_build_version.py
```diff
@@ -55,21 +57,26 @@ def describe(dir):
...
-        print 'usage: {0} <shaderc_dir> <spirv-tools_dir> <glslang_dir>'.format(sys.argv[0])
+        print(
+            'usage: {0} <shaderc_dir> <spirv-tools_dir> <glslang_dir>'.format(sys.argv[0]))
         sys.exit(1)
 
-    new_content = ('"shaderc ' + describe(sys.argv[1]).replace('"', '\\"') + '\\n"\n' +
-                   '"spirv-tools ' + describe(sys.argv[2]).replace('"', '\\"') + '\\n"\n' +
-                   '"glslang ' + describe(sys.argv[3]).replace('"', '\\"') + '\\n"\n')
+    tags = [
+        describe(p).decode('ascii').replace('"', '\\"')
```
dekimir: Why is `decode('ascii')` necessary here?

antiagainst: In Python2, `str` is `bytes`, but not in Python3.

https://docs.python.org/3/howto/pyporting.html#text-versus-binary-data

Here, `Popen.communicate()` returns `bytes`, which in Python3 is binary data, cannot be directly `+`'d (or invoking `str`-related methods) to `str`, which is text data.

dekimir: I think there should be a comment explaining this, as it's non-obvious.

antiagainst: Done.

dekimir:
+2

PR #165, opened 2016-03-08
Add a platform-independent build script

dekimir:
Take over from #161.  Implement that PR's feedback.  Make the Windows build self-contained so it works from any environment.

### DIFF
--- a/utils/build.py
+++ b/utils/build.py
@@ -15,43 +15,36 @@
 # limitations under the License.
 
 """
-Builds the Shaderc project, on Linux, Darwin, or Windows.
-
-For windows, we assume we have bash and basic shell commands, and we'll
-use the MSVC compiler.
-
-For Linux and Mac, use the default C++ compiler.
-
-We require cmake and ninja commands to be present under the prebuilts
-directory in the specified root directory.
+Builds the Shaderc project, on Linux, Mac, or Windows.
 """
 
+from __future__ import print_function
 import argparse
 import os
-import os.path
+import platform
 import subprocess
 import sys
 
 
-def run(args, cmd, cwd):
+def run(cmd, cwd, justprint):
     """Prints a command to run, and optionally runs it.
 
     Raises a RuntimeError if the command does not launch or otherwise fails.
 
     Args:
-      args: Options.  If args.dry_run is true, then only print the command.
-            Otherwise run the command after printing it.
-      cmd:  List of words in the command.
-      cwd:  Working directory for the command.
+      justprint: If true, then only print the command. Otherwise run the 
+                   command after printing it.
+      cmd:       List of words in the command.
+      cwd:       Working directory for the command.
     """
-    print cmd
-    if args.dry_run:
+    print(cmd)
+    if justprint:
         return
 
     p = subprocess.Popen(cmd, cwd=cwd)
     (_, _) = p.communicate()
-    if p.returncode !=0:
-        raise RuntimeError("Failed to run %s in %s" % (cmd, dir))
+    if p.returncode != 0:
+        raise RuntimeError("Failed to run %s in %s" % (cmd, cwd))
 
 
 def build(args):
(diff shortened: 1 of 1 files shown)


utils/build.py
```diff
@@ -0,0 +1,135 @@
...
+    cmake_command = [cmake, args.srcdir, '-GNinja',
+                     '-DCMAKE_BUILD_TYPE=%s' % args.buildtype,
+                     '-DCMAKE_INSTALL_PREFIX=%s' % (args.installdir)]
+
+    if (platform.system() == 'Windows'):
+        bat_content = '''
+call "%VS140COMNTOOLS%..\..\VC\\vcvarsall.bat"
+{cmake_command}
+{ninja} install
+        '''.strip().format(cmake_command=' '.join(cmake_command), ninja=ninja)
```
AWoloszyn: Is there any plan to have this run any of the unit tests during build?

utils/build.py
```diff
@@ -0,0 +1,135 @@
...
+    print('   Ninja      : ', ninja)
+    print('   Build dir  : ', args.builddir)
+    print('   Install dir: ', args.installdir)
+    cmake_command = [cmake, args.srcdir, '-GNinja',
+                     '-DCMAKE_BUILD_TYPE=%s' % args.buildtype,
+                     '-DCMAKE_INSTALL_PREFIX=%s' % (args.installdir)]
+
+    if (platform.system() == 'Windows'):
+        bat_content = '''
+call "%VS140COMNTOOLS%..\..\VC\\vcvarsall.bat"
```
dneto0: Not a big deal, but why the double-backslash just before vcvarsall?

Do we have to provide an argument to vcvarsal?
https://msdn.microsoft.com/en-us/library/f2ccy3wt.aspx
The arg would either be x86 or amd64.

But I defer to any testing you've done already.

utils/build.py
dneto0: Yes, we should test during the build.  On a first iteration I'd be ok letting it slide, as we will manually test the first round of results.

dneto0:
-1 for a reply whether the arg to vcvarsall is required.

utils/build.py
dekimir: Double `\` before `v` is necessary, since otherwise `build.bat` has a vertical tab in it.

The vcvarsal argument defaults to x86, which produces executables runnable on all Windows platforms.  @AWoloszyn and I discussed this and decided we don't need to get fancy with the argument.

dekimir:
Addressed feedback so far.

antiagainst:
It would be nice to run `pyformat` and also test out with Python3 before merging.

utils/build.py
```diff
@@ -0,0 +1,173 @@
...
+    print('   Ninja      : ', ninja)
+    print('   Build dir  : ', args.builddir)
+    print('   Install dir: ', args.installdir)
+    cmake_command = [cmake, args.srcdir, '-GNinja',
+                     '-DCMAKE_BUILD_TYPE=%s' % args.buildtype,
+                     '-DCMAKE_INSTALL_PREFIX=%s' % (args.installdir)]
+
+    if (OS == 'Windows' or OS.startswith('CYGWIN')):
+        bat_content = '''
+call "%VS140COMNTOOLS%..\..\VC\\vcvarsall.bat"
```
dneto0: Please verify that running vcvarsall without arguments gives you a compiler that will target 32-bit x86.

dekimir: https://msdn.microsoft.com/en-us/library/f2ccy3wt.aspx#Anchor_1

> If no arguments are provided, vcvarsall.bat configures the environment variables for using the 32-bit native compiler for x86 targets.

dekimir:
Addressed feedback so far, PTAL.

dneto0:
+2

dekimir:
Squashed and merged into master as 7926051.

PR #166, opened 2016-03-08
DO NOT MERGE: Update the interface to handle the new Glslang includer.

AWoloszyn:
This will allow subsequent changes to handle relative includes.

Will update this PR once the associated change has gone into glslang.
For review for now.


glslc/src/file_includer.cc
```diff
@@ -14,35 +14,50 @@
...
+  if (full_path.empty())
+    return MakeErrorIncludeResult("Cannot find or open include file.");
+
+  // In principle, several threads could be resolving includes at the same
+  // time.  Protect the included_files.
+
+  // Read the file and save its full path and contents into stable addresses.
+  FileInfo* new_file_info = new FileInfo{full_path, {}};
+  if (!shaderc_util::ReadFile(full_path, &(new_file_info->contents))) {
+    return MakeErrorIncludeResult("Couldn't read file");
```
antiagainst: "Cannot read file" to keep consistency with the above error message.

libshaderc_util/src/counting_includer_test.cc
```diff
@@ -19,13 +19,24 @@
...
 class ConcreteCountingIncluder : public shaderc_util::CountingIncluder {
  public:
-  std::pair<std::string, std::string> include_delegate(
-      const char* filename) const override {
-    return std::make_pair<std::string, std::string>("", "Unexpected #include");
+  virtual glslang::TShader::Includer::IncludeResult* include_delegate(
+      const char* requested, glslang::TShader::Includer::IncludeType,
+      const char* requestor,
+      size_t) override {
+    const char kError[] = "Unexpecged #include";
```
antiagainst: `Unexpecged` typo. :)

glslc/src/file_includer.cc
```diff
@@ -14,35 +14,50 @@
...
+shaderc_include_result* MakeErrorIncludeResult(const char* message) {
+  return new shaderc_include_result{"", 0, message, strlen(message)};
+}
+
+shaderc_include_result* FileIncluder::GetInclude(const char* requested_source,
+                                                 shaderc_include_type,
+                                                 const char*, size_t) {
+  // TODO(dneto): Be sensitive to requesting_source
+  // We don't actually care about the include type.
+  std::string full_path = file_finder_.FindReadableFilepath(requested_source);
```
antiagainst: `const std::string full_path`.

glslc/src/file_includer.h
```diff
@@ -34,22 +36,36 @@ class FileIncluder : public shaderc::CompileOptions::IncluderInterface {
...
-  std::unordered_set<std::string> source_files_used_;
+
+  // The full path and content of a source file.
+  struct FileInfo {
+    const std::string full_path;
+    std::vector<char> contents;
+  };
+
+  // The set of full paths of included files.
+  std::unordered_set<std::string> included_files_;
```
antiagainst: So the order of the included files doesn't matter? I don't quite understand this part. And I don't find this `included_files_` actually get used (expect `insert`ing into it)?

AWoloszyn: So this is the same as what was before just renamed from source_files_used_ to included_files.

It is used in file_path_trace. It just returns all of the files that were included. It is an unordered_set because the only thing we actually care about is the list of all files included from a base file. This then becomes the unique set of all files reachable from a base.

AWoloszyn:
Responded to comments.

libshaderc_util/src/counting_includer_test.cc
```diff
@@ -19,13 +19,24 @@
 
 namespace {
 
+auto kRelative = glslang::TShader::Includer::EIncludeRelative;
```
antiagainst: `const auto ...`

glslc/src/file_includer.h
antiagainst: Ok.

antiagainst:
+1.

AWoloszyn:
Done

libshaderc/src/shaderc.cc
```diff
@@ -159,42 +160,81 @@ class StageDeducer {
...
+    // shaderc_include_result object, so we can later release the latter.
+    return new glslang::TShader::Includer::IncludeResult{
+        std::string(include_result->source_name,
+                    include_result->source_name_length),
+        include_result->content, include_result->content_length,
+        include_result};
+  }
+
+  // Releases the given IncludeResult.
+  virtual void release_delegate(
```
Qining: Do we need a destructor to make sure we release the IncludeResult, since we are not calling 'release' in 'include_delegate' anymore?

glslc/src/file_includer.h
```diff
@@ -34,22 +36,36 @@ class FileIncluder : public shaderc::CompileOptions::IncluderInterface {
...
+  // until it's released.
+  shaderc_include_result* GetInclude(const char* requested_source,
+                                     shaderc_include_type type,
+                                     const char* requesting_source,
+                                     size_t include_depth) override;
+  // Releases an include result.
+  void ReleaseInclude(shaderc_include_result* include_result) override;
+
+  // Returns a reference to the member storing the set of included files.
   const std::unordered_set<std::string>& file_path_trace() const {
```
dneto0: For another CL.  This methods should be called "included_files" 
It's a cheap accessor to the value of a member variable of the same name.

glslc/src/file_includer.h
```diff
@@ -34,22 +36,36 @@ class FileIncluder : public shaderc::CompileOptions::IncluderInterface {
...
+  // The full path and content of a source file.
+  struct FileInfo {
+    const std::string full_path;
+    std::vector<char> contents;
+  };
+
+  // The set of full paths of included files.
+  std::unordered_set<std::string> included_files_;
+  // Guards access to the included_files_.
+  std::mutex cache_mutex_;
```
dneto0: For consistency and safety shouldn't we mutex-guard the *read* of the member variable as well, via file_path_trace?
(And in that case since it's not just a trivial accessor, the read method name should be GetIncludedFiles since it does non-trivial processing.)

libshaderc/src/shaderc.cc
```diff
@@ -159,42 +160,81 @@ class StageDeducer {
...
+                       const shaderc_include_result_release_fn result_releaser,
+                       void* user_data)
+      : resolver_(resolver),
+        result_releaser_(result_releaser),
         user_data_(user_data){};
   InternalFileIncluder()
-      : get_includer_response_(nullptr),
-        release_includer_response_(nullptr),
-        user_data_(nullptr){};
+      : resolver_(nullptr), result_releaser_(nullptr), user_data_(nullptr){};
```
dneto0: I'm suprised there's no space before the {
But I'll defer to clan-format here.

libshaderc/src/shaderc.cc
dneto0: I think it's ok this way because the release is now the responsibility of the client making the inclusion requests.

AWoloszyn:
Squashed and responded to comments. Explicitly specified that the file_includer provides the basic thread-safety guarantee. 
I.e. in order to maintain sane include results, you should only keep one of these per compilation.

libshaderc/src/shaderc.cc
Qining: But do we need to take care of the new-ed pointer of glslang::TShader::Includer::IncludeResult? If release_delegate() is missed, will the space created in line:219 be destroyed?

dneto0: That's on line 233.

dneto0:
+2

PR #167, opened 2016-03-09
Added simple C smoke test to prevent future C breakages.

AWoloszyn:
This C file just compiles a single shader, but is sufficient
to make sure that shaderc.h can always be included from C.


dneto0:
+2

PR #168, opened 2016-03-11
Fix corner case and return str in describe() as said in the blurb.

antiagainst:
When the given directory is not inside a git managed project,
`describe()` returns a `str` instance instead of `bytes` instance
in Python3, which will case problem when calling `decode()` on it.

Similar to https://github.com/KhronosGroup/SPIRV-Tools/pull/147


dekimir:
+2

antiagainst:
Just a typo fix in the commit message.

PR #170, opened 2016-03-14
Make build.py work when ninja isn't in the path.

dekimir:


dekimir:
@AWoloszyn @antiagainst @Qining @dneto0 PTAL

PR #171, opened 2016-03-16
Modify build.py not to run .bat scripts.

dekimir:


dekimir:
@dneto0 @AWoloszyn @Qining @antiagainst PTAL

AWoloszyn:
Looks good to me, tested on VS2013 and works well. (Had to change %VS140%...)

My only comment is maybe we should allow a user to optionally specify a different compiler for windows, that way this can work on past/future versions of VS.

Not required for this checkin, but something to think of for the future.

AWoloszyn:
+2

PR #173, opened 2016-03-17
Fix environment building on Cygwin.

dekimir:


dekimir:
@AWoloszyn @antiagainst @dneto0 @Qining PTAL

utils/build.py
```diff
@@ -104,12 +99,14 @@ def build(args):
                      '-DCMAKE_MAKE_PROGRAM=%s' % ninja]
 
     env = None
-    if (OS == 'Windows' or OS.startswith('CYGWIN')):
-        p = subprocess.Popen(
+    if (OS == 'Windows'):
```
antiagainst: redundant `()` in `if`-statement.

utils/build.py
```diff
@@ -104,12 +99,14 @@ def build(args):
...
+        env = parse_output_into_env(
             '"%VS140COMNTOOLS%..\\..\\VC\\vcvarsall.bat" & set',
-            stdout=subprocess.PIPE, cwd=args.builddir, shell=True)
-        env = dict([tuple(line.split('='))
-                    for line in p.communicate()[0].splitlines()])
+            args.builddir)
+    if (OS.startswith('CYGWIN')):
```
antiagainst: This should be `elif`?

utils/build.py
```diff
@@ -49,18 +49,13 @@ def run(cmd, cwd, env, justprint):
...
-    indices list.
+def parse_output_into_env(command, cwd):
+    """Runs command in cwd, expecting it to produce name=value lines on
+    output.  Parses those lines into a dictionary object.
     """
-    quoted = [(quote(e) if i in indices else e)
-              for (i, e) in enumerate(command)]
-    return ' '.join(quoted)
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, cwd=cwd, shell=True)
+    return dict([tuple(line.split('='))
```
dneto0: split is wrong here.  
For example, 
  'FOO=BAR=BAZ'.split('=)
will make a list of 3 elements.  You should only split on the first equal sign.

dneto0:
-1 for the split bug.

dekimir:
Feedback addressed.

antiagainst:
+1

dekimir:
Replaced by #176.

PR #174, opened 2016-03-17
Update README for MinGW compile

dneto0:
A bad interaction between CMake, MinGW, and Googletest prevents
MinGW from compiling Googletest with threads.
https://github.com/google/googletest/issues/606

So we must configure Googletest without threading support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,18 @@ cmake --build . --config {Release|Debug|MinSizeRel|RelWithDebInfo}
 ctest -C {Release|Debug|MinSizeRel|RelWithDebInfo}
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,6 +89,18 @@ cmake --build . --config {Release|Debug|MinSizeRel|RelWithDebInfo}
 ctest -C {Release|Debug|MinSizeRel|RelWithDebInfo}
 ```
 
+4c) Or build with MinGW on Linux for Windows:
+(Skip building threaded unit tests due to Googletest bug
```
antiagainst: `[Google Test bug](...)`

dekimir:
+2

PR #178, opened 2016-03-18
Use os.environ instead of os.putenv in build.py.

dekimir:
It seems that os.putenv somehow doesn't work on Unix.

Also use os.pathsep.join() instead of "+ os.pathsep +".

Also run pyformat.


dekimir:
@dneto0 @AWoloszyn @antiagainst @Qining PTAL.

dekimir:
Verified it works on Windows and Cygwin.

dneto0:
+2

PR #179, opened 2016-03-21
Removed the remaining threaded tests.

AWoloszyn:
Also fixed cmake target if mingw is present.


AWoloszyn:
@dneto0 @dekimir @Qining @antiagainst PTAL

PR #180, opened 2016-03-22
Fix directory-existence check in build.py.

dekimir:
Apparently, `os.path.isdir()` returns false on Windows when the directory is a symlink.

### DIFF
--- a/utils/build.py
+++ b/utils/build.py
@@ -65,7 +65,7 @@ def build(args):
 
     # Make paths absolute, and ensure directories exist.
     for d in [args.builddir, args.installdir]:
-        if not os.path.isdir(d):
+        if not os.path.exists(d):
             os.makedirs(d)
     args.srcdir = os.path.abspath(args.srcdir)
     args.builddir = os.path.abspath(args.builddir)


dekimir:
@dneto0 @AWoloszyn @Qining @antiagainst PTAL

antiagainst:
Interesting. From the manual, it should handle symlinks. And it returns `True` on LInux...

antiagainst:
+2

AWoloszyn:
-1 I do not think this is going to fix your problem.

Th issue you are running into, as far as I can see it, is that you have a symlink (in cygwin) to your out-directory. Unfortunately Windows itself can not see these symlinks. What this ACTUALLY is, is a file called out. Windows applications will not be able to see it as a directory, it is only a valid directory in cygwin.

dekimir:
@AWoloszyn: that doesn't invalidate this change, it merely means another one will be necessary.  Regardless of what the Cygwin symlink is, it's better not to attempt a `mkdir` over it.

AWoloszyn:
Good point! +2

dneto0:
I can go either way on this patch.  Merge it if you wish.

dekimir:
I decided this isn't helpful: lack of a normal directory becomes a problem sooner or later, so we may as well fail early.  Also, #181 takes care of Cygwin symlinks during escaping to native Windows.

PR #181, opened 2016-03-22
Run cygpath when restarting build.py in Windows.

dekimir:


dekimir:
@AWoloszyn @antiagainst @dneto0 @Qining PTAL

utils/build.py
```diff
@@ -91,6 +91,10 @@ def build(args):
     run(['ctest', '--output-on-failure'], args.builddir, env, args.dry_run)
 
 
+def cygpath(f):
+    p = subprocess.Popen(['cygpath', '-w', f], stdout=subprocess.PIPE)
```
AWoloszyn: Can we use check_output here instead the output will be more clear. (I can't remember if this was failing on windows, if so, ignore my comment.)

AWoloszyn:
Other than my one comment, looks good to me.

utils/build.py
dekimir: No: see #155.

AWoloszyn:
Right looks good to me.

antiagainst:
Already committed as 85207c70fcb83995e5f362dbdaf4ee7b7af1e968.

PR #182, opened 2016-03-22
Allowed third_party projects to be located elsewhere.

AWoloszyn:
This allows us to specify the location of third party projects
if we would rather have them somewhere other than third_party/


antiagainst:
LGTM

PR #185, opened 2016-03-30
Enabled relative includes in glslc.

AWoloszyn:
glslc will now by default try and resolve #included files relative to
the current file first, only after that fails will it try to include
files based on the -I parameters.

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -210,7 +210,10 @@ empty value.
 
 `-I` adds the specified directory to the search path for include files.
 
-WARNING: Implementation is not complete yet.
+Only `""` style includes are permitted. When searching for included files, first
+the directory of the current file is attempted. If this fails each directory
+supplied via `-I` on the command line is searched in the given order for the
+file.
 
 === Code Generation Options
 
@@ -386,3 +389,11 @@ output.
 
 The `GL_GOOGLE_cpp_style_line_directive` extension is implicitly turned on by
 the `GL_GOOGLE_include_directive` extension.
+
+It is invalid to have any '\' or '/' characters in any filename provided to
+an `#include` directive. The presence of any such characters may lead to
+unexpected behavior on different systems.
+
+Furthermore it is not possible to escape any characters in an `#include`
+directive and as such, any special characters that would have to be escaped in
+C may not show up in file names either.
--- a/glslc/src/file_includer.cc
+++ b/glslc/src/file_includer.cc
@@ -25,13 +25,16 @@ shaderc_include_result* MakeErrorIncludeResult(const char* message) {
   return new shaderc_include_result{"", 0, message, strlen(message)};
 }
 
-shaderc_include_result* FileIncluder::GetInclude(const char* requested_source,
-                                                 shaderc_include_type,
-                                                 const char*, size_t) {
-  // TODO(dneto): Be sensitive to requesting_source
-  // We don't actually care about the include type.
+shaderc_include_result* FileIncluder::GetInclude(
+    const char* requested_source, shaderc_include_type include_type,
+    const char* requesting_source, size_t) {
+
   const std::string full_path =
-      file_finder_.FindReadableFilepath(requested_source);
+      (include_type == shaderc_include_type_relative)
+          ? file_finder_.FindRelativeReadableFilepath(requesting_source,
+                                                      requested_source)
+          : file_finder_.FindReadableFilepath(requested_source);
+
   if (full_path.empty())
     return MakeErrorIncludeResult("Cannot find or open include file.");
 
--- a/glslc/test/include.py
+++ b/glslc/test/include.py
@@ -14,6 +14,7 @@
 
 import expect
 from glslc_test_framework import inside_glslc_testsuite
+from placeholder import SpecializedString
 from environment import File, Directory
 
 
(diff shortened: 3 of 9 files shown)


AWoloszyn:
@dekimir @dneto0 @antiagainst @Qining PTAL.

glslc/src/file_includer.cc
```diff
@@ -25,13 +25,18 @@ shaderc_include_result* MakeErrorIncludeResult(const char* message) {
   return new shaderc_include_result{"", 0, message, strlen(message)};
 }
 
-shaderc_include_result* FileIncluder::GetInclude(const char* requested_source,
-                                                 shaderc_include_type,
-                                                 const char*, size_t) {
+shaderc_include_result* FileIncluder::GetInclude(
+    const char* requested_source, shaderc_include_type include_type,
+    const char* requesting_source, size_t) {
   // TODO(dneto): Be sensitive to requesting_source
```
dneto0: Remove the TODO now. :-)

glslc/README.asciidoc
```diff
@@ -210,7 +210,9 @@ empty value.
 
 `-I` adds the specified directory to the search path for include files.
 
-WARNING: Implementation is not complete yet.
+Only `""` style includes are permitted. When searching for included files, first
```
antiagainst: Duplicate `first`.

glslc/README.asciidoc
```diff
@@ -210,7 +210,9 @@ empty value.
...
+the directory of the current file is attempted first, if this fails each
+-I supplied on the command line is searched in turn for the given file.
```
antiagainst: directories supplied via `-I` ... searched in the given order ...

antiagainst:
Just some minor issues.
+1.

I wonder what if I include myself? Will that create a crash? :)

glslc/README.asciidoc
```diff
@@ -210,7 +210,9 @@ empty value.
 
 `-I` adds the specified directory to the search path for include files.
 
-WARNING: Implementation is not complete yet.
+Only `""` style includes are permitted. When searching for included files, first
+the directory of the current file is attempted first, if this fails each
```
dekimir: Replace comma with a period or semi-colon.

glslc/test/include.py
```diff
@@ -306,3 +306,203 @@ class TestWrongPoundVersionInIncludedFile(expect.ErrorMessage):
...
+
+
+@inside_glslc_testsuite('Include')
+class VerifyRelativeIncludeWithDashI(expect.StdoutMatch):
+    """Tests #including a relative file from a -I directory."""
+
+    environment = Directory('.', [
+        File('a.vert', '#version 140\ncontent a\n#include "b.glsl"\n'),
+        Directory('foo', [
+            File('b.glsl', '#include "c.glsl"\ncontent b\n'),
```
dekimir: Consider putting `b.glsl` into a subdir of `foo` -- that way you know it's resolving `c.glsl` via `-I`.

glslc/test/include.py
```diff
@@ -306,3 +306,203 @@ class TestWrongPoundVersionInIncludedFile(expect.ErrorMessage):
...
+ content c
+#line 1 "foo/b.glsl"
+ content b
+#line 3 "a.vert"
+
+"""
+
+
+@inside_glslc_testsuite('Include')
+class VerifyRelativeNeighbourDirectory(expect.StdoutMatch):
```
dekimir: Nit: we usually use American spelling in code.

libshaderc_util/include/libshaderc_util/file_finder.h
```diff
@@ -33,14 +33,17 @@ class FileFinder {
...
-  // absolute filenames will work as expected.  Note that a "." entry on the
-  // search path may be prepended to an absolute filename (eg, "/a/b/c") to
-  // create a relative result (eg, ".//a/b/c").
   std::string FindReadableFilepath(const std::string& filename) const;
 
+  // Searches for a read-openable file based on filename, which must be
+  // non-empty. The search is first attempted as a path relative to
+  // the requesting_file parameter. If no file is found relative to the
+  // requesting_file then this acts as FindReadableFilepath does. If
+  // requesting_file does not contain a '/' character then it is assumed to be
```
dekimir: What about backslash?

libshaderc_util/src/file_finder.cc
```diff
@@ -42,4 +42,24 @@ std::string FileFinder::FindReadableFilepath(
   return "";
 }
 
+std::string FileFinder::FindRelativeReadableFilepath(
+    const std::string& requesting_file, const std::string& filename) const {
+  assert(!filename.empty());
+
+  std::string relative_path =
```
dekimir: This name is confusing.  I'd use `dirname` instead.

libshaderc_util/src/file_finder.cc
```diff
@@ -42,4 +42,24 @@ std::string FileFinder::FindReadableFilepath(
...
+      requesting_file.substr(0, requesting_file.find_last_of("/\\"));
```
dekimir: Would it be better to use `string_piece` here?

glslc/test/include.py
```diff
@@ -306,3 +306,203 @@ class TestWrongPoundVersionInIncludedFile(expect.ErrorMessage):
...
+            File('b.glsl', '#include "c.glsl"\ncontent b\n')
+        ]),
+    ])
+
+    glslc_args = ['-E', 'a.vert']
+
+    expected_error = [
+        "foo/b.glsl:1: error: '#include' : Cannot find or open include file.\n",
+        '1 error generated.\n'
+    ]
```
dekimir: Is something like `#include "./././b.glsl"` expected to work?

dekimir: Are there tests for backslashes in the filename?  What about multiple consecutive slashes?

dekimir: What happens if there are multiple possible relative resolutions?  I guess the same question may be asked for multiple `-I` dirs containing the same file...

dekimir: Not for this PR, but we need more `-I` tests.  We don't seem to have any on master at the moment.

glslc/test/include.py
AWoloszyn: This was intended to test the "b.glsl" include, but I have made the change, as it makes the test more robust.

glslc/test/include.py
AWoloszyn: Is it possible to have multiple relative resolutions?

glslc/test/include.py
dneto0: I honour the licence taken here to assert small cultural micro-aggressions.  ;-)

glslc/test/include.py
dneto0: I'd like to see a test of a relative include where the original filename is provided as an absolute path.
And another where the -I argument is provided with an absolute path.

Note that in our test framework, the TestCase has a 'directory' member telling you where the test case is being run.

dneto0:
-1
I'll be mostly happy once @dekimir is happy.  Except I also want some testing for absolute paths.

glslc/test/include.py
AWoloszyn: I am going to specify that backslashes in the filename are banned, (not the path but the name) because this will lead to surprising behavior across platforms. For multiple successive slashes, it will do what the OS itself does in those cases.

AWoloszyn:
@antiagainst  As for including yourself, the same thing that would happen in C, which I am guessing.... infinite memory usage followed by crash. 

#including yourself is not actually invalid, and there can be some valid ways to do it

#ifdef FOO
do_stuff
#else 
#define FOO
#include "self"
#endif

AWoloszyn:
I think this addresses all comments.
@dekimir  @dneto0 @antiagainst

libshaderc_util/include/libshaderc_util/file_finder.h
```diff
@@ -39,8 +39,8 @@ class FileFinder {
   // non-empty. The search is first attempted as a path relative to
   // the requesting_file parameter. If no file is found relative to the
   // requesting_file then this acts as FindReadableFilepath does. If
-  // requesting_file does not contain a '/' character then it is assumed to be
-  // a filename and the request will be relative to the current directory.
+  // requesting_file does not contain a '/' or a '/'character then it is assumed
```
dneto0: Second one should be \

glslc/test/placeholder.py
```diff
@@ -115,3 +116,19 @@ def instantiate_for_glslc_args(self, testcase):
 
     def instantiate_for_expectation(self, testcase):
         return os.path.join(testcase.directory, self.filename)
+
+
+class SpecializedString(PlaceHolder):
+    """Returns a string that has been specialized based on TestCase."""
+
+    def __init__(self, filename):
```
dneto0: The constructor argument might not always be a filename.  We need a more generic name for the member.  Something like "template".

glslc/test/placeholder.py
```diff
@@ -115,3 +116,19 @@ def instantiate_for_glslc_args(self, testcase):
 
     def instantiate_for_expectation(self, testcase):
         return os.path.join(testcase.directory, self.filename)
+
+
+class SpecializedString(PlaceHolder):
+    """Returns a string that has been specialized based on TestCase."""
```
dneto0: This docstring could use a little more detail.  What does specialization mean?  It means template instantiation where $Var is replaced by the Var member of the Testcase.

glslc/README.asciidoc
```diff
@@ -211,8 +211,8 @@ empty value.
 `-I` adds the specified directory to the search path for include files.
 
 Only `""` style includes are permitted. When searching for included files, first
-the directory of the current file is attempted first, if this fails each
--I supplied on the command line is searched in turn for the given file.
+the directory of the current file is attempted. If this fails each directory
+supplied via -I on the command line is searched in the given order for the file.
```
antiagainst: `-I` should be quoted.

glslc/README.asciidoc
```diff
@@ -388,3 +388,11 @@ output.
 
 The `GL_GOOGLE_cpp_style_line_directive` extension is implicitly turned on by
 the `GL_GOOGLE_include_directive` extension.
+
+It is invalid to have any '\' or '/' characters in any filename provided to
+an #include directive. The presence of any such characters may lead to
```
antiagainst: `#include` should be quoted.

glslc/test/placeholder.py
AWoloszyn: Unfortunately:
glslc/test/glslc_test_framework.py
requires that this parameter is "filename".

I plan on adding a follow-up CL in the future that changes this.

glslc/test/include.py
```diff
@@ -506,3 +509,53 @@ class VerifyRelativeOnlyToSelf(expect.ErrorMessage):
         "foo/b.glsl:1: error: '#include' : Cannot find or open include file.\n",
         '1 error generated.\n'
     ]
+
+
+@inside_glslc_testsuite('Include')
+class VerifyRelativeFromAbsolutePath(expect.StdoutMatch):
+    """Tests that a relative includes are only relative to yourself."""
```
dneto0: remove "a"

glslc/test/include.py
```diff
@@ -506,3 +509,53 @@ class VerifyRelativeOnlyToSelf(expect.ErrorMessage):
...
+#line 0 "$directory/b.glsl"
+ content b
+#line 3 "$directory/a.vert"
+
+""")
+
+
+@inside_glslc_testsuite('Include')
+class VerifyDashIAbsolutePath(expect.StdoutMatch):
+    """Tests that a relative includes are only relative to yourself."""
```
dneto0: Remove "a"

dneto0:
-1
From me, just comment fixes.

AWoloszyn:
Responded to all feedback.

glslc/test/placeholder.py
```diff
@@ -115,3 +116,24 @@ def instantiate_for_glslc_args(self, testcase):
 
     def instantiate_for_expectation(self, testcase):
         return os.path.join(testcase.directory, self.filename)
+
+
+class SpecializedString(PlaceHolder):
+    """Returns a string that has been specialized based on TestCase.
+
+    The string is specialized by expanding is as a string.Template
```
dneto0: "expanding it as a"...

AWoloszyn:
Done!

dneto0:
+2

PR #187, opened 2016-04-01
libspirv.h in SPIRV-Tools removed macros for version and revision.

antiagainst:
Counterpart for https://github.com/KhronosGroup/SPIRV-Tools/pull/170

### DIFF
--- a/third_party/Android.mk
+++ b/third_party/Android.mk
@@ -123,11 +123,12 @@ LOCAL_SRC_FILES:= \
 		source/operand.cpp \
 		source/print.cpp \
 		source/spirv_endian.cpp \
+		source/spirv_target_env.cpp \
 		source/table.cpp \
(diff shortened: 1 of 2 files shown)


third_party/Android.mk
```diff
@@ -107,11 +107,12 @@ LOCAL_SRC_FILES:= \
 		source/operand.cpp \
 		source/print.cpp \
 		source/spirv_endian.cpp \
+		source/spirv_target_env.cpp \
```
dekimir: This file should go into a separate commit with a different title.

dekimir: Clarification: by "this file", I mean `third_party/Android.mk`.

antiagainst: Done.

dekimir:
+2 but pls don't squash

PR #188, opened 2016-04-01
Add rules to build grammar tables in Android.mk for SPIRV-Tools.

antiagainst:


dneto0:
+1

third_party/Android.mk
```diff
@@ -85,13 +85,29 @@ include $(BUILD_STATIC_LIBRARY)
...
+
+define gen_spvtools_grammar_tables
+$(call generate-file-dir,$(SPVTOOLS_OUT_PATH)/opcode.inc)
+$(1)/opcode.inc $(1)/operand.inc: $(SPVTOOLS_LOCAL_PATH)/utils/generate_grammar_tables.py \
+                                  $(SPVTOOLS_LOCAL_PATH)/source/grammar.json
+		@$(HOST_PYTHON) $(SPVTOOLS_LOCAL_PATH)/utils/generate_grammar_tables.py \
+		                $(SPVTOOLS_LOCAL_PATH)/source/grammar.json \
+		                --opcode-file=$(1)/opcode.inc \
+		                --operand-file=$(1)/operand.inc
+		@echo "[$(TARGET_ARCH_ABI)] Grammar: opcode.inc & operand.inc <= grammar.json"
```
AWoloszyn: Don't need quotes here (and it doesn't fit with the rest of the output.

third_party/Android.mk
```diff
@@ -85,13 +85,29 @@ include $(BUILD_STATIC_LIBRARY)
...
+$(SPVTOOLS_LOCAL_PATH)/source/opcode.cpp: $(1)/opcode.inc
+$(SPVTOOLS_LOCAL_PATH)/source/operand.cpp: $(1)/operand.inc
```
AWoloszyn: I'm willing to let this go, but it does create a fairly strange build dependency, because now for the arm64 build depends on the x86 opcode.inc file.

The only thing I can thing that would solve this is as part of your build do something like copy $(SPVTOOLS_LOCAL_PATH)/source/opcode.cpp -> $(1)/opcode.cpp and change your build to use that opcode.cpp instead of $(SPVTOOLS_LOCAL_PATH)/source/opcode.cpp. 

Then arm64 would only depend on arm64 stuff. However that is very complicated, and I don't think will actually make any difference in practice.

third_party/Android.mk
antiagainst: I'm afraid I cannot remove that because of the `<=` substring. Aligned them in 81247ec.

dneto0: Might need the quotes since the "<" character is special to the shell.

third_party/Android.mk
antiagainst: Wow! That's something I was not aware of, but makes sense for me. Compilation works fine though.

AWoloszyn:
+2

third_party/Android.mk
dneto0: How about adding a fake-ish target in between:

```
$(SPVTOOLS_LOCAL_PATH)/source/opcode.cpp : $(1)/opcode.inc.$(APP_ABI)
$(1)/opcode.inc.$(APP_ABI) : $(1)/opcode.inc
     touch $@
```

PR #190, opened 2016-04-04
Reflect ExtInstSet generation steps of SPIRV-Tools in Android.mk.

antiagainst:
Counterpart for https://github.com/KhronosGroup/SPIRV-Tools/pull/174.

### DIFF
--- a/third_party/Android.mk
+++ b/third_party/Android.mk
@@ -88,16 +88,24 @@ LOCAL_PATH := $(SPVTOOLS_LOCAL_PATH)
 SPVTOOLS_OUT_PATH=$(abspath $(TARGET_OUT))
 
 define gen_spvtools_grammar_tables
-$(call generate-file-dir,$(SPVTOOLS_OUT_PATH)/opcode.inc)
-$(1)/opcode.inc $(1)/operand.inc: $(SPVTOOLS_LOCAL_PATH)/utils/generate_grammar_tables.py \
-                                  $(SPVTOOLS_LOCAL_PATH)/source/grammar.json
+$(call generate-file-dir,$(1)/core.insts.inc)
+$(1)/core.insts.inc $(1)/operand.kinds.inc $(1)/glsl.std.450.insts.inc $(1)/opencl.std.insts.inc: \
+        $(SPVTOOLS_LOCAL_PATH)/utils/generate_grammar_tables.py \
+        $(SPVTOOLS_LOCAL_PATH)/source/spirv.core.grammar.json \
+        $(SPVTOOLS_LOCAL_PATH)/source/extinst.glsl.std.450.grammar.json \
+        $(SPVTOOLS_LOCAL_PATH)/source/extinst.opencl.std.grammar.json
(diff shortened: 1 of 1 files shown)


third_party/Android.mk
```diff
@@ -89,15 +89,23 @@ SPVTOOLS_OUT_PATH=$(abspath $(TARGET_OUT))
...
-		                --opcode-file=$(1)/opcode.inc \
-		                --operand-file=$(1)/operand.inc
-		@echo "[$(TARGET_ARCH_ABI)] Grammar        : opcode.inc & operand.inc <= grammar.json"
-$(SPVTOOLS_LOCAL_PATH)/source/opcode.cpp: $(1)/opcode.inc
-$(SPVTOOLS_LOCAL_PATH)/source/operand.cpp: $(1)/operand.inc
+		                --spirv-core-grammar=$(SPVTOOLS_LOCAL_PATH)/source/spirv.core.grammar.json \
+		                --extinst-glsl-grammar=$(SPVTOOLS_LOCAL_PATH)/source/extinst.glsl.std.450.grammar.json \
+		                --extinst-opencl-grammar=$(SPVTOOLS_LOCAL_PATH)/source/extinst.opencl.std.grammar.json \
+		                --core-insts-output=$(1)/core.insts.inc \
+		                --glsl-insts-output=$(1)/glsl.std.450.insts.inc \
```
AWoloszyn: Can this script be run with a selection of arguments?
i.e. what happens if I were to omit --spirv-core-grammar="".

I ask because this would look a lot better as a set of 3(or 4) rules, 1 per target, rather than one large complicated one. It could also be factored into a single function so you could generate a new rule something like:

$(eval $(call gen_spvtools_grammar_tables, $(SPVTOOLS_OUT_PATH), core-insts-output, core.insts.inc)
$(eval $(call gen_spvtools_grammar_tables, $(SPVTOOLS_OUT_PATH),  glsl-insts-output, glsl.std.450.insts.inc)

antiagainst: I'm not sure that's the correct way to go because it means to duplicate the `Capability` enum in the `glsl.std.450` & `opencl.std` grammar, too. That causes synchronization problem. Besides, `glsl.std.450` & 'opencl.std` ExtInstSet *depend* on the core grammar naturally, so I'd prefer that consumers of these two ExtInstSets should refer to the core grammar.

AWoloszyn: That's fair, I don't have a strong opinion, it just seems a bit unfortunate to have a single makefile rule that generates 4 different outputs. (And it's fairly conceivable to see it grow in the future)

AWoloszyn:
+2 from me.

third_party/Android.mk
antiagainst: ;) Sometimes (I mean, here) I feel strong pulse to insert tabs/spaces.

dneto0:
+2 (redundant), depending on https://github.com/KhronosGroup/SPIRV-Tools/pull/174 going in

PR #191, opened 2016-04-05
Improve the #include behaviour description.

dekimir:

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -388,9 +388,10 @@ the `GL_GOOGLE_include_directive` extension.
 The file argument to `#include` must be enclosed in double quotes.  It must be a
 relative path, using whatever path separator the OS supports.  However, the last
 path element -- the name of the file itself -- must not contain either '/' or
-'\', regardless of which path separator is used.  For example, the following is
-invalid: `#include "dir\subdir\subsubdir\filen/ame.vert"`.  This will not be
-flagged as an error but will instead trigger undefined behavior.
+'\', regardless of which path separator is used.  This will not be flagged as an
+error but will instead trigger undefined behavior.  For example, let's say there
+is a file named `f\ilename.vert` on a Unix OS.  It is not possible to craft a
+`#include` that includes that file.
 
 Furthermore, it is not possible to escape any characters in a `#include`
 directive, so the file argument cannot contain any special characters that need
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -208,12 +208,9 @@ empty value.
 
 ==== `-I`
 
-`-I` adds the specified directory to the search path for include files.
-
-Only `""` style includes are permitted. When searching for included files, first
-the directory of the current file is attempted. If this fails each directory
-supplied via `-I` on the command line is searched in the given order for the
-file.
+`-Idirectory' or `-I directory` adds the specified directory to the search path
+for include files.  The directory may be an absolute path or a relative path to
+the current working directory.
 
 === Code Generation Options
 
@@ -381,8 +378,6 @@ The glslc compiler extends GLSL with the include syntax by turning on the
 
 WARNING: This section is still evolving. Expect changes.
 
-TODO: Expected behaviors of the `#include` directive.
-
 If `#include` directives are used in a shader, there will be an `#extension
 GL_GOOGLE_include_directive : enable` line generated into the preprocessed
 output.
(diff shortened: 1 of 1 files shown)


glslc/README.asciidoc
```diff
@@ -208,12 +208,9 @@ empty value.
...
 
-`-I` adds the specified directory to the search path for include files.
-
-Only `""` style includes are permitted. When searching for included files, first
-the directory of the current file is attempted. If this fails each directory
-supplied via `-I` on the command line is searched in the given order for the
-file.
+`-Idirectory` adds the specified directory to the search path for include files.
+The directory may be an absolute path or a relative path to the location where
+`glslc` is run.  It must use path separators that the OS supports.
```
dneto0: the bit about "relative path to the location where glslc is run"  sounds like it could be mis-read as the path to glslc.  Could we be more clear by saying "the current working directory"?

Secondly, "It must use path separators that the OS supports" seems... redundant?

dneto0: We've lost the orderedness property of the search.

glslc/README.asciidoc
```diff
@@ -390,10 +387,17 @@ output.
 The `GL_GOOGLE_cpp_style_line_directive` extension is implicitly turned on by
 the `GL_GOOGLE_include_directive` extension.
 
-It is invalid to have any '\' or '/' characters in any filename provided to
-an `#include` directive. The presence of any such characters may lead to
-unexpected behavior on different systems.
+The file argument to `#include` must be enclosed in double quotes.  It must be a
+relative path, using whatever path separator the OS supports.  However, the last
+path element -- the name of the file itself -- must not contain either '/' or
+'\', regardless of which path separator is used.  For example, the following is
```
dneto0: I think this is a weird way to explain it. 
We should say that for such an example, we'll use 'ame.vert' as a filename in path 'dir\subdir\subsubsdir\filen'.   And whether that path is valid depends on the current platform.

dneto0:
-1

glslc/README.asciidoc
```diff
@@ -390,10 +387,17 @@ output.
 The `GL_GOOGLE_cpp_style_line_directive` extension is implicitly turned on by
```
antiagainst: It seems that the `TODO` in the above (Github won't let me comment on that line) can be deleted now.

glslc/README.asciidoc
```diff
@@ -208,12 +208,9 @@ empty value.
 
 ==== `-I`
 
-`-I` adds the specified directory to the search path for include files.
-
-Only `""` style includes are permitted. When searching for included files, first
-the directory of the current file is attempted. If this fails each directory
-supplied via `-I` on the command line is searched in the given order for the
-file.
+`-Idirectory` adds the specified directory to the search path for include files.
```
antiagainst: By `-Idirectory`, it seems to me that `-I directory` is not valid. I'd suggest just using `-I` here.

glslc/README.asciidoc
dekimir: Hm, I'm of the opposite opinion: to me, it's weird to describe the "ame.vert" scenario, and it's not wise to bake in the current implementation convenience into the spec.  "Undefined behaviour" is simpler and more future-proof.

glslc/README.asciidoc
dekimir: I'd rather spell out both versions than leave it unclear with just `-I`.

dekimir:
Addressed feedback.

glslc/README.asciidoc
antiagainst: Works for me.

glslc/README.asciidoc
```diff
@@ -390,10 +385,18 @@ output.
...
-unexpected behavior on different systems.
-
-Furthermore it is not possible to escape any characters in an `#include`
-directive and as such, any special characters that would have to be escaped in
-C may not show up in file names either.
+The file argument to `#include` must be enclosed in double quotes.  It must be a
+relative path, using whatever path separator the OS supports.  However, the last
+path element -- the name of the file itself -- must not contain either '/' or
+'\', regardless of which path separator is used.  For example, the following is
+invalid: `#include "dir\subdir\subsubdir\filen/ame.vert"`.  This will not be
```
dneto0: This *is* a valid request, it's just that glslc will interpret it as a request for file "ame.vert" in relative path "dir\subdir\subsubdir\filen".  We should say *that* instead of saying it's invalid.

dneto0:
-1 nearly there

glslc/README.asciidoc
dekimir: Reworded to avoid this and state the problem directly.

dneto0:
+2

dekimir:
Squashed and merged to master via afb959a.

PR #192, opened 2016-04-07
Add -I to the glslc synopsis.

dekimir:
@antiagainst @AWoloszyn @dneto0 @Qining PTAL


dekimir:
Hm, on second thought, there are many more flags missing from this synopsis.  It's stale-bait -- maybe we should get rid of it here and direct the reader to run `glslc --help`.

antiagainst:
+2

PR #193, opened 2016-04-07
Enable dependency info dumping tests for abspath

Qining:
Enable tests like:
```
glslc -M /usr/local/a.vert
```
The expected result is:
```
a.vert.spv: /usr/local/a.vert <abspath of included files>
```

### DIFF
--- a/glslc/test/option_dash_M.py
+++ b/glslc/test/option_dash_M.py
@@ -268,14 +268,8 @@ class TestDashCapMDashCapMT(DependencyInfoStdoutMatch):
     dependency_rules_expected = [{'target': 'target',
                                   'dependency': {'shader.vert'}}]
 
-# TODO:(qining) Add tests for absolute input path with #include directives, the
-# dependent file should also be represented in absolute path.
-# e.g. glslc -M /usr/local/a.vert
-#   => a.vert.spv: /usr/local/a.vert /usr/local/b.vert /usr/local/subdir/c.vert
-# These tests needs to be enabled after we have full-fledged includer.
 
-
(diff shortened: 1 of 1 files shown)


glslc/test/option_dash_M.py
```diff
@@ -268,32 +268,37 @@ class TestDashCapMDashCapMT(DependencyInfoStdoutMatch):
...
 class TestDashCapMInputAbsolutePathWithInclude(DependencyInfoStdoutMatch):
     """Tests -M have included files represented in absolute paths when the input
     file is represented in absolute path.
     E.g. Assume a.vert has '#include "b.vert"'
          glslc -M /usr/local/a.vert
       => a.vert.spv: /usr/local/a.vert /usr/local/b.vert
     """
-    shader_a = FileShader('#version 140\n#include "b.vert"\nvoid main(){}\n',
-                          '.vert')
-    shader_b = FileShader('void foo(){}\n', '.vert')
```
Qining: My original code is wrong, as `include "b.vert"` won't point to shader_b.

dekimir:
Note that we don't support absolute paths in `#include` anymore.

Qining:
Yes, and we don't have tests for `include "<abs path>"`. The enabled tests are only for:
"when input file is represented in absolute path, the dependent files should be represented in absolute paths too."

antiagainst:
+1.

As discussed yesterday, it would be better to move all the test framework related stuff to their expected files. But that would be another pull request.

dekimir:
+2

dneto0:
Rebased and pushed into master as a18ea5bee5a

PR #194, opened 2016-04-11
Remove revision from SPV_ENV_*.

dekimir:
Shaderc response to https://github.com/KhronosGroup/SPIRV-Tools/pull/178.


antiagainst:
+2

PR #195, opened 2016-04-12
Test Vulkan-specific shaders

dneto0:
Add unit and integration tests check that Vulkan-specific shaders
compile for the default compiler target, and when we explictly specify
the Vulkan target.

Add unit tests to check that OpenGL-specific shaders fail to compile
when targeting Vulkan.

### DIFF
--- a/glslc/test/option_target_env.py
+++ b/glslc/test/option_target_env.py
@@ -30,6 +30,11 @@ def opengl_vertex_shader():
 void main() { int t = gl_VertexID; }"""
 
 
+def vulkan_vertex_shader():
+    return """#version 310 es
+void main() { int t = gl_VertexIndex; }"""
+
+
 @inside_glslc_testsuite('OptionTargetEnv')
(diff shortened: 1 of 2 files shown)


glslc/test/option_target_env.py
```diff
@@ -66,6 +71,28 @@ class TestTargetEnvEqOpenglWithOpenGlVertexShader(expect.ValidObjectFile):
...
+
+@inside_glslc_testsuite('OptionTargetEnv')
+class TestTargetEnvEqVulkanWithVulkanShader(expect.ValidObjectFile):
+    """Tests that compiling a Vulkan-specific shader succeeds with
+    --target-env=vulkan"""
+    shader = FileShader(vulkan_vertex_shader(), '.vert')
+    glslc_args = ['--target-env=vulkan', '-c', shader]
+
+
+# Note: Negative tests are covered in the libshaderc_util unit tests.
```
dekimir: One could argue that this is stale-bait. I personally don't care too much.

dekimir:
+2

PR #196, opened 2016-04-13
Support Glslang with its new HLSL parser library

dneto0:
This commit requires a matching update to
github.com/google/glslang to catch up to
github.com/KhronosGroup/glslang.

Also remove include directories for SPIRV-Tools that no
longer exist.

### DIFF
--- a/libshaderc_util/CMakeLists.txt
+++ b/libshaderc_util/CMakeLists.txt
@@ -25,7 +25,7 @@ target_include_directories(shaderc_util
   PUBLIC include PRIVATE ${glslang_SOURCE_DIR})
 find_package(Threads)
 target_link_libraries(shaderc_util PRIVATE
-  glslang OSDependent OGLCompiler glslang SPIRV
+  glslang OSDependent OGLCompiler HLSL glslang SPIRV
   SPIRV-Tools ${CMAKE_THREAD_LIBS_INIT})
 
 shaderc_add_tests(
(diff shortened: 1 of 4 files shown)


libshaderc_util/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ target_include_directories(shaderc_util
   PUBLIC include PRIVATE ${glslang_SOURCE_DIR})
 find_package(Threads)
 target_link_libraries(shaderc_util PRIVATE
```
AWoloszyn: Do we remember why we declared this as a PRIVATE library dependency. I think it is fairly clearly a PUBLIC one as anyone who links shader_util must also link to glslang etc etc etc.

IIRC If we make this public we can remove all of the glslang-related includes from glslc.

AWoloszyn:
One small comment/question but does not have to be handled on this commit.
Looks good to me +2

libshaderc_util/CMakeLists.txt
dneto0: Good question.  I think it should be  PUBLIC here.  In fact I think Glslang's CMakeLists.txt should be fixed to export them as public as well.  No?

dneto0: Or rather, instead, if we get it all right.

dneto0:
Also updated google/glslang to current KhronosGroup/master

antiagainst:
It would be nice to be able to configure glslang to turn off HLSL frontend. But that's a separate issue.

PR #198, opened 2016-04-15
Fixed Android.mk for Windows.

AWoloszyn:
Missing , in $(call host-cp) merged arguments on windows into broken
path.

cd - does not work on windows, replace with cd $(ROOT_SHADERC_PATH)


dneto0:
+2

PR #200, opened 2016-04-18
Default glslc to SPIR-V 1.1.

dekimir:


antiagainst:
-1.

I think we also need to update `DisassembleBinary()` in `libshaderc_util/src/compiler.cc`.

dekimir:
Duh, sorry about that.  Fixed now.

But this begs another question: do we have a way to dictate the SPIR-V version to glslang?  And if not, how do we know which one it'll produce?

dekimir:
After some discussion, this is not desirable at this time.  It will become interesting if and when Vulkan defaults to a newer SPIR-V version.  Abandoning...

PR #203, opened 2016-04-18
SPIRV-Tools now requires SPIRV-Headers.

antiagainst:
Counterpart of https://github.com/KhronosGroup/SPIRV-Tools/pull/186.


dekimir:
+2

PR #204, opened 2016-04-18
Compilation fails for null input file name string

dneto0:
When compilation fails, the compilation result length is 0.


dekimir:
+2

dneto0:
Rebased and pushed onto master as c2dcb799e

Issue #214, opened 2016-05-08
Support for compiling to C/C++ header


Themaister:
For many applications, it would be very convenient to be able to compile GLSL directly to a SPIR-V encoded as a static const uint32_t[].

In smaller projects, and especially for embedded/mobile, it is very useful to bake the SPIR-V shaders directly into the binary.

dekimir:
Shaderc follows closely the glslang design here, and glslang [only has functions](https://github.com/KhronosGroup/glslang/blob/master/SPIRV/GlslangToSpv.h#L45) to compile into a `vector<unsigned int>` for now.

Themaister:
True, but emitting a C header is essentially just a different way of dumping the u32 vector to disk, so it shouldn't affect glslang at all. It is standalone enough that this could even be a separate binary.

dekimir:
Sorry, I thought you were proposing a libshaderc interface extension.  But it sounds like you just want to take the `spv_binary` and print its elements into a `.h` file?  That's doable in ~4 lines of code, I think.  Or are you proposing that we add a flag to glslc to do it automatically?

Themaister:
I was thinking about adding some flags to glslc to dump out a header instead of raw .spv.
Again, trivial to do, but it's very convenient.

dekimir:
@Themaister: makes sense.  We'll come up with a suitable flag name and implement.

dneto0:
I think we should mimic GCC's -masm=dialect option.
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#index-masm_003d_0040var_007bdialect_007d-2832

So you'd use -S to emit assembly and -masm=c to say you want C source.

One difficulty is in saying what variable name you want for the result. You might want to do this with several binaries.
So how about we just emit a brace-initialization list of values, and it's up to you to wrap the result.
eg. the output could be
```
  {  0x07230203, 0x00010000, 0x00080000, 1, 0 }
```
And you'd put that into a "binary.inc"
Then in your C code (or C++ code) you'd say
```
static uint32_t empty_binary[] = 
#include "binary.inc"
;
```

The point is that all the decisions about integrating it into the rest of your program are left to you, rather than adding lots of config parameters to glslc.

In this case I'd be tempted to use -masm=c-init-list as the option.

Themaister:
[...]

PR #218, opened 2016-05-15
-mfmt=bin/c/num to emit binary as it is, numbers or C init list

Qining:
Issue: #214

Add command line option: -masm=c-init-list to emit SPIRV binary code as
a C-style uint32_t array initializer list in the output file.

-masm=c-init-list should only be specified when the compilation output
is SPIRV binary code. When used with disassembly mode (-S) and
preprocessing-only mode (-E), it will result in error and fail the
compilation. Note that dumping dependency info as compilation output (-M
or -MM) also implies preprocessing-only (-E) and will lead to error.
But dumpping dependency info to extra .d file (-MD or -MMD) won't cause
error as the compilation output is still SPIRV binary code.

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -252,6 +252,39 @@ code, such as identifier names and line numbers.
 NOTE: Currently this option has no effect.  Full functionality depends on
 glslang support for generating debug info.
 
+==== `-mfmt=<format>`
+
+`-mfmt=<format>` selects output format for compilation output in SPIR-V binary
+code form.  Supported options are listed in the
+<<binary-output-format-options,binary output format options>> table. This
+option is only valid to be used when the compilation output is SPIR-V binary
+code. Specifying any options listed below when the output is not SPIR-V binary
+code, like disassembly (with `-S` specified), text (with `-M`, `-MM` or `-E`
+specified) will trigger an error.
+
+[[binary-output-format-options]]
+.Binary Output Format Options
+[cols="20%,80%"]
+|===
+|Format option  |Description
+
+|bin            |Output SPIR-V binary code as a sequence of binary 32-bitwords
+                 in host native endianness. This is the default format for
+                 SPIR-V binary compilation output.
+|num            |Output SPIR-V binary code as a text file containing a list of
+                 comma-separated hex numbers. +
+                 Example: `glslc -c -mfmt=num main.vert -o output_file.txt` +
+                 Content of the output_file.txt: +
+                 0x07230203,0x00010000,0x00080001,0x00000006...
+|c              |Output SPIR-V binary code as a text file containing C-style +
+                 initializer list. +
+                 This is just wrapping the output of `num` option with curly
+                 brackets. +
+                 Example: `glslc -c -mfmt=c main.vert -o output_file.txt` +
+                 Content of output_file.txt: +
+                 {0x07230203, 0x00010000, 0x00080001, 0x00000006...}
+|===
+
 === Warning and Error Options
 
 ==== `-w`
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -14,8 +14,11 @@
 
 #include "file_compiler.h"
 
+#include <cassert>
 #include <fstream>
+#include <iomanip>
 #include <iostream>
+#include <sstream>
 
 #include "file.h"
 #include "file_includer.h"
@@ -27,6 +30,35 @@
 namespace {
 using shaderc_util::string_piece;
 
+// A helper function to emit SPIR-V binary code as a list of hex numbers in
+// text form. Returns true if a non-empty compilation result is emitted
+// successfully. Return false if nothing should be emitted, either because the
+// compilation result is empty, or the compilation output is not SPIR-V binary
+// code.
+template <typename CompilationResultType>
+bool EmitSpirvBinaryAsCommaSeparatedNumbers(const CompilationResultType& result,
+                                            std::ostream* out) {
+  // Return early if the compilation output is not in SPIR-V binary code form.
+  if (!std::is_same<CompilationResultType,
+                    shaderc::SpvCompilationResult>::value)
+    return false;
+  // Return early if the compilation result is empty.
+  if (result.cbegin() == result.cend()) return false;
+  std::ios::fmtflags output_stream_flag_cache(out->flags());
+  *out << std::hex << std::setfill('0');
+  auto RI = result.cbegin();
+  *out << "0x" << std::setw(8) << *RI++;
+  for (size_t counter = 1; RI != result.cend(); RI++, counter++) {
+    *out << ",";
+    // Break line for every four words.
+    if (counter % 4 == 0) {
+      *out << std::endl;
+    }
+    *out << "0x" << std::setw(8) << *RI;
+  }
+  out->flags(output_stream_flag_cache);
+  return true;
+}
 }  // anonymous namespace
 
 namespace glslc {
@@ -168,8 +200,40 @@ bool FileCompiler::EmitCompiledResult(
     }
   }
 
-  // Write compilation output to output file.
-  out->write(compilation_output.data(), compilation_output.size());
+  // Write compilation output to output file. If an output format for SPIR-V
+  // binary code is specified, it is handled here.
+  switch (binary_emission_format_) {
+    case SpirvBinaryEmissionFormat::Unspecified:
+    case SpirvBinaryEmissionFormat::Binary:
+      // The output format is unspecified or specified as binary output.
+      out->write(compilation_output.data(), compilation_output.size());
+      break;
+    case SpirvBinaryEmissionFormat::Numbers:
+      // The output format is specified to be a list of hex numbers, the
+      // compilation output must be in SPIR-V binary code form.
+      assert(output_type_ == OutputType::SpirvBinary);
+      if (EmitSpirvBinaryAsCommaSeparatedNumbers(result, out)) {
+        // Only emits the end-of-line character when the emitted compilation
+        // result is not empty.
+        *out << std::endl;
+      }
+      break;
+    case SpirvBinaryEmissionFormat::CInitList:
+      // The output format is specified to be a C-style initializer list, the
+      // compilation output must be in SPIR-V binary code form.
+      assert(output_type_ == OutputType::SpirvBinary);
+      if (result.begin() != result.end()) {
+        // Only emits the '{' when the compilation result is not empty.
+        *out << "{";
+      }
+      if (EmitSpirvBinaryAsCommaSeparatedNumbers(result, out)) {
+        // Only emits the end-of-line character when the emitted compilation
+        // result is not empty.
+        *out << "}" << std::endl;
+      }
+      break;
+  }
+
   // Write error message to std::cerr.
   std::cerr << result.GetErrorMessage();
   if (out->fail()) {
@@ -247,6 +311,37 @@ bool FileCompiler::ValidateOptions(size_t num_files) {
       std::cerr << "glslc: error: " << dependency_info_dumping_hander_error_msg
                 << std::endl;
       return false;
+    }
+  }
+
+  // If the output format is specified to be a binary, a list of hex numbers or
+  // a C-style initializer list, the output must be in SPIR-V binary code form.
+  if (binary_emission_format_ != SpirvBinaryEmissionFormat::Unspecified) {
+    if (output_type_ != OutputType::SpirvBinary) {
+      std::cerr << "glslc: error: cannot emit output as a ";
+      switch (binary_emission_format_) {
+        case SpirvBinaryEmissionFormat::Binary:
+          std::cerr << "binary";
+          break;
+        case SpirvBinaryEmissionFormat::Numbers:
+          std::cerr << "list of hex numbers";
+          break;
+        case SpirvBinaryEmissionFormat::CInitList:
+          std::cerr << "C-style initializer list";
+          break;
+        case SpirvBinaryEmissionFormat::Unspecified:
+          // The compiler should never be here at runtime. This case is added to
+          // complete the switch cases.
+          break;
+      }
+      std::cerr << " when the output is not SPIR-V binary code" << std::endl;
+      return false;
+    }
+    if (dependency_info_dumping_handler_ &&
+        dependency_info_dumping_handler_->DumpingAsCompilationOutput()) {
+      std::cerr << "glslc: error: cannot dump dependency info when specifying "
+                   "any binary output format"
+                << std::endl;
     }
   }
 
--- a/glslc/src/file_compiler.h
+++ b/glslc/src/file_compiler.h
@@ -29,8 +29,19 @@ namespace glslc {
 // SPIR-V files or preprocessed output.
 class FileCompiler {
  public:
+  enum class SpirvBinaryEmissionFormat {
+    Unspecified,  // No binary output format specified, this is the only valid
+                  // option when the compilation output is not in SPIR-V binary
+                  // code form.
+    Binary,       // Emits SPIR-V binary code directly.
+    Numbers,      // Emits SPIR-V binary code as a list of hex numbers.
+    CInitList,    // Emits SPIR-V bianry code as a C-style initializer list
+                  // of hex numbers.
+  };
+
   FileCompiler()
       : output_type_(OutputType::SpirvBinary),
+        binary_emission_format_(SpirvBinaryEmissionFormat::Unspecified),
         needs_linking_(true),
         total_warnings_(0),
         total_errors_(0) {}
@@ -61,6 +72,11 @@ class FileCompiler {
   // Sets the output filename. A name of "-" indicates standard output.
   void SetOutputFileName(const shaderc_util::string_piece& file) {
     output_file_name_ = file;
+  }
+
+  // Sets the format for SPIR-V binary compilation output.
+  void SetSpirvBinaryOutputFormat(SpirvBinaryEmissionFormat format) {
+    binary_emission_format_ = format;
   }
 
   // Returns false if any options are incompatible. The num_files parameter
@@ -166,7 +182,9 @@ class FileCompiler {
   std::string GetCandidateOutputFileName(std::string input_filename);
 
   // Returns true if the compiler's output is preprocessed text.
-  bool PreprocessingOnly() { return output_type_ == OutputType::PreprocessedText; }
+  bool PreprocessingOnly() {
+    return output_type_ == OutputType::PreprocessedText;
+  }
 
   // Performs actual SPIR-V compilation on the contents of input files.
   shaderc::Compiler compiler_;
@@ -177,6 +195,10 @@ class FileCompiler {
 
   // What kind of output will be produced?
   OutputType output_type_;
+
+  // The Flag to indicate to which format the output SPIR-V binary code should
+  // be emitted.
+  SpirvBinaryEmissionFormat binary_emission_format_;
 
   // A FileFinder used to substitute #include directives in the source code.
   shaderc_util::FileFinder include_file_finder_;
--- a/glslc/src/main.cc
+++ b/glslc/src/main.cc
@@ -59,6 +59,10 @@ Options:
   -std=<value>      Version and profile for input files. Possible values
                     are concatenations of version and profile, e.g. 310es,
                     450core, etc.
+  -mfmt=<format>    Output SPIR-V binary code using the selected format. This
+                    option may be specified only when the compilation output is
+                    in SPIR-V binary code form. Available options include bin, c
+                    and num. By default the binary output format is bin.
   -M                Generate make dependencies. Implies -E and -w.
   -MM               An alias for -M.
   -MD               Generate make dependencies and compile.
@@ -167,6 +171,23 @@ int main(int argc, char** argv) {
         return 1;
       }
       compiler.options().SetTargetEnvironment(target_env, 0);
+    } else if (arg.starts_with("-mfmt=")) {
+      const string_piece binary_output_format =
+          arg.substr(std::strlen("-mfmt="));
+      if (binary_output_format == "bin") {
+        compiler.SetSpirvBinaryOutputFormat(
+            glslc::FileCompiler::SpirvBinaryEmissionFormat::Binary);
+      } else if (binary_output_format == "num") {
+        compiler.SetSpirvBinaryOutputFormat(
+            glslc::FileCompiler::SpirvBinaryEmissionFormat::Numbers);
+      } else if (binary_output_format == "c") {
+        compiler.SetSpirvBinaryOutputFormat(
+            glslc::FileCompiler::SpirvBinaryEmissionFormat::CInitList);
+      } else {
+        std::cerr << "glslc: error: invalid value '" << binary_output_format
+                  << "' in '-mfmt=" << binary_output_format << "'" << std::endl;
+        return 1;
+      }
     } else if (arg.starts_with("-x")) {
       string_piece option_arg;
       if (!GetOptionArgument(argc, argv, &i, "-x", &option_arg)) {
--- a/glslc/test/expect.py
+++ b/glslc/test/expect.py
@@ -19,6 +19,7 @@ as superclass and providing the expected_* variables required by the check_*()
 methods in the mixin classes.
 """
 import os
+import re
 from glslc_test_framework import GlslCTest
 
 
@@ -238,11 +239,20 @@ class ValidFileContents(GlslCTest):
             return False, 'Cannot find file: ' + target_filename
         with open(target_filename, 'r') as target_file:
             file_contents = target_file.read()
-            if file_contents == self.expected_file_contents:
-                return True, ''
-            return False, ('Incorrect file output: \n{act}\nExpected:\n{exp}'
-                           ''.format(act=file_contents,
-                                     exp=self.expected_file_contents))
+            if type(self.expected_file_contents) == str:
+                if file_contents == self.expected_file_contents:
+                    return True, ''
+                return False, ('Incorrect file output: \n{act}\nExpected:\n{exp}'
+                               ''.format(act=file_contents,
+                                         exp=self.expected_file_contents))
+            elif isinstance(self.expected_file_contents, type(re.compile(''))):
+                if self.expected_file_contents.search(file_contents):
+                    return True, ''
+                return False, (
+                    'Incorrect file output: \n{act}\n'
+                    'Expected matching regex pattern:\n{exp}'.format(
+                        act=file_contents,
+                        exp=self.expected_file_contents.pattern))
         return False, ('Could not open target file ' + target_filename +
                        ' for reading')
 
--- a/glslc/test/option_mfmt.py
+++ b/glslc/test/option_mfmt.py
@@ -1,0 +1,197 @@
+# Copyright 2016 The Shaderc Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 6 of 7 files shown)


glslc/test/option_masm.py
```diff
@@ -0,0 +1,127 @@
+# Copyright 2015 The Shaderc Authors. All rights reserved.
```
antiagainst: This is year 2016. :)

Qining: Fixed.

Qining:
PTAL @dekimir @dneto0 @AWoloszyn @antiagainst

glslc/README.asciidoc
```diff
@@ -252,6 +252,26 @@ code, such as identifier names and line numbers.
...
+<<output-dialect-selection,dialect selection table>>
+[[output-dialect-selection]]
+.Output Dialect Selection
+|===
+|Dialect option |Description
+
+|c-init-list    |Output SPIRV binary code as a text file containing C-style +
+                 initializer list. This option is valid when the compilation +
+                 output is SPIRV binary code. It is not valid to use when the +
+                 output is not SPIV binary code, like when `-S`, `-E`, `-M`, +
```
dekimir: typo: SPIR-V

glslc/README.asciidoc
```diff
@@ -252,6 +252,26 @@ code, such as identifier names and line numbers.
...
+Select dialect for compilation output. Supported options are listed in
+<<output-dialect-selection,dialect selection table>>
+[[output-dialect-selection]]
+.Output Dialect Selection
+|===
+|Dialect option |Description
+
+|c-init-list    |Output SPIRV binary code as a text file containing C-style +
+                 initializer list. This option is valid when the compilation +
+                 output is SPIRV binary code. It is not valid to use when the +
```
dekimir: Consider "SPIR-V"?

glslc/README.asciidoc
```diff
@@ -252,6 +252,26 @@ code, such as identifier names and line numbers.
...
+                 output is not SPIV binary code, like when `-S`, `-E`, `-M`, +
+                 `-MM` are turned on. +
+
+                 Example: `glslc -c main.vert -o output_file.txt` +
```
dekimir: I would have expected a `-masm=c-init-list` in this example.

glslc/src/file_compiler.cc
```diff
@@ -168,8 +170,30 @@ bool FileCompiler::EmitCompiledResult(
     }
   }
 
+  // Should write the result as a C-style initializer list if:
+  //  1) the result is SPIRV binary code;
+  //  2) the compiler is set to emit result as a C-style initializer list;
+  //  3) the compiler is not dumping dependency info as compilation output.
+  const bool should_emit_spirv_binary_as_c_initializer_list =
```
dekimir: The above comment just repeats the (already readable) code below it.  I'd rather see an explanation of why we must defend against incompatible flags this far down the call stack.

glslc/src/file_compiler.cc
```diff
@@ -168,8 +170,30 @@ bool FileCompiler::EmitCompiledResult(
...
+        dependency_info_dumping_handler_->DumpingAsCompilationOutput());
+
   // Write compilation output to output file.
-  out->write(compilation_output.data(), compilation_output.size());
+  if (should_emit_spirv_binary_as_c_initializer_list) {
+    // The output should be written to the output stream as a C-style
+    // initializer list.
+    *out << "{";
+    for (uint32_t word : result) {
+      *out << "0x" << std::uppercase << std::setfill('0') << std::setw(8)
```
dekimir: Set the stream formatting outside the loop, because it's the same for all words.

Also, isn't there a way to force `0x` via `std::showbase`?

glslc/src/file_compiler.cc
```diff
@@ -168,8 +170,30 @@ bool FileCompiler::EmitCompiledResult(
...
+           << std::hex << word << ",";
+    }
+    *out << "}" << std::endl;
```
dekimir: Restore the stream's previous format here.

glslc/src/main.cc
```diff
@@ -59,6 +59,11 @@ An input file of - represents standard input.
   -std=<value>      Version and profile for input files. Possible values
                     are concatenations of version and profile, e.g. 310es,
                     450core, etc.
+  -masm=<dialect>   Output SPIRV binary code using selected dialect. The only
+                    supported choice for now is c-init-list. When set to
```
dekimir: s/choice/dialect/ for clarity.

glslc/src/main.cc
```diff
@@ -59,6 +59,11 @@ An input file of - represents standard input.
   -std=<value>      Version and profile for input files. Possible values
                     are concatenations of version and profile, e.g. 310es,
                     450core, etc.
+  -masm=<dialect>   Output SPIRV binary code using selected dialect. The only
```
dekimir: SPIR-V

glslc/src/main.cc
```diff
@@ -59,6 +59,11 @@ An input file of - represents standard input.
...
+                    supported choice for now is c-init-list. When set to
+                    c-init-list, output SPIRV binary code will be emitted as a
+                    C-style initializer list. c-init-list is only valid when
+                    the compilation output is SPIRV binary code.
```
dekimir: SPIR-V

glslc/test/option_masm.py
```diff
@@ -0,0 +1,127 @@
...
+        '0x00040005,0x00000004,0x6E69616D,0x00000000,0x00020013,0x00000002,'
+        '0x00030021,0x00000003,0x00000002,0x00050036,0x00000002,0x00000004,'
+        '0x00000000,0x00000003,0x000200F8,0x00000005,0x000100FD,0x00010038,}'
+        '\n'
+    ]
+
+
+@inside_glslc_testsuite('OptionMasm')
+class TestSingleFileWithLinking(expect.ValidFileContents):
+    """Tests that -masm=c-init-list works when linkding is enabled (no -c
```
dekimir: typo: "linking"

glslc/test/option_masm.py
```diff
@@ -0,0 +1,127 @@
...
+    ]
+
+
+@inside_glslc_testsuite('OptionMasm')
+class TestErrorWhenOutputDisasembly(expect.ErrorMessage):
+    """Tests that specifying '-masm=c-init-list' when the compiler is set to
+    disassembly mode fails the compilation.
+    """
+    shader = FileShader(MINIMAL_SHADER, '.vert')
+    glslc_args = [shader, '-masm=c-init-list', '-S', '-o', 'output_file']
```
dekimir: Is `-o` strictly needed here?

glslc/src/file_compiler.cc
```diff
@@ -250,6 +274,16 @@ bool FileCompiler::ValidateOptions(size_t num_files) {
     }
   }
 
+  // If the output should be emitted as a C-style initializer list, the output
+  // must be in SPIRV binary code form.
+  if (emit_binary_as_c_init_list_ && output_type_ != OutputType::SpirvBinary) {
+    std::cerr
+        << "glslc: error: cannot emit output as a C-style initializer list"
+           " when the output is not SPIRV binary code"
```
dekimir: SPIR-V

glslc/src/file_compiler.h
```diff
@@ -63,6 +64,12 @@ class FileCompiler {
     output_file_name_ = file;
   }
 
+  // Sets the flag to indicate emitting SPIRV binary code as a C-style
```
dneto0: SPIR-V

glslc/test/option_masm.py
```diff
@@ -0,0 +1,127 @@
...
+class TestSingleFileWorksWithDashC(expect.ValidFileContents):
+    """Tests that -masm=c-init-list works with -c for single input file. SPIRV
+    binary code output should be emitted as a C-style initializer list in the
+    output file.
+    """
+    shader = FileShader(MINIMAL_SHADER, '.vert')
+    glslc_args = [shader, '-c', '-masm=c-init-list', '-o', 'output_file']
+    target_filename = 'output_file'
+    expected_file_contents = [
+        '{0x07230203,0x00010000,0x00080001,0x00000006,0x00000000,0x00020011,'
```
dneto0: This is going to be a very fragile test as Glslang evolves.
You should only check the up through the first word since the magic word won't change.

glslc/src/file_compiler.cc
```diff
@@ -168,8 +170,30 @@ bool FileCompiler::EmitCompiledResult(
...
+
   // Write compilation output to output file.
-  out->write(compilation_output.data(), compilation_output.size());
+  if (should_emit_spirv_binary_as_c_initializer_list) {
+    // The output should be written to the output stream as a C-style
+    // initializer list.
+    *out << "{";
+    for (uint32_t word : result) {
+      *out << "0x" << std::uppercase << std::setfill('0') << std::setw(8)
+           << std::hex << word << ",";
```
dneto0: I'd prefer to wrap after every 4 words.
That will make it easier to inspect the result.
Also, I think we shouldn't emit the trailing comma.

AWoloszyn: I have a slight preference to not emitting the "{" as well. It allows this to be used in more contexts, and I don't think it adds any extra complexity to anyone needing to include it.

glslc/README.asciidoc
```diff
@@ -252,6 +252,26 @@ code, such as identifier names and line numbers.
 NOTE: Currently this option has no effect.  Full functionality depends on
 glslang support for generating debug info.
 
+==== `-masm=<dialect>`
+
+Select dialect for compilation output. Supported options are listed in
```
dneto0: I expected -masm=dialect to only take effect when you're already selecting assembly output with -S.
That follows the example of GCC.
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#index-masm_003d_0040var_007bdialect_007d-2837

And in that case,  you should be able to select between c-init-list or spv.

glslc/test/option_masm.py
```diff
@@ -0,0 +1,127 @@
...
+        '0x00030021,0x00000003,0x00000002,0x00050036,0x00000002,0x00000004,'
+        '0x00000000,0x00000003,0x000200F8,0x00000005,0x000100FD,0x00010038,}'
+        '\n'
+    ]
+
+
+@inside_glslc_testsuite('OptionMasm')
+class TestErrorWhenOutputDisasembly(expect.ErrorMessage):
+    """Tests that specifying '-masm=c-init-list' when the compiler is set to
+    disassembly mode fails the compilation.
```
dneto0: I expected this to be the only the case that would work.  :-)

dneto0:
done my review.
High level item:  I think -masm should only take effect with -S

glslc/src/file_compiler.cc
dekimir: @AWoloszyn: that can be controlled by introducing more dialect options.

dneto0:
Discussed offline.
To address my complaint, we'll change the name of the option to -mfmt
and have options "c" for what is now c-init-list, "bin" being the current binary output form (the default), and "num" for a list of comma-separated numbers.

Qining:
@dekimir @dneto0 @AWoloszyn @antiagainst PTAL.

glslc/README.asciidoc
```diff
@@ -252,22 +252,30 @@ code, such as identifier names and line numbers.
...
+|bin            |Output SPIR-V binary code as it is.
+|num            |Output SPIR-V binary code as a text file containing a list of
+                 comma-separated hex numbers.+
+                 Example: `glslc -c -mfmt=num main.vert -o output_file.txt` +
+                 Content of the output_file.txt: +
+                 0x07230203,0x00010000,0x00080001,0x00000006...
+|c              |Output SPIR-V binary code as a text file containing C-style +
+                 initializer list. +
+                 This is just wrapping the output of `num` option with curly
+                 brakets.
```
dneto0: "brackets".

glslc/test/expect.py
```diff
@@ -238,11 +239,18 @@ def check_file(self, status):
...
-            return False, ('Incorrect file output: \n{act}\nExpected:\n{exp}'
-                           ''.format(act=file_contents,
-                                     exp=self.expected_file_contents))
+            if type(self.expected_file_contents) == str:
+                if file_contents == self.expected_file_contents:
+                    return True, ''
+                return False, ('Incorrect file output: \n{act}\nExpected:\n{exp}'
+                               ''.format(act=file_contents,
+                                         exp=self.expected_file_contents))
+            elif type(self.expected_file_contents) == type(re.compile('dummy regex')):
```
antiagainst: I would suggest using `isinstance` and `re.compile('')`.

glslc/README.asciidoc
```diff
@@ -252,22 +252,30 @@ code, such as identifier names and line numbers.
...
-|c-init-list    |Output SPIRV binary code as a text file containing C-style +
-                 initializer list. This option is valid when the compilation +
-                 output is SPIRV binary code. It is not valid to use when the +
-                 output is not SPIV binary code, like when `-S`, `-E`, `-M`, +
-                 `-MM` are turned on. +
-
-                 Example: `glslc -c main.vert -o output_file.txt` +
+|Format option  |Description
+
+|bin            |Output SPIR-V binary code as it is.
```
dneto0: "as it is" -> "as a sequence of binary 32-bitwords in host native endianness.  This is the default."

glslc/README.asciidoc
```diff
@@ -252,6 +252,34 @@ code, such as identifier names and line numbers.
 NOTE: Currently this option has no effect.  Full functionality depends on
 glslang support for generating debug info.
 
+==== `-mfmt=<format>`
+
+Select output format for compilation output in SPIR-V binary code form.
```
antiagainst: Missing sentence subjects. Please follow how the other command line options are described.

glslc/README.asciidoc
```diff
@@ -252,6 +252,34 @@ code, such as identifier names and line numbers.
...
+Supported options are listed in <<binary-output-format-selection,binary output
+format selection table>>. This option is only valid to be used when the
+compilation output is SPIR-V binary code. Specifying any options listed below
+when the output is not SPIR-V binary code, like disassembly (with `-S`
+specified), text (with `-M`, `-MM` or `-E` specified) will trigger an error.
+[[binary-output-format-selection]]
```
antiagainst: It's better to add an empty line before the table.

glslc/README.asciidoc
```diff
@@ -252,6 +252,34 @@ code, such as identifier names and line numbers.
 NOTE: Currently this option has no effect.  Full functionality depends on
 glslang support for generating debug info.
 
+==== `-mfmt=<format>`
+
+Select output format for compilation output in SPIR-V binary code form.
+Supported options are listed in <<binary-output-format-selection,binary output
```
antiagainst: in the ... table.

glslc/src/file_compiler.cc
```diff
@@ -170,30 +171,42 @@ bool FileCompiler::EmitCompiledResult(
...
+      // initializer list. In such cases the compilation output must be in
+      // SPIR-V binary code form.
+      assert(output_type_ == OutputType::SpirvBinary);
+      if (binary_emission_format_ == SpirvBinaryEmissionFormat::CInitList) {
+        *out << "{";
+      }
+      std::ios::fmtflags output_stream_flag_cache(out->flags());
+      *out << std::hex << std::setfill('0');
+      auto RI = result.begin();
+      *out << "0x" << std::setw(8) << *RI++;
```
dneto0: This is a bit fragile:  If result is empty, then you'll have an invalid dereference.
Fortunately SPIR-V modules are always at least 5 words long.

glslc/src/file_compiler.h
```diff
@@ -107,10 +124,22 @@ class FileCompiler {
  private:
   enum class OutputType {
     SpirvBinary,  // A binary module, as defined by the SPIR-V specification.
-    SpirvAssemblyText,  // Assembly syntax defined by the SPIRV-Tools project.
+    SpirvAssemblyText,  // Assembly syntax defined by the SPIR-V-Tools project.
```
antiagainst: Why adding a `-` here?

glslc/src/file_compiler.h
```diff
@@ -64,12 +64,22 @@ class FileCompiler {
     output_file_name_ = file;
   }
 
-  // Sets the flag to indicate emitting SPIRV binary code as a C-style
-  // initializer list for a 32-bit unsigned integer array.
-  void SetEmitSpirvBinaryAsCInitializerList() {
-    emit_binary_as_c_init_list_ = true;
+  // Sets the flag to indicate emitting SPIR-V binary code directly.
+  void SetEmitSpirvBinaryAsBinary() {
```
dneto0: I recommend having a single setter that takes a SpirvBinaryEmissionFormat enum value as its parameter.

glslc/src/main.cc
```diff
@@ -59,11 +59,10 @@ An input file of - represents standard input.
   -std=<value>      Version and profile for input files. Possible values
                     are concatenations of version and profile, e.g. 310es,
                     450core, etc.
-  -masm=<dialect>   Output SPIRV binary code using selected dialect. The only
-                    supported choice for now is c-init-list. When set to
-                    c-init-list, output SPIRV binary code will be emitted as a
-                    C-style initializer list. c-init-list is only valid when
-                    the compilation output is SPIRV binary code.
+  -mfmt=<format>    Output SPIR-V binary code using selected format. This option
```
dneto0: "using the selected"

glslc/src/main.cc
```diff
@@ -59,11 +59,10 @@ An input file of - represents standard input.
...
+                    is valid to be specified only when the compilation output
+                    is in SPIR-V binary code form. Available options includes
+                    bin, c and num.
```
dneto0: "include"

glslc/README.asciidoc
```diff
@@ -252,6 +252,34 @@ code, such as identifier names and line numbers.
...
+format selection table>>. This option is only valid to be used when the
+compilation output is SPIR-V binary code. Specifying any options listed below
+when the output is not SPIR-V binary code, like disassembly (with `-S`
+specified), text (with `-M`, `-MM` or `-E` specified) will trigger an error.
+[[binary-output-format-selection]]
+.Binary Output Format Selections
```
antiagainst: I would suggest "Binary Output Format Options".

glslc/src/main.cc
```diff
@@ -59,11 +59,10 @@ An input file of - represents standard input.
   -std=<value>      Version and profile for input files. Possible values
                     are concatenations of version and profile, e.g. 310es,
                     450core, etc.
-  -masm=<dialect>   Output SPIRV binary code using selected dialect. The only
-                    supported choice for now is c-init-list. When set to
-                    c-init-list, output SPIRV binary code will be emitted as a
-                    C-style initializer list. c-init-list is only valid when
-                    the compilation output is SPIRV binary code.
+  -mfmt=<format>    Output SPIR-V binary code using selected format. This option
+                    is valid to be specified only when the compilation output
```
dneto0: Grammar tweak:    "is valid to be" -> "may be"

glslc/src/file_compiler.h
```diff
@@ -107,10 +124,22 @@ class FileCompiler {
...
   enum class OutputType {
     SpirvBinary,  // A binary module, as defined by the SPIR-V specification.
-    SpirvAssemblyText,  // Assembly syntax defined by the SPIRV-Tools project.
+    SpirvAssemblyText,  // Assembly syntax defined by the SPIR-V-Tools project.
     PreprocessedText,   // Preprocessed source code.
   };
 
+  enum class SpirvBinaryEmissionFormat {
+    Unspecified =
+        0,          // No binary output format specified, this is the only valid
```
antiagainst: Is it necessary to add discriminator to `Unspecified`?

glslc/test/option_mfmt.py
```diff
@@ -0,0 +1,158 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import expect
+from glslc_test_framework import inside_glslc_testsuite
+from placeholder import FileShader
+import re
```
antiagainst: stdlib goes first. ;)

Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/file_compiler.h
Qining: Removed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/main.cc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/README.asciidoc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/main.cc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/main.cc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/file_compiler.h
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/file_compiler.cc
Qining: This does trigger a seg fault when the compilation fails...
Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/README.asciidoc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/test/expect.py
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/README.asciidoc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/README.asciidoc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/README.asciidoc
Qining: Fixed in d7942c13f40e2efafe1ea526b929cc074f6a6280

glslc/src/file_compiler.h
Qining: Hi @dekimir, do you have a suggestion for this? To have the enum value as a parameter, we will need to expose the enum type to main.cc. Do you think that is a better design?

glslc/test/option_mfmt.py
```diff
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
```
dneto0: Shouldn't imports be listed alphabetically?
However, I didn't see that in the Python style guide.

glslc/test/option_mfmt.py
```diff
@@ -0,0 +1,192 @@
...
+
+@inside_glslc_testsuite('OptionMfmt')
+class TestFmtNumWorksWithDashC(expect.ValidFileContents):
+    """Tests that -mfmt=num works with -c for single input file. SPIR-V binary
+    code output should be emitted as a list of hex numbers in the output file.
+    """
+    shader = FileShader(MINIMAL_SHADER, '.vert')
+    glslc_args = [shader, '-c', '-mfmt=num', '-o', 'output_file']
+    target_filename = 'output_file'
+    expected_file_contents = re.compile("^0x07230203.*$", re.S)
```
dneto0: The .*$ here is redundant.  But meh.

glslc/test/option_mfmt.py
```diff
@@ -0,0 +1,192 @@
...
+
+import re
+import expect
+from glslc_test_framework import inside_glslc_testsuite
+from placeholder import FileShader
+
+MINIMAL_SHADER = '#version 140\nvoid main() {}'
+ERROR_SHADER = '#version 140\n#error\nvoid main() {}'
+
+
```
dneto0: If you want a stronger test, i.e. that checks an exact sequence of words, then you could use the assembler feature of glslc. Then you know *exactly* what words should go into the result string.
Also, that way you can also check the line-wrap functionality while still having a tightly controlled sequence of words, since even an empty SPIR-V module compiles to 5 words.

Optional.

glslc/README.asciidoc
```diff
@@ -252,6 +252,39 @@ code, such as identifier names and line numbers.
 NOTE: Currently this option has no effect.  Full functionality depends on
 glslang support for generating debug info.
 
+==== `-mfmt=<format>`
+
+`-mfmt=<format>` selects output format for compilation output in SPIR-V binary
+code form.  Supported options are listed in
```
dneto0: "listed in the" ....

glslc/src/file_compiler.cc
```diff
@@ -27,6 +30,32 @@
 namespace {
 using shaderc_util::string_piece;
 
+// A helper function to emit SPIR-V binary code as a list of hex numbers in
+// text form. Returns true if the compilation result is emitted successfully.
```
dneto0: "Returns true if a non-empty compilation result is emitted successfully."

It wasn't clear to me why *this function* should treat an empty result as a failure.  Just state that upfront.

glslc/src/file_compiler.cc
```diff
@@ -27,6 +30,32 @@
...
+template <typename CompilationResultType>
+bool EmitSpirvBinaryAsCommaSeparatedNumbers(
+    const CompilationResultType& result, std::ostream* out) {
+  // Return early if the compilation output is not in SPIR-V binary code form.
+  if (!std::is_same<CompilationResultType,
+                    shaderc::SpvCompilationResult>::value)
+    return false;
+  // Return early if the compilation result is empty.
+  if (result.begin() == result.end()) return false;
```
dneto0: Since the arg is const, should probably use cbegin() and cend() consistently.
It's just an accident that begin and end return const iterators here.

glslc/src/file_compiler.cc
```diff
@@ -168,8 +197,36 @@ bool FileCompiler::EmitCompiledResult(
     }
   }
 
-  // Write compilation output to output file.
-  out->write(compilation_output.data(), compilation_output.size());
+  // Write compilation output to output file. If an output format for SPIR-V
+  // binary code is specified, it is handled here.
+  switch (binary_emission_format_) {
```
dneto0: I find it weird to asert the output format *inside* the switch.  We should just put the entire switch into an if(output format == binary) clause.

dneto0:
Done review.  Plus advice to expose binary emission format as public inner class to the FileCompiler, and have a single set-bin-output-format method on FileCompiler.

glslc/src/file_compiler.cc
Qining: This is because we have `Unspecified` case in the switch statement. Probably we can pull that out and use an `if-else` to separate the specified/unspecified cases. Then we can have a assert out of the switch.

glslc/test/option_mfmt.py
Qining: Added a constraint at the end to check no trailing comma is there.

glslc/test/option_mfmt.py
Qining: You are right, they should be in alphabetical order. Just fixed.

glslc/src/file_compiler.cc
Qining: Fixed.

dneto0:
+2

Qining:
Squashed and merged

PR #220, opened 2016-06-02
MinGW linking C executables: statically link libgcc

dneto0:
This avoids needing to ship libgcc_s_sjlj-1.dll with your
executable.

The rule still applies to C++ executables.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -19,6 +19,14 @@ function(shaderc_default_c_compile_options TARGET)
       # requires clang to be built with compiler-rt.
       target_link_libraries(${TARGET} PRIVATE --coverage)
     endif()
+    if (NOT SHADERC_ENABLE_SHARED_CRT)
+      if (WIN32)
+	# For MinGW cross compile, statically link to the libgcc runtime.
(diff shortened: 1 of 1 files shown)


cmake/utils.cmake
```diff
@@ -19,6 +19,14 @@ function(shaderc_default_c_compile_options TARGET)
       # requires clang to be built with compiler-rt.
       target_link_libraries(${TARGET} PRIVATE --coverage)
     endif()
+    if (NOT SHADERC_ENABLE_SHARED_CRT)
+      if (WIN32)
+	# For MinGW cross compile, statically link to the libgcc runtime.
+	# But it still depends on MSVCRT.dll.
+	set_target_properties(${TARGET} PROPERTIES
+		LINK_FLAGS "-static -static-libgcc")
```
dekimir: Will this screw up static linking with MSVC? (Assuming we still support that.)

dneto0: No, because this entire block is guarded at line 11 with "not ${MSVC}".

dneto0:
Further testing showed that the C++ path still needed the -static and -static-libgcc arguments.  Pushed a new commit to restore them.

PR #225, opened 2016-06-10
SPIRV-Tools now depends on external SPIRV-Headers repo

dneto0:


antiagainst:
Oh, sorry, my bad.

+2

PR #228, opened 2016-06-20
Fix tests for catching up with the master of glslang.

antiagainst:


glslc/test/option_dash_D.py
```diff
@@ -181,11 +181,11 @@ class TestDashCapDWithDashStdAndVersion(expect.ErrorMessage):
         """#version 310 es
         void main(){X}
         void foo(){Y}""", '.vert')
-    glslc_args = ['-DX=', '-DY=return 3;', '-std=450compatibility', shader]
+    glslc_args = ['-DX=', '-DY=return 3;', '-std=450core', shader]
```
antiagainst: Changed `compatibility` to `core` since otherwise it errors out with:

```
#version: compilation for SPIR-V does not support the compatibility profile
```

This test's purpose is to test `-D` with `-std` and `#version`, it doesn't really matter whether we use `compatibility` or `core` here.

glslc/test/assembly.py
```diff
@@ -124,10 +124,8 @@ class TestShaderStageWithAssemblyFile(expect.ErrorMessage):
     glslc_args = ['-c', '-fshader-stage=vertex', shader]
 
     expected_error = [
-        shader, ': error: #version: Desktop shaders for SPIR-V require '
-        'version 140 or higher\n',
         shader, ":2: error: '' :  syntax error\n",
-        '2 errors generated.\n']
+        '1 error generated.\n']
```
Qining: It seems like this causes a test failure.

antiagainst: It's expected since google/glslang is not updated yet.

PR #230, opened 2016-07-06
Changing Vulkan vs. GL doesn't force teardown of glslang

dneto0:
Glslang's builtin function table cache is now sensitive
to the Spv vs. non-Spv and Vulkan vs. GL.
That is https://github.com/KhronosGroup/glslang/issues/166 is fixed.
So now we don't have to tear down Glslang's internal state when
the "message rules" change. That's how Vulkan vs. GL is encoded.


antiagainst:
Please also submit a patch against the [GTest part of glslang](https://github.com/KhronosGroup/glslang/blob/master/gtests/Initializer.h).

antiagainst:
+2

PR #232, opened 2016-07-15
Fix tests for catching up with upstream glslang

antiagainst:
* Desktop shaders compiling to SPIR-V require version 330 or higher.
* A new control for reporting cascading errors is introduced. (We'd better turn off cascading errors, but that can happen in another PR.)

### DIFF
--- a/glslc/test/assembly.py
+++ b/glslc/test/assembly.py
@@ -124,8 +124,10 @@ class TestShaderStageWithAssemblyFile(expect.ErrorMessage):
     glslc_args = ['-c', '-fshader-stage=vertex', shader]
 
     expected_error = [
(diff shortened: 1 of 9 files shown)


Qining:
+2 lgtm

antiagainst:
It turns out that [SPIRV-Headers](https://github.com/KhronosGroup/SPIRV-Headers) is updated with new SPIR-V versions, which breaks tests in [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools).

So failures caused by submitting this PR is not actually rooted in this PR.

Have https://github.com/KhronosGroup/SPIRV-Tools/pull/266 to fix it.

PR #236, opened 2016-08-05
Fix SHADERC_SKIP_TESTS configuration to actually skip all tests.

antiagainst:
Together with https://github.com/KhronosGroup/SPIRV-Tools/pull/301.

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -21,10 +21,10 @@ function(shaderc_default_c_compile_options TARGET)
     endif()
     if (NOT SHADERC_ENABLE_SHARED_CRT)
       if (WIN32)
-	# For MinGW cross compile, statically link to the libgcc runtime.
-	# But it still depends on MSVCRT.dll.
-	set_target_properties(${TARGET} PROPERTIES
-		LINK_FLAGS "-static -static-libgcc")
+        # For MinGW cross compile, statically link to the libgcc runtime.
+        # But it still depends on MSVCRT.dll.
+        set_target_properties(${TARGET} PROPERTIES
+          LINK_FLAGS "-static -static-libgcc")
       endif(WIN32)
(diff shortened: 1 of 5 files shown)


cmake/utils.cmake
```diff
@@ -21,10 +21,10 @@ function(shaderc_default_c_compile_options TARGET)
     endif()
     if (NOT SHADERC_ENABLE_SHARED_CRT)
       if (WIN32)
-	# For MinGW cross compile, statically link to the libgcc runtime.
-	# But it still depends on MSVCRT.dll.
-	set_target_properties(${TARGET} PROPERTIES
-		LINK_FLAGS "-static -static-libgcc")
+  # For MinGW cross compile, statically link to the libgcc runtime.
```
dneto0: Indentation was damaged here..  ?!

cmake/utils.cmake
```diff
@@ -40,10 +40,10 @@ function(shaderc_default_compile_options TARGET)
     target_compile_options(${TARGET} PRIVATE -std=c++11)
     if (NOT SHADERC_ENABLE_SHARED_CRT)
       if (WIN32)
-	# For MinGW cross compile, statically link to the C++ runtime.
-	# But it still depends on MSVCRT.dll.
-	set_target_properties(${TARGET} PROPERTIES
-		LINK_FLAGS "-static -static-libgcc -static-libstdc++")
+  # For MinGW cross compile, statically link to the C++ runtime.
```
dneto0: Simlar

dneto0:
Ok except for bad changes to indentation

-1

cmake/utils.cmake
antiagainst: Hmm. I was to fix the mixed indent characters earlier. It seems just didn't put enough spaces for tabs.

cmake/utils.cmake
antiagainst: Done.

PR #237, opened 2016-08-09
Build SPIRV-Tools file Instruction.cpp instead of Id.cpp

dneto0:
SPIRV-Tools source file list has changed.


AWoloszyn:
+2 LGTM.

Issue #238, opened 2016-08-10
build invokes python scripts via undefined PYTHON_EXECUTABLE


dneto0:
This is a bug in the CMakeLists.txt files.
This causes the build to fail on Windows.  Instead of running add_copyright.py, it just opens up an editor  (e.g. developer studio) to edit that script.

PR #239, opened 2016-08-10
Invoke python via PYTHON_EXE discovered by cmake

dneto0:
Previously, the build was invoking python via the undefined
variable PYTHON_EXECUTABLE.

Fixes https://github.com/google/shaderc/issues/238


AWoloszyn:
+2 LGTM.

PR #242, opened 2016-09-08
Add util/parse_number.cpp to third_party/Android.mk

Qining:
This CL is to correspond with https://github.com/KhronosGroup/SPIRV-Tools/pull/400


dneto0:
+2

PR #247, opened 2016-09-21
Change to use SPIRV-Tools C++ interface for Disassembling.

antiagainst:


libshaderc_util/src/spirv_tools_wrapper.cc
```diff
@@ -72,15 +45,21 @@ bool SpirvToolsAssemble(const string_piece assembly, spv_binary* binary,
   *binary = nullptr;
   errors->clear();
 
-  const bool result =
+  const bool success =
```
dneto0: Isn't there a tools.Assemble method?

antiagainst: `Assemble()` needs to do a copy to construct a `vector` from `spv_binary`. So for performance reason, I didn't use it here.

dneto0: Ok. Good.   I get the feeling that someday I'll add a way to do binary parsing based where the input binary is presented as begin/end forward iterators.  :-)

PR #248, opened 2016-09-21
Add support for optimization after compilation in the library.

antiagainst:


libshaderc/include/shaderc/shaderc.h
```diff
@@ -74,6 +74,13 @@ typedef enum {
   shaderc_compilation_status_invalid_assembly,
 } shaderc_compilation_status;
 
+// Optimization level.
+typedef enum {
+  shaderc_optimization_level_none = 0,  // no optimization
+  shaderc_optimization_level_zero,      // no optimization
```
AWoloszyn: Is there any specific reason that we have none and zero that look like they mean the same thing.

antiagainst: It's just a conceptual difference; none means not specify any optimization level at all. But that may be too nuanced. I'll remove it.

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -264,10 +278,14 @@ class Compiler {
   // output.
   bool generate_debug_info_;
 
+  // Total Number of optimization passes to perform.
+  uint32_t num_opt_passes_;
```
dneto0: num_opt_passes_ is likely redundant.
Isn't it just enabled_opt_passes_.size()?

libshaderc_util/src/spirv_tools_wrapper.cc
```diff
@@ -79,4 +81,22 @@ bool SpirvToolsAssemble(const string_piece assembly, spv_binary* binary,
...
+                        std::vector<uint32_t>* binary, std::string* errors) {
+  using namespace spvtools;
+
+  errors->clear();
+
+  spvtools::Optimizer optimizer(SPV_ENV_VULKAN_1_0);
+  std::ostringstream oss;
+  optimizer.SetMessageConsumer(
+      [errors](spv_message_level_t, const char*, const spv_position_t&,
+               const char* message) { *errors = message; });
```
dneto0: I think you meant the body to do *oss << message.
Then later do *errors = oss.str()

libshaderc_util/include/libshaderc_util/spirv_tools_wrapper.h
```diff
@@ -34,6 +34,22 @@ bool SpirvToolsAssemble(const string_piece assembly, spv_binary* binary,
 // *text_or_error.
 bool SpirvToolsDisassemble(const std::vector<uint32_t>& binary,
                            std::string* text_or_error);
+
+// Used to tell SpirvToolsOptimize() whether to enable a certain pass.
+//
+// TODO(antiagainst): This may need to consider pass order once we have the
+// interface ready for users to specify optimization on a per pass base.
+// But now let's be simple.
+struct EnableOptimizationPass {
```
dneto0: This is too weak an abstraction.  
It's important to get this right because it's at the API surface.

We have two choices:
1) The SpirvToolsOptimize interface should use the same "levels" of Zero, Size and have that interpreted inside into the more fine grain passes.  Or
2) The SpirvToolsOptimize interface should take a list of the opaque "transform" objects exposed by the spirv-tools optimize.hpp.   Then EnableOptimizationPass should just be vector of those things.  
The upside of 2 is flexibility, the downside is you now expose the interior API details but that's ok because "SpirvTools*" is in the name of the function!  The other extra complexity is that you now need to track *which* object is the strip-debug-info transform object.  But I'm sure there's several ways to solve that.

libshaderc_util/include/libshaderc_util/compiler.h
antiagainst: Previously it's useful since `EnableOptPass` is just a struct. Now not anymore. Removed.

libshaderc_util/src/spirv_tools_wrapper.cc
antiagainst: Right. Actually the previous impl. will just retain the last message. With `*oss <<` we actually have them all.

libshaderc_util/include/libshaderc_util/spirv_tools_wrapper.h
antiagainst: Good suggestion! I've changed the impl. to use `std::vector` and now we have order support.

I've also added support for unify-constant, but have no only of testing it since this is optimizing the code compiled via glslang directly.

libshaderc_util/src/spirv_tools_wrapper.cc
```diff
@@ -79,4 +82,43 @@ bool SpirvToolsAssemble(const string_piece assembly, spv_binary* binary,
   return result;
 }
 
+bool SpirvToolsOptimize(const std::vector<PassId>& enabled_passes,
+                        std::vector<uint32_t>* binary, std::string* errors) {
+  using namespace spvtools;
```
dneto0: Remove this.

antiagainst: Done.

antiagainst:
@dneto0: addressed all comments and squashed. Ready to be merged.

libshaderc_util/src/spirv_tools_wrapper.cc
```diff
@@ -79,4 +82,41 @@ bool SpirvToolsAssemble(const string_piece assembly, spv_binary* binary,
...
+                        std::vector<uint32_t>* binary, std::string* errors) {
+  errors->clear();
+  if (enabled_passes.empty()) return true;
+  if (std::all_of(
+          enabled_passes.cbegin(), enabled_passes.cend(),
+          [](const PassId& pass) { return pass == PassId::kNullPass; })) {
+    return true;
+  }
+
+  spvtools::Optimizer optimizer(SPV_ENV_VULKAN_1_0);
```
dneto0: Hm.   This is a smell.  It seems like we should be passing in the environment or a context object.

But ok for now.

antiagainst: Sure. I see the issue reported. Will do that latter.

PR #252, opened 2016-09-26
glslc support for optimization levels

antiagainst:


antiagainst:
@dneto0: PTAL

glslc/README.asciidoc
```diff
@@ -249,9 +251,29 @@ the current working directory.
...
+pass.
+
 NOTE: Currently this option has no effect.  Full functionality depends on
 glslang support for generating debug info.
 
+==== `-O0`, `-Os`
+
+`-O` specifies which optimization level to use:
+
+* `-O0` means "no optimization". This level compiles the fastest and generates
```
dneto0: "compiles the fastest" is too easily mis-interpreted as "makes the code run as fast as possible"

glslc/README.asciidoc
```diff
@@ -249,9 +251,29 @@ the current working directory.
...
+  the most debuggable code.
+* `-Os` enables optimizations to reduce code size. This level runs the
+  following optimization passes one by one:
+  - strip-debug-info: removes all debug instructions. (This pass will be turned
```
dneto0: I'd rather not describe the details of what this does.  It will get out of date pretty quickly.

glslc/README.asciidoc
antiagainst: I copied that from clang. :)
Removed.

glslc/README.asciidoc
antiagainst: Removed.

But I think we still need a way to tell the user what exactly we are running in case they are interested?

dneto0: Maybe, but it's going to get stale.  Maybe add a verbose option to print the pass names?

antiagainst: That's reasonable. And... for clang:

http://stackoverflow.com/questions/15548023/clang-optimization-levels

It seems `-###` is the trick.

Anyway, we can worry about this later.

antiagainst: #253 for this.

Issue #251, opened 2016-09-26
spirv_tools_wrapper.h should take an environment enum or be based on a context object


dneto0:

dneto0:
Right now the optimizer method hardcodes Vulkan 1.0, which is based on SPIR-V 1.0.  That affects what instruction tables are available.
This will become an issue once graphics APIs start using SPIR-V 1.1 or later.

antiagainst:
Done via #255.

PR #255, opened 2016-09-30
Plumb target environment to SPIRV-Tools

antiagainst:
Address #251.

* Chanage the C interface for assembling to accept options.

Also together with the C interface change, new overloads for assembling
are added into the C++ interface.

* Sort out target environments used in different components.

SPIRV-Tools has its own target environment definition, and public
shaderc API has its own. A new one is added into `Compiler` in
`shaderc_util`, for better isolation.

Now `shaderc_util::Compiler`'s interface accepts target environment,
not glslang messages anymore. All the conversion to glslang messages
are done in the implementation now.

### DIFF
--- a/libshaderc_util/include/libshaderc_util/compiler.h
+++ b/libshaderc_util/include/libshaderc_util/compiler.h
@@ -111,6 +111,8 @@ class Compiler {
     Size,  // Optimization towards reducing code size.
   };
 
+  // Creates an default compiler instance targeting at Vulkan environment. Uses
+  // version 110 and no profile specification as the default for GLSL.
   Compiler()
       // The default version for glsl is 110, or 100 if you are using an es
       // profile. But we want to default to a non-es profile.
--- a/libshaderc_util/include/libshaderc_util/compiler.h
+++ b/libshaderc_util/include/libshaderc_util/compiler.h
@@ -22,7 +22,6 @@
 #include <utility>
 
 #include "glslang/Public/ShaderLang.h"
-#include "libshaderc_util/spirv_tools_wrapper.h"
 
 #include "counting_includer.h"
 #include "file_finder.h"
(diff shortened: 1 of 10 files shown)


libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -111,7 +121,7 @@ class Compiler {
         suppress_warnings_(false),
```
dneto0: Maybe the constructor blurb should say it will default the target env to Vulkan.

dneto0: Alternately, there could be a Default = Vulkan value  enum class TargetEnv.  Then it would be more clear.
But I like forcing the callers to think about what they mean.
So I think I want to avoid a Default value.

libshaderc_util/src/spirv_tools_wrapper.cc
```diff
@@ -36,13 +38,25 @@ void SpirvToolsOutputDiagnostic(spv_diagnostic diagnostic, std::string* dest) {
...
-}  // anonymous namespace
+// Gets the corresponding target environment used in SPIRV-Tools.
+spv_target_env GetSpirvToolsTargetEnv(Compiler::TargetEnv env) {
+  switch (env) {
+    case Compiler::TargetEnv::Vulkan:
+      return SPV_ENV_VULKAN_1_0;
+    case Compiler::TargetEnv::OpenGL:
+      return SPV_ENV_OPENGL_4_5;
+    case Compiler::TargetEnv::OpenGLCompat:
+      return SPV_ENV_OPENGL_4_5;  // TODO(antiagainst): is this correct?
```
dneto0: I'm not 100% sure about this, but let's go with it.  I don't think that OpenGL restricted what versions could be used with compatibility profile.

dneto0:
FYI: this fixes a long-standing request I had to avoid using Glslagn's EshMessages enum in Shaderc's library interface.  Nice.

libshaderc_util/include/libshaderc_util/compiler.h
antiagainst: Done in the following commit.

PR #256, opened 2016-10-01
Clarify Shaderc wraps around other tools

dneto0:
Describe why it exists: what does Shaderc add on top?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,25 @@ At the moment it includes:
 - [`glslc`](glslc), a command line compiler for GLSL to SPIR-V, and
 - `libshaderc` a library API for doing the same.
 
+Shaderc wraps around core functionality in
+[Glslang](https://github.com/KhronosGroup/glslang)
+and [SPIRV-Tools](https://github.com/KhronosGroup/SPIRV-Tools).  Shaderc aims to
+to provide:
+* a command line compiler with GCC- and Clang-like usage, for better
+  integration with build systems
+* an API where functionality can be added without breaking existing clients
+* an API supporting standard concurrency patterns across multiple
+  operating systems
+* increased functionality such as file `#include` support
+
 ## Status
 
-Shaderc is experimental, and subject to significant incompatible changes.
+Shaderc is approaching maturity, but is still subject to incompatible changes.
+
+Shaderc has been shipping in the
+[Android NDK](https://developer.android.com/ndk/index.html) since version r12b.
+(The NDK build uses sources from https://android.googlesource.com/platform/external/shaderc/.
+Those repos are downstream from GitHub.)
 
 For licensing terms, please see the [`LICENSE`](LICENSE) file.  If interested in
 contributing to this project, please see [`CONTRIBUTING.md`](CONTRIBUTING.md)


dneto0:
Looking for feedback from @johnkslang: Does this clarification help?  Am I missing anything else significant?

johnkslang:
Looks good.  

Somehow, glslang should have native #include functionality.  I can see some aspect of it might depend on environment.  When file-base processing is done offline, it makes sense to use #include functionality that is file-system based, and that's true of both glslang and Shaderc, so seems more a current accident of functionality levels than of a true reason to distinguish between the tools.  Is there a way of further characterizing what should be distinct about the #include functionality present in Shaderc?

antiagainst:
I feel like the newly added paragraph is kinda overlap with what is already in the `File Organization` Section. But I see they emphasize different things. So I'm OK.

antiagainst:
As a side comment, it may be handy for developers if we create tags for each NDK shipment.

dneto0:
@johnkslang:  I agree that having some of the file-based part of #include functionality in a separate library is partly an accident.  However it did meet our goal of trying to minimize the amount of code that went into glslang itself, since *all* clients would pay the code-size cost.  It's fair to argue the dividing line was chosen poorly, and perhaps should have just talked with you about it before implementing.  Other factors were: we were far more tentative at the time, there was no gtest-based facility yet in Glslang at the time, and we felt strongly about the need for fine-grain testing of the feature, but without blocking its rollout.

@antiagainst: About the tags.  Any tags in Shaderc are not automatically carried over into the AOSP repos that the NDK builds use.  So it would be a manual process to go back and tag *this* repository with what ended up in the NDK.   In any case, the NDK release always have git commit signatures for their own contents, in "repo.prop" files.  Those signatures reference the commits in the AOSP repos.  I prefer to let that automatic mechanism be the one and only place where that info is recorded, and avoid inserting a manual and hence error-prone step here.
Maybe the right thing to do is to point to the repos at https://android.googlesource.com/platform/external/shaderc/ for the authoritative content of what's actually in the NDK. 
This would also accommodate the potential case where a patch is applied to the NDK copy of the code but which hasn't been pushed back into GitHub.  (I hope to never do that, but ....)

dneto0:
I've added a commit to point to where NDK takes its sources from, and to say that that's downstream from GitHub

johnkslang:
@dneto0 Perfectly understandable how we got here! I only want the #include thing rationalized rather than being a defining characteristic of the differences between shaderc and glslang.

dneto0:
In https://github.com/google/shaderc/pull/256/commits/d5411d8443ece0c24f5bd1f6280929c175711413 I've added a parenthetical remark about how `#include` support might fully move into Glslang.
I see how the previous wording could be understood that support for `#include` is a *defining* characteristic of Shaderc.  It's just an example of where functionality materially differs today.   (My basic point was that we need to tell people it's there.  Even better would be to finally publish the extension doc.)

Note that full file-based `#include` support is also closely tied to dependency-dumping features of glslc, i.e. the -M family of options.

johnkslang:
Yes, I think #include is complex in that it has environment-specific behavior, and hence will have support in multiple forms.  Thanks for clarifying.

dneto0:
Rebased and squashed.

PR #257, opened 2016-10-05
When copying tests, only catch ENOENT exception

dneto0:
Before, we were catching too-broad a set of exceptions.
On Windows sometimes you might not be able to remove the
original directory, e.g. since a process might be running
in that directory.  That can lead to spurious test failures where
tests would run but compare to stale base results.

Also:
- sort the test file list, for nicer debugging.
- fix linting warnings for unused variables.

### DIFF
--- a/utils/copy-tests-if-necessary.py
+++ b/utils/copy-tests-if-necessary.py
@@ -23,6 +23,7 @@ intermediate-dir is optional, it specifies that there the additional directory
 between the root, and the tests/binary.
 """
 
+import errno
 import os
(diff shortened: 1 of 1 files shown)


dneto0:
The Travis failure looks spurious.
https://travis-ci.org/google/shaderc/builds/165319733

antiagainst:
@dneto0: I usually click the button on the top right to restart the errored job. It will typically be fine the second run.

dneto0:
Rebased

PR #264, opened 2016-10-21
Rename GlslInitialzer -> GlslangInitializer

dneto0:


dneto0:
Rebased and pushed into master as e6b4795

PR #266, opened 2016-10-21
glslc: expose HLSL compilation support from Glslang

dneto0:
Support "-x hlsl", i.e. hlsl is a valid language specifier.

Files with .hlsl extension default to HLSL.

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -11,8 +11,7 @@
 
 ----
 glslc [-c|-S|-E]
-      [-x glsl] [-std=standard]
-      [-x hlsl]
+      [-x ...] [-std=standard]
       [-fshader-stage=...]
       [--target-env=...]
       [-g]
@@ -191,7 +190,7 @@ under Vulkan semantics.
 ==== `-x`
 
 `-x` lets you specify the language of the input shader files. Valid languages
-are 'glsl' and 'hlsl'.  If the file extension is `hlsl` then the default language
+are `glsl` and `hlsl`.  If the file extension is `hlsl` then the default language
 is HLSL.  Otherwise the default is 'glsl'.
 
 [[compilation-stage-selection-options]]
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -12,6 +12,7 @@
 ----
 glslc [-c|-S|-E]
       [-x glsl] [-std=standard]
+      [-x hlsl]
       [-fshader-stage=...]
       [--target-env=...]
       [-g]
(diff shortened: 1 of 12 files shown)


glslc/README.asciidoc
```diff
@@ -12,6 +12,7 @@
 ----
 glslc [-c|-S|-E]
       [-x glsl] [-std=standard]
+      [-x hlsl]
```
antiagainst: I think we should write something like `[-x ...]` here to be consistent with the following.

BTW, `-std` should mention that it is not applicable to `-x hlsl` (if this is correct).

glslc/README.asciidoc
```diff
@@ -189,8 +190,9 @@ under Vulkan semantics.
 
 ==== `-x`
 
-`-x` lets you specify the language of the input shader files. Right now, the
-only accepted argument is `glsl`.
+`-x` lets you specify the language of the input shader files. Valid languages
+are 'glsl' and 'hlsl'.  If the file extension is `hlsl` then the default language
```
antiagainst: formatting: backtick for `glsl` and `hlsl`

glslc/src/main.cc
```diff
@@ -329,15 +347,22 @@ int main(int argc, char** argv) {
...
       // If current_fshader_stage is shaderc_glsl_infer_from_source, that means
       // we didn't set forced shader kinds (otherwise an error should have
       // already been emitted before). So we should deduce the shader kind
       // from the file name. If current_fshader_stage is specifed to one of
       // the forced shader kinds, use that for the following compilation.
-      input_files.emplace_back(
-          arg.str(), current_fshader_stage == shaderc_glsl_infer_from_source
-                         ? glslc::DeduceDefaultShaderKindFromFileName(arg)
-                         : current_fshader_stage);
+      input_files.emplace_back(InputFileSpec{
```
antiagainst: I think we can directly place the three parameters in `emplace_back`?

glslc/README.asciidoc
dneto0: Yeah, I wasn't sure.  :-0
Next commit tests that -std is ignored for HLSL, and documents it.

glslc/README.asciidoc
dneto0: Done in next commits

dneto0:
All feedback addressed.

glslc/src/main.cc
antiagainst: I see.

dneto0:
Squashed and merged into master as cdefe18ac0

PR #271, opened 2016-11-08
C, C++ APIs are sensitive to entry_point_name

dneto0:
Before, the C API ignored the entry point name.
Before, the C++ API didn't have entry points that accepted
an entry point name.

### DIFF
--- a/libshaderc/CMakeLists.txt
+++ b/libshaderc/CMakeLists.txt
@@ -23,7 +23,7 @@ target_link_libraries(shaderc PRIVATE SPIRV-Tools)
 shaderc_add_tests(
   TEST_PREFIX shaderc
   LINK_LIBS shaderc
-  INCLUDE_DIRS include ${glslang_SOURCE_DIR}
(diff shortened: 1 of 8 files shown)


AWoloszyn:
I am a bit hesitant to have more api-breaking changes. Although I understand the need in this case. The biggest problem is that anyone using the HPP at this point, ( and the .h in other changes) has no compile-time way to know which version of the API they are using. I would really like to get in, (before this change or in the future soon) a way to be able to

```
#if SHADERC_API_VERSION < N
  - Do Old Thing
#else
  - Do new Thing
#endif

dneto0:
@AWoloszyn This is not a breaking change.
The C API signature is the same. It's just that now we pass the entry point name down into Glslang.
The C++ API gains new methods, and the old ones stay around, and use "main" like they always have.  The diff might look a little weird in that I inserted the expanded-parameter-list method before the existing method.   The lack of breakage of existing tests verify that the old API still works.

AWoloszyn:
My mistake! I apologize, in that case LGTM!

PR #276, opened 2016-11-17
Support resource limits in C, C++ API

dneto0:


libshaderc/include/shaderc/shaderc.h
```diff
@@ -255,6 +342,10 @@ void shaderc_compile_options_set_target_env(shaderc_compile_options_t options,
 void shaderc_compile_options_set_warnings_as_errors(
     shaderc_compile_options_t options);
 
+// Sets a resource limit.
+void shaderc_compile_options_set_limit(
+    shaderc_compile_options_t options, shaderc_limit limit, int value);
```
antiagainst: I'm not particular clear about this part tho: should we use an explicitly sized integer type here?

libshaderc/include/shaderc/shaderc.h
```diff
@@ -86,6 +86,93 @@ typedef enum {
   shaderc_optimization_level_size,  // optimize towards reducing code size
 } shaderc_optimization_level;
 
+// Resource limits.
+typedef enum {
```
AWoloszyn: It is unfortunate to have to specify everything twice, once in the .inl file and once here, but I understand the desire to keep the header as clean as possible.

libshaderc_util/include/libshaderc_util/resources.inc
```diff
@@ -0,0 +1,100 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// These are the resource limits in a glslang::TBuiltInResource.
+// The first field is the string name to be used in a configuration setting.
+// The second field is the fieldname in TBuiltInResource.
```
antiagainst: But I see three fields...

libshaderc/include/shaderc/shaderc.h
dneto0: Yeah.  This is like circularity we've seen before:
* the public header is in libshaderc/shaderc.h
* but libshaderc_util code has to talk about those enums to translate down into Glslang.
So we need two mirrored enum sets.  The best I can do is the single resources.inc file that makes the association.

libshaderc/include/shaderc/shaderc.h
dneto0: Glslang uses int.  If we went explicitly sized then there could potentially be an inconvenient  mismatch someday.

libshaderc_util/include/libshaderc_util/resources.inc
dneto0: good catch.  :-)

dneto0:
Fixed the comment in resources.inc.
But I have a legitimate gcc failure to fix.

dneto0:
Squashed

PR #279, opened 2016-11-28
Update tests for more strict glslango

dneto0:
Glslang now warns when an HLSL shader entry point can't
be found that either matches the default of "main" or
one specified via -fentry-point.


AWoloszyn:
Other than the commit message "glslango" looks good.

dneto0:
Comments addressed.  :-)

dneto0:
Note. The bots will not go green because this requires a simultaneous update to google/glslang.

PR #285, opened 2016-12-13
Add rule for creating SPIRV-Tools generators.inc file

dneto0:
This goes with https://github.com/KhronosGroup/SPIRV-Tools/pull/504


dneto0:
Rebased and pushed into master as 4ff1d25

PR #13, opened 2015-08-18
Added a single library that contains everything needed to use libshaderc

AWoloszyn:
libshaderc_combined contains all of its own static dependencies, and
is a single library that can be linked into a program.

PR #287, opened 2016-12-13
Set the tool to Google Shaderc over Glslang

dneto0:
It's the top 16 bits of the generator word.

Test output expects a relatively recent versions of:
- SPIRV-Tools: parses tools identities from the SPIR-V XML Registry file
- SPIRV-Headers: defines the "Google over Shaderc" generator as tool #13

### DIFF
--- a/libshaderc_util/src/compiler.cc
+++ b/libshaderc_util/src/compiler.cc
@@ -230,6 +230,14 @@ std::tuple<bool, std::vector<uint32_t>, size_t> Compiler::Compile(
   // Note the call to GlslangToSpv also populates compilation_output_data.
   glslang::GlslangToSpv(*program.getIntermediate(used_shader_stage), spirv);
 
+  // Set the tool field (the top 16-bits) in the generator word to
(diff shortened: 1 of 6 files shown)


libshaderc_util/src/compiler.cc
```diff
@@ -230,6 +230,14 @@ std::tuple<bool, std::vector<uint32_t>, size_t> Compiler::Compile(
   // Note the call to GlslangToSpv also populates compilation_output_data.
   glslang::GlslangToSpv(*program.getIntermediate(used_shader_stage), spirv);
 
+  // Set the tool field (the top 16-bits) in the generator word to
+  // 'Shaderc over Glslang'.
```
AWoloszyn: Nit:
Elsewhere we use 'Google Shaderc over Glslang'

dneto0: Point taken. 
The vendor field is "Google" and the tool field is "Shaderc over Glslang".
The SPIRV-Tools disassembler jams the two together.  But that's a detail we don't need to address here.

PR #293, opened 2017-01-04
Allow trailing whitespace in version lines in CHANGES

dneto0:
This enables the case where the source code is checked
out with DOS line endings on a Linux system.  There will
be a trailing carriage return before Python sees the line
ending.


dneto0:
Rebased

PR #294, opened 2017-01-05
Adapt to Glslang include message changes

dneto0:
Also the #include directive now requires a newline at the end.
That is a requirement from C++.

This is required so we can refresh google/shaderc from upstream


dneto0:
Bot failures are expected.  This change *requires* the corresponding update to google/glslang

PR #295, opened 2017-01-07
Adapt to Glslang Includer interface changes.

dneto0:
See Glslang test https://github.com/KhronosGroup/glslang/pull/663

All Shaderc tests pass with this change and the Glslang change.


dneto0:
Rebased.  Removed WIP since the Glslang change is now in its master branch.
Please review.

dneto0:
FYI. I expect the bots to fail since they point to google/glslang.  This change requires a simultaneous  refresh of that tree from upstream KhronosGroup/glslang.

dneto0:
Shaderc are lots of #include tests under glslc/test.   So coverage is ok.

PR #297, opened 2017-01-12
Update Travis CI configuration.

antiagainst:
* We do not require sudo privilege.
* No need for brew update.
* Mac OS X is renamed to macOS.
* Reduce the clone depth of third party projects to speed up.
* `nproc`'s counterpart on macOS is `sysctl -n hw.ncpu`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ compiler:
 matrix:
   fast_finish: true # Show final status immediately if a test fails.
   exclude:
-    # Skip GCC builds on Mac OS X.
+    # Skip GCC builds on macOS.
     - os: osx
       compiler: gcc
   include:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,7 +21,7 @@ compiler:
 matrix:
   fast_finish: true # Show final status immediately if a test fails.
   exclude:
-    # Skip GCC builds on Mac OS X.
+    # Skip GCC builds on macOS.
```
dneto0: "macOS" looks less official than "Mac OS X".
Doesn't really matter to me.

antiagainst: That's what Apple decides. :)

antiagainst:
Well, it seems once I turn `sudo` off, the compiler crashes. I vaguely remember previously we have the same problem because of resource limit on Travis. Maybe the same issue. Anyway, keep `sudo` for now.

PR #298, opened 2017-01-19
SPIRV-Tools added source/validate_decorations.cpp

dneto0:


dneto0:
Auto +2 for this kind of change.

PR #299, opened 2017-02-03
Auto bind uniforms

dneto0:


glslc/README.asciidoc
```diff
@@ -128,6 +129,20 @@ multiple files generated. A filename of `-` represents standard output.
 
 === Language and Mode Selection Options
 
+[[option-f-auto-bind-uniforms]]
+==== `-fauto-bind-uniforms`
```
antiagainst: I'm a little bit unclear about the parameter name. Should we call it `*uniforms` or `*resources`? It feels like to me that `uniform` is from the old OpenGL/GLSL world, while [in Vulkan, resources](https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/html/vkspec.html#descriptorsets) is what used to stand for those data passed in via bindings.

And I really don't like the way they call *everything* as uniforms.

glslc/test/option_fauto_bind_uniforms.py
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import expect
+from glslc_test_framework import inside_glslc_testsuite
+from placeholder import FileShader
+
+# A GLSL shader with uniforms without explicit bindings.
+GLSL_SHADER_WITH_UNIFORMS_WITHOUT_BINDINGS = """#version 140
```
antiagainst: I am curious about a test case that has some resources with explicit bindings but some not. But that's essentially testing glslang. Hmm. Never mind.

antiagainst:
Generally LGTM. But I'm strongly leaning towards calling the option as `-fauto-bind-resources`.

AWoloszyn:
Unless we are also auto-binding other types of resources (SSBOs for example), then I would prefer to keep this more explicit.

```glsl
layout (set=0, binding=1) buffer myStorageBuffer
{
    vec4 myElement[];
};

**Comment by antiagainst** [comment]
Yep, that depends on how glslang implements it. I haven't checked the code for auto binding in glslang yet, but it would be surprising to me that glslang only supports the `uniform` keyword in GLSL. From the perspective of resource binding, they are no difference. If that's true, then `-fauto-bind-uniforms` is fine but later we may need `-fauto-bind-buffers` or something.

File: glslc/test/option_fauto_bind_uniforms.py (hunk 2 above)
Comment (dneto0) [inline glslc/test/option_fauto_bind_uniforms.py]: I too was on the fence about this scenario.

Yeah, I tried that manually.  Glslang makes sure they don't overlap.
As far as I can tell, Glslang didn't document the binding scheme, which is why I was reluctant to test that far.
This feature is for the case where source doesn't have any, because it originates from something other than Vulkan.

**Comment by dneto0** [comment]
I confirmed that the --amb option in Glslang does not apply to SSBOs.

```
#version 450

uniform Block { int x; int y; } my_block;
buffer Buffer { int x; int y; } my_buff;

void main() {
int x = my_block.x;
x = my_buff.x;
}
```
Produces a Binding decoration only for my_block. Both my_block and my_buff get DescriptorSet bindings.

dneto0:
Rebased

PR #300, opened 2017-02-03
Add CMake configuration to install shaderc libraries.

antiagainst:


antiagainst:
glslang and SPIRV-tools already install themselves. So it makes sense to install shaderc artifacts to complete the picture.

PR #301, opened 2017-02-06
Fix installation of libshaderc_combined.a on Windows.

antiagainst:


libshaderc/CMakeLists.txt
```diff
@@ -38,8 +38,18 @@ shaderc_combine_static_lib(shaderc_combined shaderc)
...
 # install() directive to install it. Install it like a normal file.
-get_target_property(shaderc_combined_location shaderc_combined LOCATION)
-install(FILES ${shaderc_combined_location} DESTINATION lib)
+get_target_property(generated_location shaderc_combined LOCATION)
+string(REGEX MATCH "Visual Studio .*" vs_generator "${CMAKE_GENERATOR}")
+if (NOT "${vs_generator}" STREQUAL "")
+  # With Visual Studio generators, the LOCATION property is not properly
+  # expanded according to the current build configuration. We need to work
+  # around this problem by manually substitution.
+  string(REPLACE "$(Configuration)" "\${CMAKE_INSTALL_CONFIG_NAME}"
```
dneto0: Is it really $(Configuration) instead of ${Configuration} ?

antiagainst: Yes. That's actually what cmake uses to make up the path of the artifact. Clearly wrong, right? :)

dneto0: Ew.

PR #306, opened 2017-02-27
Enable NV_EXTENSIONS for Android build of glslang

dgkoch:


dgkoch:
@dneto0 appreciated. Our Android build system is slightly different.

dneto0:
Ok. This passes my internal Android builds.  It works with very latest sources from upstream KhronosGroup/glslang as well.
I'll wait for the Travis and AppVeyor bots to come back green before pushing onto master.

PR #307, opened 2017-03-01
Use OpenCL extended instruction set from SPIRV-Headers

dneto0:


dneto0:
travis-ci has a partial service outage which appears to be causing very long delays in its bot checks for OSX.  I will go ahead with the merge.

PR #309, opened 2017-03-03
Add comment to explain how to return an error in callback

realitix:
It's a small addition but very useful!


realitix:
I signed it!

dneto0:
Yes, you're right.  We missed documenting that in this interface.

This interface eventually tunnels down into Glslang, and we document the "empty string means error" convention there:
https://github.com/KhronosGroup/glslang/blob/master/glslang/Public/ShaderLang.h#L335
But that doesn't help you here.  :-(

This code looks good.  I plan to add a note at the source_name data member to describe the convention.

dneto0:
Rebased and pushed into master as  48b5f885b

dneto0:
I've pushed cb4f0f6261 to describe the error convention in the includer result struct itself.

dneto0:
Thank you for the contribution!

PR #310, opened 2017-03-06
Add Bindings section in Readme

realitix:
It's a good idea to reference shaderc binding in this Readme.
What do you think about that ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,3 +216,7 @@ directory.
 [khr-glslang]: https://github.com/KhronosGroup/glslang
 [google-glslang]: https://github.com/google/glslang
 [spirv-tools]: https://github.com/KhronosGroup/SPIRV-Tools
+
+## Bindings
+
+**Python:** [pyshaderc](https://github.com/realitix/pyshaderc)


dneto0:
Yes, I'm willing to add your binding to the readme once I can make it work.

I tried to do "pip install pyshaderc".  But it failed.
Do I have to put libshader_combined.a in a special place for it to be found?

realitix:
Thanks @dneto0 for your interest. I have updated the packaging, can you try again ? If there is an error, can you paste the error.

dneto0:
I've taken a closer look.  I used "pip install --no-clean pyshaderc" to try to install but also avoid cleaning up its build files.

This is how it fails:
```
x86_64-linux-gnu-gcc -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-Bsymbolic-functions -Wl,-z,relro -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -D_FORTIFY_SOURCE=2 -g -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security build/temp.linux-x86_64-2.7/build/temp.linux-x86_64-2.7/_pyshaderc.o -L/tmp/pip_build_foo/pyshaderc/_cffi_build -lshaderc_combined -lstdc++ -o build/lib.linux-x86_64-2.7/pyshaderc/_pyshaderc.so

/usr/bin/ld: /tmp/pip_build_foo/pyshaderc/_cffi_build/libshaderc_combined.a(shaderc.cc.o): unrecognized relocation (0x2a) in section `.text'

/usr/bin/ld: final link failed: Bad value

```

I see that you have a copy of libshaderc_combined.a in the _cffi_build subdirectory.
I moved that aside and placed my own copy of libshaderc_combined.a there.  Then the above step passes.

If this discussion goes on much longer, maybe it should move to an issue on your repo. :-)

realitix:
I have created an issue in my repo: https://github.com/realitix/pyshaderc/issues/1

antiagainst:
I used to have a hard time to compile shaderc and use it via FFI in Rust. I end up feeding CMake with extra  configuration to make it work, especially for Windows.

Here is the build script: https://github.com/google/shaderc-rs/blob/master/build/build.rs. Hope it is useful in some sense.

realitix:
Thanks @antiagainst !
I'm wondering about something. I ship libshaderc_combined.a but I have compiled it on my machine (linux 64 bits), it can't work on a 32 bit machine I think. I should put a static library for each environment and detect it during the `pip install`. What do you think about that @dneto0 ?

dneto0:
realitix@: I would just ship a 32-bit version.  Hopefully we don't get shader compilation that requires more than a couple of gigabytes of memory!

realitix:
@dneto0 I will not provide the static library, I am going to do like @antiagainst , i will compile the library during the install.

realitix:
@dneto0 Can you try again, pyshaderc 1.0.9 should work everywhere now since it compiles shaderc on the fly.

realitix:
Maybe you could add the @antiagainst rust binding too.

dneto0:
Yes, it would be good to add the rust binding as well.  Let's do that as part of a separate PR.

FYI.  I've been maintaining a known-good branch in shaderc that has a JSON file listing what commits from various projects work together.  There's also a simple Python script to read that file and check out the right sources in the right configuration.   That can be used to avoid guesswork about what should work with what, and might be helpful to people maintaining 3rd party bindings.  I need to describe this in the README

realitix:
@dneto0 Good to know the `know-good` branch! I will add that information in my readme too

realitix:
This pull request is ok since you have added manually the content. Thanks

PR #317, opened 2017-03-17
Add pervasive dependency to extension_enum.inc

dneto0:
Before the first build, the Android.mk build system can't see
for itself the transitive dependency via
table.h -> extensions.h -> extension_enum.inc.
Since table.h is so fundamental, make all (non-test) .cpp files
depend on extension_enum.inc.


third_party/Android.mk
```diff
@@ -112,6 +112,62 @@ LOCAL_PATH := $(SPVTOOLS_LOCAL_PATH)
...
 
+SPVTOOLS_SRC_FILES := \
+		source/assembly_grammar.cpp \
+		source/binary.cpp \
+		source/diagnostic.cpp \
+		source/disassemble.cpp \
+		source/ext_inst.cpp \
+		source/enum_string_mapping.cpp \
+		source/extensions.cpp \
+		source/extensions.h \
```
AWoloszyn: Can we remove this, as it does not participate in the build. (And causes ndk-build to throw warnings)

dneto0:
Squashed and pushed into master.

PR #319, opened 2017-03-18
Support repos where .git is a file, not a dir

chadversary:
In repos created with the git-worktree and git-submodule
commands, '.git' is a text file that points to the real git directory.

update_shaderc_sources.py fails on such repos. The fix is an easy
one-liner.


dneto0:
Rebased and pushed into known-good as 1486bd038a0

PR #320, opened 2017-03-18
Re-add the 'known-good' remote if missing

chadversary:
If a subrepo exists but does not have a remote named 'known-good', then
update_shaderc_sources.py fails. This situation can happen if you try to
re-use an existing repo that you cloned manually without
update_shaderc_sources.py; or if you accidentally delete the
'known-good' remote.

To fix this, teach the script to check if the remote exists before
running git-fetch. If the remote is missing, then run `git remote add`.

### DIFF
--- a/update_shaderc_sources.py
+++ b/update_shaderc_sources.py
@@ -34,7 +34,7 @@ SITE_TO_HOST = { 'github' : 'github.com' }
 VERBOSE = True
 
 
-def command_output(cmd, directory):
+def command_output(cmd, directory, fail_ok=False):
     """Runs a command in a directory and returns its standard output stream.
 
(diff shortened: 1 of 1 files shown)


update_shaderc_sources.py
```diff
@@ -34,7 +34,7 @@
 VERBOSE = True
 
 
-def command_output(cmd, directory):
+def command_output(cmd, directory, fail_ok=False):
```
dneto0: Nit:  I would have added phrased this in the positive sense:  "muss_pass" option that defaults to True.

Not blocking though.

dneto0:
Rebased and pushed into known-good as ff245919042a

PR #321, opened 2017-03-18
Skip fetch when subrepo already has known-good commit

chadversary:
This speeds up the update script. It completes in
0.08 seconds on my machine when all known-good commits are present.


dneto0:
Rebased and pushed into known-good as  fcbba654

Issue #327, opened 2017-03-22
glslc writes bad dependency file if `-o -`


chadversary:
If glslc is given option '-o -' and one of the -M* dependency
options, then it wrote an incorrect dependency file. For example:

    $ glslc -MD -S -o - a.glsl | sed ...
    $ ls
    -.d a.glsl
    $ cat ./-.d
    -: a.glsl

 It doesn't make sense to write a dependency file when the compilation
  output is stdout. So glslc should catch this error when parsing the command line flags.j

chadversary:
More investigation. I think glslc should allow `-MD` with `-o -`, but only when `-MT <depfile>` is also given.

chadversary:
NEVERMIND. glslc correctly behaves the same as gcc when given `-o -` with `-M*` options. Here is a test that demonstrates the gcc behavior.

```
#!/bin/bash

setup() {
    # Clean!
    rm -f hello* ./-*

    cat >hello.c <<-EOF
	#include "hello.h"
	
	int main(void) {
	    hello();
	    return 0;
	}
	EOF

    cat >hello.h <<-EOF
	#include <stdio.h>
	
	static inline void
	hello(void) {
	    printf("hello!\n");
	}
	EOF
}

start_test() {
    test_name="$1"
    echo "test $test_name ..."
}

pass() {
    echo "test $test_name ... pass"
}

fail() {
    echo "test $test_name ... fail"
    exit 1
}

testA() {
    start_test "A"
    setup || fail
    gcc -S -o - -MMD hello.c >/dev/null || fail
    [[ -f ./-.d ]] || fail
    [[ $(cat ./-.d) = "-: hello.c hello.h" ]] || fail
    pass
}

testB() {
    start_test "B"
    setup || fail
    gcc -S -o - -MMD -MF hello.o.d -MT hello.o hello.c >/dev/null || fail
    [[ -f hello.o.d ]] || fail
    [[ $(cat hello.o.d) = "hello.o: hello.c hello.h" ]] || fail
    pass
}

testA
testB
```

dneto0:
My favourite kind of bug.  :-)

Thanks for the explanation.   Yes, we used gcc as the model for this feature.  I'm glad we are close enough to it in behaviour.

PR #328, opened 2017-03-22
glsl: Don't write dependency info when compilation output is stdout

chadversary:
If glslc is given option '-o -' and one of the -M* dependency
options, then it wrote an incorrect dependency file. For example:

    $ glslc -MD -S -o - a.glsl | sed ...
    $ ls
    -.d a.glsl
    $ cat ./-.d
    -: a.glsl

It doesn't make sense to write a dependency file when the compilation
output is stdout. So catch the error while validating command line
flags, emit a helpful error message, and fail.

Fixes: https://github.com/google/shaderc/issues/327

### DIFF
--- a/glslc/src/file_compiler.cc
+++ b/glslc/src/file_compiler.cc
@@ -321,6 +321,15 @@ bool FileCompiler::ValidateOptions(size_t num_files) {
                 << std::endl;
       return false;
     }
+
+    if (output_file_name_ == "-") {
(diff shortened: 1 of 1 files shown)


chadversary:
I'm closing this pull request. I'll re-open if and when I have time to fix the CI failures.

chadversary:
NEVERMIND. glslc correctly behaves the same as gcc regarding this issue. See https://github.com/google/shaderc/issues/327 for details.

PR #330, opened 2017-04-12
Update tests for different Glslang error message

dneto0:
This is necessary before we can refresh google/glslang from upstream

### DIFF
--- a/glslc/test/option_dash_x.py
+++ b/glslc/test/option_dash_x.py
@@ -19,6 +19,9 @@ from placeholder import FileShader
 MINIMAL_SHADER = "#version 140\nvoid main(){}"
 # This one is valid GLSL but not valid HLSL.
 GLSL_VERTEX_SHADER = "#version 140\nvoid main(){ gl_Position = vec4(1.0);}"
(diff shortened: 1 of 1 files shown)


glslc/test/option_dash_x.py
```diff
@@ -57,6 +60,17 @@ class TestDashXHlslOnGlslShader(expect.ErrorMessageSubstr):
 
     shader = FileShader(GLSL_VERTEX_SHADER, '.vert')
     glslc_args = ['-x', 'hlsl', '-c', shader]
+    # This is not a terribly good error message since HLSL doesn't support
```
antiagainst: Maybe you mean "bad" actually? ;-P

dneto0:
Note: Builds all fail because this  change *requires* the corresponding update to google/glslang.

glslc/test/option_dash_x.py
dneto0: :-) 
I actually filed https://github.com/KhronosGroup/glslang/issues/834 about the quality of the error message.  If that gets "fixed" then I'll have to react again.

Thanks for the LGTM.

PR #331, opened 2017-04-13
React to Glslang #version error message for HLSL

dneto0:


dneto0:
Required to refresh google/glslang from upstream

dneto0:
The bots should fail on this since google/glslang is not updated yet.

Issue #334, opened 2017-05-09
GCC 7.1 issues noexcept-type warning


dneto0:
libshaderc_util/include/libshaderc_util/string_piece.h:85:10: error: mangled name for ‘size_t shaderc_util::string_piece::find_first_not_matching(T) [with T = int (*)(int) throw ()]’ will change in C++17 because the exception specification is part of a function type [-Werror=noexcept-type]
   size_t find_first_not_matching(T callee) {
          ^~~~~~~~~~~~~~~~~~~~~~~

It seems informative, but I don't know how to fix the code.  I've tried adding noexcept(noexcept(callee('a')))  to the template signature to no avail.

PR #335, opened 2017-05-09
Avoid noexcept-type in GCC 7.1 for test code

dneto0:
Fixes https://github.com/google/shaderc/issues/334


dneto0:
An alternative is to delete the templated function.  It's only used in tests.  :-)

dneto0:
Rebased. Keeping +2

PR #336, opened 2017-05-11
Added bit_stream.cpp to third_party/Android.mk

atgoo:


dneto0:
Closing because I already made an equivalent change: 44823cb4

PR #338, opened 2017-05-12
Avoid mem leaks in counting includer test

dneto0:
This is caught by leak checks in Clang 3.6 address sanitizer.

This is just one of many memory leaks in our tests.  :-(

### DIFF
--- a/libshaderc_util/src/counting_includer_test.cc
+++ b/libshaderc_util/src/counting_includer_test.cc
@@ -14,8 +14,10 @@
 
 #include "libshaderc_util/counting_includer.h"
 
-#include <gmock/gmock.h>
 #include <thread>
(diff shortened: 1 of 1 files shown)


dneto0:
Rebased.  Keeping the +2

antiagainst:
Actually we can use GitHub UI to do the rebasing now. Pay attention to the merging button you'll see a triangle and then you can select "rebase and merge". Will save the pain of manually rebasing every time. :)

dneto0:
But I'm still old-fashioned.  When I rebase I re-run the tests.  Also I can rerun an Android build for a few ABIs.  :-)

antiagainst:
That's fair. But for small trivial changes, maybe we can leverage it.

PR #341, opened 2017-05-25
Add -h to glslc

dneto0:
It's the same as --help


antiagainst:
It's interesting that GCC and Clang does not support `-h` as an option to print the help message. I don't know what's the reason behind. If we are to be "consistent" with GCC/Clang, maybe we should also avoid supporting `-h`. But I know their behavior are not unassailable rules.

dneto0:
Interesting.  I thought it was just an oversight on our part.  Never mind then.

PR #343, opened 2017-06-07
Set binding base numbers for various resource types, add --hlsl-iomap

dneto0:


dneto0:
Test failure was that same istringstream bug I found when making SPIRV-Tools util/parse_number.h.   So apply the same fix.  But simpler here since I only need an unsigned integer rather than a fully general template.

dneto0:
Rebased

dneto0:
Rebased

glslc/src/main.cc
```diff
@@ -144,6 +180,69 @@ bool SetResourceLimits(const std::string& str, shaderc::CompileOptions* options,
...
+// consumed.  If no stage name is provided, then *stage contains
+// shaderc_glsl_infer_from_source.
+bool GetOptionalStageThenOffsetArgument(const shaderc_util::string_piece option,
+                                        std::ostream* errs, int argc,
+                                        char** argv, int* index,
+                                        shaderc_shader_kind* shader_kind,
+                                        uint32_t* offset) {
+  int& argi = *index;
+  if (argi + 1 >= argc) {
+    *errs << "glsc: error: Option " << option
```
dneto0: Thanks.  Fixed in next commit.

glslc/src/main.cc
```diff
@@ -200,6 +310,44 @@ int main(int argc, char** argv) {
       }
     } else if (arg == "-fauto-bind-uniforms") {
       compiler.options().SetAutoBindUniforms(true);
+    } else if (arg == "-fhlsl-iomap") {
+      compiler.options().SetHlslIoMapping(true);
+    } else if (arg == "-fimage-binding-base") {
```
dneto0: That's a pretty cool trick.   I thought about making a lookup table, but your technique is better.  I'll try it out and see if I like it. For another CL though.

glslc/src/main.cc
```diff
@@ -156,6 +255,17 @@ int main(int argc, char** argv) {
   glslc::FileCompiler compiler;
   bool success = true;
   bool has_stdin_input = false;
+  shaderc_shader_kind arg_stage;  // Shader stage for a single option.
+  uint32_t arg_base;              // Binding base for a single option.
```
dneto0: Done in next commit.

dneto0:
Squashed a few of the fixups to clean up the history.  But kept most commits separate.

PR #348, opened 2017-06-13
Add -fhlsl-offsets, and -fresource-set-binding

dneto0:


glslc/src/main.cc
```diff
@@ -336,6 +343,43 @@ int main(int argc, char** argv) {
                                               &arg_stage, &arg_base))
         return 1;
       set_binding_base(arg_stage, u_kind, arg_base);
+    } else if (arg == "-fresource-set-binding") {
+      auto need_three_args_err = [arg]() {
+        std::cerr << "glsc: error: Option " << arg
```
antiagainst: We can actually embed the content of `arg` directly here to avoid the capture?

glslc/README.asciidoc
```diff
@@ -218,6 +221,36 @@ If a stage is specified, only affects the specified stage.
 
 This only affects HLSL compilation.
 
+[[option-f-register-set-binding]]
```
dneto0: Fixed in next commit

dneto0: Fixed in next commit.

glslc/README.asciidoc
```diff
@@ -218,6 +221,36 @@ If a stage is specified, only affects the specified stage.
...
+==== `-fresource-set-binding`
+
+Option `-fresource-set-binding [stage] <reg0> <set0> <binding1>` sets
```
dneto0: Good catch!  Fixed in next commit.

glslc/src/main.cc
dneto0: Fixed in next commit.

dneto0:
All feedback addressed.  Also, rebased against master (including #351)

PR #350, opened 2017-06-14
Fix Glslang test dir path for MSVC build

dneto0:


dneto0:
Grr. I need to get to a Windows machine to test this.

dneto0:
Second attempt.  Let's see what the bots say.

Issue #353, opened 2017-07-04
CMake Build use GNUInstallDirs


d3x0r:
Should use GNUInstallDirs module to get proper output names for install.  Not all systems have 'lib' as the target for libraries, but sometimes use lib64, and this library takes care of the target names.

https://cmake.org/cmake/help/v3.4/module/GNUInstallDirs.html

include( GNUInstallDirs )
and then use ${CMAKE_INSTALL_LIBDIR} to get the correct name of library to install to. (lib64 vs lib)
and ${CMAKE_INSTALL_BINDIR} for target binary directory (although that seems to always be 'bin' anyway)

dneto0:
Fixed on master with squash/rebase of your PR:  see commit e27569f020

PR #354, opened 2017-07-04
Build Updates; use GNUInstallDirs; optionally disable install

d3x0r:
#353 
Add option to prevent installation; since the product is a static library, projects which use this won't need the source and headers installed.

Ran into an issue in that 'third_party' directory is empty, but in the released LunarG Vulkan SDK there are lots more CMakeLists.txt to update accordingly.


dneto0:
Rebased, squashed, and pushed into master as e27569f02

PR #358, opened 2017-07-31
Adapt to Glslang error changes

dneto0:
- Glslang issues an error for unknown version or profile
- Glslang supports 320es, so that changes one error message.


dneto0:
I expect bots to fail because this requires a refresh of google/glslang from upstream KhronosGroup/glslang.    In fact, this pull request *enables* that refresh.

PR #359, opened 2017-08-01
Add Appveyor configs for VS 2015, VS 2017

dneto0:


antiagainst:
Are we dropping Visual Studio 2013 support?

antiagainst:
glslang and SPIRV-Tools supports Visual Studio 2013.

So I think may be we should check 2013 and 2015

dneto0:
I didn't mean to exclude VS2013.  I'll add it now.
The main purpose was to add VS2017 because LunarG hit an issue in SPIRV-Tools just before their latest Vulkan SDK release, and they shipped an older version of Shaderc instead.

PR #360, opened 2017-08-02
Support GLSL 4.6 and ESSL 3.2

dneto0:
Requires updated sources from KhronosGroup/glslang


dneto0:
The bots should fail since updated glslang code changes an error message; now complaining about extraneous semicolons. 
So this MR can only go in if we also update google/glslang

Issue #363, opened 2017-08-18
Default build target should not depend on HTML generation via asciidoctor


dneto0:
Some asciidoctor installations are broken in a mysterious way.  HTML documentation generation is not critical, so don't force the default build target to depend on it.

PR #364, opened 2017-08-18
Don't build HTML docs by default

dneto0:
Some asciidoctor installations are broken in a mysterious way. HTML
documentation generation is not critical, so don't force the default
build target to depend on it.

Fixes https://github.com/google/shaderc/issues/363


dneto0:
Thanks @antiagainst !  I'll wait until bots pass.

dneto0:
Rebased. Keeping approval.

Issue #356, opened 2017-07-14
Memory Leaks caused by Singletons


Malacath92:
Using shaderc in our code consistently causes ~5.5MB of leaked memory. The issue seems to stem from singletons in shaderc_compiler_initialize and some callee in shaderc_compile_into_spv(CompileToSpecifiedOutputType).
Subsequent compiler initializations don't cause additional leaks.

I understand that resources will be cleared when exiting the program, but we don't want keep around that memory, the example-usage in shaderc.hpp seem to imply that it wasn't necessary as well.
On top of that it shadows our memory leaks since a total of 1252 leaks are caused by shaderc.

See this super simple code example that suffers from the leaks:
```
#include <iostream>

#include <shaderc\shaderc.hpp>

int main(int argc, char **argv) {
	{
		shaderc_compiler_t compiler = shaderc_compiler_initialize();
		{
			char *shader = "#version 430 void main() {}";
			size_t size = strlen(shader);
			shaderc_compilation_result_t compileResult = shaderc_compile_into_spv(
				compiler, shader, size,
				shaderc_shader_kind::shaderc_glsl_default_fragment_shader, "test.frag", "main", nullptr);
			shaderc_result_release(compileResult);
		} // Leaks 5259.32 KB
		shaderc_compiler_release(compiler);
	} // Leaks 386.32 KB
	return 0;
}
````

**Comment by dneto0** [comment]
I ran your example through valgrind: `valgrind --leak-check=full --show-leak-kinds=all `

I'm attaching the report here:
[log.txt](https://github.com/google/shaderc/files/1149629/log.txt)

The biggest leak is a glslang data structures (symbol tables?).
It could be that Shaderc isn't properly cleaning up glslang upon release, but I haven't verified that.  Glslang has some interesting logic to handle table initialization in the presence of threads, and I'm not clear on how all that is cleaned up, or if it can be.

**Comment by Malacath92** [comment]
If we simply remove the static allocation of the GlslangInitializer in shaderc_compiler_initialize and rewrite it together with shaderc_compiler_release to:
```
shaderc_compiler_t shaderc_compiler_initialize() {
  shaderc_compiler_t compiler = new (std::nothrow) shaderc_compiler;
  compiler->initializer = new shaderc_util::GlslangInitializer;;
  return compiler;
}

void shaderc_compiler_release(shaderc_compiler_t compiler) {
	delete compiler->initializer;
	delete compiler;
}
```
[...]

PR #365, opened 2017-08-25
Fixed leaking compiler->initializer

Jasper-Bekkers:
https://github.com/google/shaderc/issues/356

Fixes a big part of that issue. 

Someone still needs to fix the leak for InitializeMemoryPools().

### DIFF
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -449,14 +449,17 @@ void shaderc_compile_options_set_hlsl_register_set_and_binding(
 }
 
 shaderc_compiler_t shaderc_compiler_initialize() {
-  static shaderc_util::GlslangInitializer* initializer =
+  shaderc_util::GlslangInitializer* initializer =
       new shaderc_util::GlslangInitializer;
   shaderc_compiler_t compiler = new (std::nothrow) shaderc_compiler;
   compiler->initializer = initializer;
   return compiler;
 }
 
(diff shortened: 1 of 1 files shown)


Jasper-Bekkers:
I signed it!

dneto0:
I'd like @AWoloszyn to review this.  He designed the original mechanism.

dneto0:
Oh, wait.  I think this completely breaks the protection scheme that GlslangInitializer is supposed to be implementing.  Different compiler objects can end up destroying Glslang's internal state while a different compiler instance is still using it.
See the TODO comment at GlslangInitializer's definition.

libshaderc/src/shaderc.cc
```diff
@@ -449,14 +449,17 @@ void shaderc_compile_options_set_hlsl_register_set_and_binding(
 }
 
 shaderc_compiler_t shaderc_compiler_initialize() {
-  static shaderc_util::GlslangInitializer* initializer =
+  shaderc_util::GlslangInitializer* initializer =
```
AWoloszyn: This will unfortunately fail in the case where you have 2 instances of the compiler at once.
I absolutely understand and agree with what you want to do, and why.

I think the right solution here would be to have the GlslangInitializer be basically a static refcount. In shaderc_compiler_release we decrement the refcount, and clear the state.

There would need to be a bit of extra locking but I thin it should work in the general case.

dneto0:
Closing.
#356 is already on file to address memory leaks.

PR #377, opened 2017-09-26
Serialize inclusions by the CountingIncluder

dneto0:
Should fix ASAN failures in the CountingIncluderTest:
The ConcreteCountingIncluder include delegate is not safe
to call concurrently.  It's just simpler/nicer for the base
class to protect us in this way.


dneto0:
Rebased

PR #379, opened 2017-09-27
Appveyor: deploy to BinTray instead of GitHub Release page

antiagainst:
This should fix the current deploy failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,9 +71,10 @@ Shaderc into.
 
 ## Getting and building Shaderc
 
-On Windows, instead of building from source, you can get the artifacts for
-the top of the tree of the master branch deployed by the Appveyor continuous
-integration service from [BinTray][bintray].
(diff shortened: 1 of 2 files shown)


antiagainst:
Example: https://bintray.com/antiagainst/spirv-toolchain/shaderc#files

README.md
```diff
@@ -73,7 +73,7 @@ Shaderc into.
 
 On Windows, instead of building from source, you can get the artifacts for
```
dneto0: Please call this experimental. Especially because it's on a BinTray account owned by you rather than by Google.

Also say "The underlying service is not hosted by Google or by GitHub."

antiagainst: Done

antiagainst:
Landed as https://github.com/google/shaderc/commit/97e0e134c3f2f5fe214f089d66016119c7e8bd80

PR #386, opened 2017-10-02
Allow SPIRV-Headers to be checked out at third_party/

antiagainst:
This makes the directory hierarchy less complicated.


dneto0:
Waiting for bots

PR #387, opened 2017-10-02
Include SPIRV-Tools before glslang

antiagainst:
This makes glslang link to SPIRV-Tools for optimiztion correctly.


dneto0:
Will wait for bots to pass.

PR #394, opened 2017-10-10
Add HLSL legalization passes to compiler.cc

atgoo:
Run a group of optimization passes if translating from HLSL. Can be
disabled with compiler_.EnableHlslLegalization(false), enabled by
default.

### DIFF
--- a/libshaderc_util/src/compiler.cc
+++ b/libshaderc_util/src/compiler.cc
@@ -243,8 +243,13 @@ std::tuple<bool, std::vector<uint32_t>, size_t> Compiler::Compile(
   // 'spirv' is an alias for the compilation_output_data. This alias is added
   // to serve as an input for the call to DissassemblyBinary.
   std::vector<uint32_t>& spirv = compilation_output_data;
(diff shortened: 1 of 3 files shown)


atgoo:
OS X build failure is unrelated to the CL.

libshaderc_util/src/compiler.cc
```diff
@@ -243,8 +243,13 @@ std::tuple<bool, std::vector<uint32_t>, size_t> Compiler::Compile(
   // 'spirv' is an alias for the compilation_output_data. This alias is added
   // to serve as an input for the call to DissassemblyBinary.
   std::vector<uint32_t>& spirv = compilation_output_data;
+  glslang::SpvOptions options;
+  options.generateDebugInfo = false;
+  options.disableOptimizer = true;
```
dneto0: This was surprising to see, but yes, this maintains previous behaviour.
I got tripped up because there are still OpName instructions in the output, but that's Glslang's choice for now.

dneto0:
Rebased and pushed into master as 16f8253e0

PR #396, opened 2017-10-12
Travis on OSX: Install updated ruby

dneto0:
That's needed by the brew install


dneto0:
This is my guess of what's needed.  Let's see what the bots say

antiagainst:
Instead of doing this, I think we can just use `make` instead of `ninja` to build.

dneto0:
Closing in favour of using "make" instead.  :-(

PR #400, opened 2017-10-20
Allow override of SPVTOOLS_LOCAL_PATH

dneto0:


third_party/Android.mk
```diff
@@ -104,4 +104,9 @@ LOCAL_C_INCLUDES:=$(GLSLANG_LOCAL_PATH) \
 LOCAL_STATIC_LIBRARIES:=OSDependent OGLCompiler SPIRV HLSL
 include $(BUILD_STATIC_LIBRARY)
 
+# Set the location of SPIRV-Tools.
+# All the user to overrid it, but default it to under our third_party directory.
+SPVTOOLS_LOCAL_PATH ?= $(THIRD_PARTY_PATH)/spirv-tools
```
atgoo: What if the variable exist but is empty? This seems more reliable.
```
ifeq ($(SPVTOOLS_LOCAL_PATH),)
  SPVTOOLS_LOCAL_PATH:=$(THIRD_PARTY_PATH)/spirv-tools
endif
```

dneto0: That's what the ?= syntax means.  :-)

See https://www.gnu.org/software/make/manual/html_node/Setting.html#Setting

atgoo: No, it doesn't. It doesn't check the value of the variable, but its origin.

If you set
export SPVTOOLS_LOCAL_PATH=bla/bla/bla
and the reset it
export SPVTOOLS_LOCAL_PATH=
then the variable still exists but is empty. ?= will do nothing.

But if you do
unset SPVTOOLS_LOCAL_PATH
then the variable wouldn't exist anymore and ?= will set new value.

dneto0: Ok, it's not identical.  :-)

If someone does SPVTOOLS_LOCAL_PATH=  then we should trust them.
They're trying to build, with sources hanging off the root directory.  Weird, but ok.

atgoo: Most people use
VAR=
instead of
unset VAR

dneto0: Ok. I give in.
Rebased and squashed. 

Please take another look.

PR #402, opened 2017-10-24
Avoid -fPIC for MinGW builds

dneto0:
Android NDK build uses a MinGW build that warns if -fPIC is specified
because PIC is always turned on anyway.  That causes an error since
we have warnings as errors.


dneto0:
This modifies a commit by @Kangz where support for a shared library was added.

I've built under MinGW build with my proposed PR, and tested on a Windows machine, and it passes, provided the libshaderc_shared.dll is placed appropriately.

PR #408, opened 2018-01-02
Adapt to Glslang codegen version update

dneto0:
Glslang codegen updates to 3, for memory barrier semantics changes.

This is required to refresh google/glslang from upstream KhronosGroup/glslang


dneto0:
> I assume the bot will be happy after refreshing google/glslang.

Yes.  
I refreshed google/glslang at the same time as pushing this into master.

PR #411, opened 2018-01-12
Make tests resilient to better SPIR-V optimizations

dneto0:
A test shader should use the resources or samplers whose
bindings are being checked.


dneto0:
Bots will fail until Glslang adjusts to latest SPIRV-Tools, e.g. by including https://github.com/KhronosGroup/glslang/pull/1221 or better.

PR #416, opened 2018-01-31
WIP:  Force SPIR-V 1.0 binaries

dneto0:
That's the only version of SPIR-V accepted by Vulkan 1.0 and OpenGL 4.6
client APIS.


dneto0:
This is WIP because the Glslang API call doesn't actually force the SPIR-V version.
See https://github.com/KhronosGroup/glslang/issues/1236

dneto0:
Superceded by https://github.com/google/shaderc/pull/445
(Also, this PR was a response to buggy behaviour in Glslang that was quickly fixed.)

PR #418, opened 2018-02-01
Deprecate OpenGL compatibility profile for SPIR-V compilation

dneto0:
Disable its tests.
    
    Glslang doesn't support SPIR-V code generation for OpenGL
    compatibility mode.  GL_ARB_gl_spirv disables OpenGL compatibility
    mode, since SPIR-V doesn't support certain features.
    
    Shaderc still offers this as a mode, but it will likely cause
    compilation failure.


dneto0:
This PR includes code from #416

dneto0:
Amending-only.   Rerun with updated Glslang, which includes updates to its test golden results based on newer SPIRV-Tools.
I expect this to pass bots now.

dneto0:
Rebased.  Keeping the LGTM.

dneto0:
Superceded by #453

PR #420, opened 2018-02-06
Glslang updated its generator version number to 4

dneto0:


dneto0:
Requires an update to google/glslang to pass the bots

PR #425, opened 2018-02-26
Glslang now requires bindings on resources

dneto0:
Either supply a binding via layout annotation in the shader,
or use the auto-map-uniforms option.

This goes along with a refresh of google/glslang from upstream.  Bots will fail on this PR since they won't pick up the Glslang changes in https://github.com/KhronosGroup/glslang/pull/1256


dneto0:
Rebased.

PR #428, opened 2018-02-28
Update glslc README for -fauto-map-locations

dneto0:
Added to the "Code generation options" section.

### DIFF
--- a/glslc/README.asciidoc
+++ b/glslc/README.asciidoc
@@ -20,7 +20,8 @@ glslc [-c|-S|-E]
       [-x ...] [-std=standard]
       [ ... options for resource bindings ... ]
       [-fhlsl-offsets]
-      [-fentry-point=...]
+      [-fentry-point=<name>]
+      [-fauto-map-locations]
       [-flimit=...]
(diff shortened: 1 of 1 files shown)


glslc/README.asciidoc
```diff
@@ -349,11 +350,27 @@ Use HLSL packing rules instead of GLSL rules when determining offsets of
...
+
+Option `-fauto-map-locations` directs the compiler to automatically assign
+location numbers to user-defined stage input and output variables.
+
+Client APIs normally require adjacent stages to agree on their I/O interface.
+The compiler only sees one stage at a time, so it is strongly recommended that
+you avoid relying on this option to assign locations.
+
+Instead, an explicit binding number should be specified in the shader source by
+using a `location` layout qualifier.  For example:
```
antiagainst: It's unclear to me whether this feature is specifically for GLSL (which the example is) or both GLSL and HLSL. For HLSL, I think we should promote `[[vk::location(X)]]` as the suggested way since it is supported by both compilers.

dneto0:
All feedback addressed.

PR #439, opened 2018-03-06
Add build result badges to README.

ehsannas:


dneto0:
Cool!  It appears to have much lower latency than the existing CI boks.

But help me understand.

1. Does this eliminate the Travis-CI bots? (Due to triggers change somehwere/how?) The Travis-CI status is not displayed on *this* PR.   But the Appveyor test is running.

2.The "Details" on each of those builds references something which looks like it's not publicly viewable.     ??

3. The windows build failed in this case.  By default ctest doesn't emit failure details, so it's not very useful.  Either add --output-on-failure to ctest, or set CTEST_OUTPUT_ON_FAILURE=1 in the test environment to see error details.

ehsannas:
It is indeed much faster than existing bots. Also all launch in parallel, and all usually finish within 30 minutes of launch.

1. Travis bots are still running (it's at the very bottom of the long list of jobs!).
2. By design logs are not public, and there is no way to change that :-(
3. Sounds good. I will add that to the build scripts for windows.

Thanks

dneto0:
1. Thanks.   Yikes the UI was... suboptimal?  (See, I blamed something other than me.  :-) )

2. Ooof.  Failures are not actionable until someone at Google copy/pastes the problem.

3. Thanks

AWoloszyn:
You may want 
core.eol = lf

Not 100% sure though

ehsannas:
:+1: Thanks @AWoloszyn ! I'll give that a try... I'll try anything at this point! :)

kokoro/linux/build.sh
```diff
@@ -102,7 +102,7 @@ fi
 echo $(date): Starting ctest...
 if [ $SKIP_TESTS = "False" ]
 then
-  ctest --output-on-failure
+  ctest --output-on-failure -j4
```
AWoloszyn: ctest --output-on-failure `nproc` 

Should work in both msys, and normal bash.

AWoloszyn: Actually that wont work on Mac, but 
ctest --output-on-failure `getconf _NPROCESSORS_ONLN` does seem to work on Mac/Win(msys)/Linux

ehsannas: ah! Thanks 👍

dneto0: I'm ok with  a -j4 for now. We can fine-tune later.
We've had apparent problems with too much parallelism on the free CI systems, and have tuned down to j4 there. Who knows what the Kokoro machines are like.

ehsannas: According to [kokoro documentation](https://g3doc.corp.google.com/devtools/kokoro/g3doc/development/gcp_ubuntu_config.md#os-kernel-cpu-memory), all the linux/mac/windows machines have 4 logical cores at the moment. But that'll likely increase in the future.

kokoro/windows/build.bat
```diff
@@ -48,27 +49,47 @@ cmake -DRE2_BUILD_TESTING=OFF -DCMAKE_C_COMPILER="C:/Program Files (x86)/Microso
...
 
 :: #########################################
 :: Run the tests.
 :: #########################################
 echo "Running Tests... %DATE% %TIME%"
-ctest -C %BUILD_TYPE% --output-on-failure -j4
+::ctest -C %BUILD_TYPE% --output-on-failure -j4
 echo "Tests Completed %DATE% %TIME%"
 
+pwd
```
dneto0: So I guess I should stop paying attention to this PR for a while..... :-)

ehsannas: Yes, please ignore all changes to this PR until we can figure out why we're not able to cleanup and the build is done. Sadly there's no easy way to figure things out except by making a change and looking at the logs.

PR #440, opened 2018-03-06
Glslang generator version 5

dneto0:
Make OpArrayLength result type be an integer with signedness of 0.

### DIFF
--- a/glslc/test/assembly.py
+++ b/glslc/test/assembly.py
@@ -21,7 +21,7 @@ def assembly_comments():
     return """
     ; SPIR-V
     ; Version: 1.0
-    ; Generator: Google Shaderc over Glslang; 4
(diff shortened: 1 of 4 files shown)


dneto0:
This is required for refreshing google/glslang from upstream.
Reacts to Glslang version bump from 3 days ago: 
https://github.com/KhronosGroup/glslang/commit/0216f24f0e34ab57866c1286dbd52fd8d26492cb

dneto0:
>  But I'm wondering whether we should just drop testing the generator number, or use regex to make it more resilient to bumps. Does it harm anything if we drop testing the SPIR-V header completely?

Entirely fair.   The generator bump didn't happen for 2 years, then we've had a flurry of them.
Until now I've wanted to check the header in detail out of an abundance of caution.

PR #447, opened 2018-03-22
Use standard SPIRV-Tools transformation recipes

antiagainst:
Previously we manually manage the list of passes.
This commit converts to use the SPIRV-Tools standard
ones.

### DIFF
--- a/libshaderc/include/shaderc/shaderc.h
+++ b/libshaderc/include/shaderc/shaderc.h
@@ -122,9 +122,9 @@ typedef enum {
 
 // Optimization level.
 typedef enum {
-  shaderc_optimization_level_zero,         // no optimization
-  shaderc_optimization_level_performance,  // optimize towards performance
+  shaderc_optimization_level_zero,  // no optimization
   shaderc_optimization_level_size,  // optimize towards reducing code size
+  shaderc_optimization_level_performance,  // optimize towards performance
 } shaderc_optimization_level;
 
 // Resource limits.
--- a/libshaderc/include/shaderc/shaderc.h
+++ b/libshaderc/include/shaderc/shaderc.h
@@ -122,7 +122,8 @@ typedef enum {
 
 // Optimization level.
 typedef enum {
-  shaderc_optimization_level_zero,  // no optimization
+  shaderc_optimization_level_zero,         // no optimization
+  shaderc_optimization_level_performance,  // optimize towards performance
   shaderc_optimization_level_size,  // optimize towards reducing code size
 } shaderc_optimization_level;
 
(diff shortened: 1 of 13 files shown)


antiagainst:
I haven't added the API parameters for invoking performance passes. I'm not sure what we should call it. `shaderc_optimization_level_one`, `shaderc_optimization_level_three`, `shaderc_optimization_level_performance`?

https://github.com/google/shaderc/blob/c9931580dbf84657796e83e61bb682df51fd72ee/libshaderc/include/shaderc/shaderc.h#L124-L127

antiagainst:
Anyway, that can happen in a following up PR. This PR is basically internal clean up.

ehsannas:
Looks like the bots ran before my fixes went in.

antiagainst:
Okay, added `shader_optimization_level_performance` into the API and `-O` for `glslc`.

chrisvarns:
Just tried this out, FYI `shaderc_compile_options_set_optimization_level` in `libshaderc/src/shaderc.cc` is missing the switch for OptimizationLevel::Performance atm.

antiagainst:
@chrisvarns: Good catch! Thanks for pointing out! It turned out to be that we always run legalization passes for HLSL source code so my tests, using HLSL, does not reveal the problem previously. Changed to GLSL to be clear.

libshaderc/include/shaderc/shaderc.h
```diff
@@ -122,7 +122,8 @@ typedef enum {
 
 // Optimization level.
 typedef enum {
-  shaderc_optimization_level_zero,  // no optimization
+  shaderc_optimization_level_zero,         // no optimization
+  shaderc_optimization_level_performance,  // optimize towards performance
```
dneto0: Don't insert a value in between two existing values.  That breaks compatibility with anybody who has code that is compiled against the old header. Their code might ask for size optimization and now they would end up getting performance.

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -128,8 +128,9 @@ class Compiler {
 
   // Supported optimization levels.
   enum class OptimizationLevel {
-    Zero,  // No optimization.
-    Size,  // Optimization towards reducing code size.
+    Zero,         // No optimization.
+    Performance,  // Optimization towards better performance.
```
dneto0: Keep these in same order as the C header.    So reorder to match requested reorder there.

libshaderc/include/shaderc/shaderc.h
antiagainst: Urgh. Yes. I don't know what I was doing...

libshaderc_util/include/libshaderc_util/compiler.h
antiagainst: Fixed.

Issue #449, opened 2018-03-23
Windows dll's don't get installed


chrisvarns:
If I attempt to "install" with `ninja install`, the libs get copied but not the dll's.

dneto0:
shaderc_shared is listed as an install target: https://github.com/google/shaderc/blob/master/libshaderc/CMakeLists.txt#L33 

I'm less experienced with CMake on Windows DLLs.  He @AWoloszyn  or @antiagainst can you take a look?

AWoloszyn:
We should also be adding
```
RUNTIME DESTINATOIN ${CMAKE_INSTALL_BINDIR}
```
To that line.
The .dll is neither a Library or an Archive.
I can get a MR together to fix this. Hopefully that doesn't break the other platforms.

dneto0:
Indeed, install directives inside SPIRV-Tools have that line.  Thanks!

AWoloszyn:
Fixed by #451

PR #451, opened 2018-04-12
Also install shaderc_shared.dll on windows.

AWoloszyn:
We were not installing shaderc_shared.dll on windows during
install, only the shaderc_shared.lib link library.


AWoloszyn:
Fixes https://github.com/google/shaderc/issues/449

PR #452, opened 2018-04-13
Support -fhlsl_functionality1

dneto0:

PR #453, opened 2018-04-17
Document that OpenGL compatibility profile results in error

dneto0:


dneto0:
Superceded by https://github.com/google/shaderc/pull/452

Issue #457, opened 2018-05-06
Windows: size mismatch between file output and console output


Nimbal:
When using the option "-o -" to output the compilation result to std out, the console output is a few bytes larger than the file output. I noticed because Vulkan doesn't like SPIR-V code that is not a multiple of 4.

System specs:

- Windows 10 64 bit
- Shell: cmd.exe
- shaderc v2018.0-dev beginning-464-g5fd1b25
- spirv-tools v2018.3-dev v2018.2-83-ge7c2e91
- glslang 6.2.2596-110-g3bb4c48c

This happens with every shader I've tried. In the attached [zip file](https://github.com/google/shaderc/files/1977598/shader_size.zip), you can find a `shader.vert` file and two SPIR-V files. The output file `ok.spv` was generated with `glslc -o ok.spv shader.vert` and is 996 bytes. The file `broken.spv` was generated with `glslc -o - shader.vert > broken.spv` and is 1001 bytes.

After some investigation, it seems that something between compilation and writing it out to console replaces all "LF" characters (0x0A) with the Windows newline "CRLF" (0x0D, 0x0A). Removing the CR characters makes the shader work again.

dneto0:
The issue is at https://github.com/google/shaderc/blob/master/libshaderc_util/src/io.cc#L95 
The internal IO abstraction turns on binary mode exactly when the file name is not "-".

In other words, it was never anticipated that standard output would be used for binaries, only for things like preprocessed output.

PR #458, opened 2018-05-09
Force binary mode when writing a binary to stdout

dneto0:
For the test: also refactor the glslc test expectation code to
be able to check the length and header of a binary residing in a
Python "bytes" object, not just in a file.

Fixes https://github.com/google/shaderc/issues/457


dneto0:
I can reproduce the failure on Windows. Oddly, I haven't yet reproduced the failure on Linux.

dneto0:
Rebased.  Will watch the bots.  I expect only the pip issue on OSX

antiagainst:
Only pip issue remaining. I'll merge this.

PR #460, opened 2018-05-10
Link SPIRV-Tools into Glslang

### LINKED ISSUE #459: Glslang tests fail because its HLSL legalization is disabled
Commit https://github.com/google/shaderc/commit/2a0f3a3e348072ad6a9e53874188146b895705ba changed the order of project inclusion so that Glslang comes before SPIRV-Tools.  The idea was to have Shaderc control the optimization recipes rather than the ones in Glslang.

However, that means Glslang's self-tests fail because the "compoundsuffix.frag.hlsl" compilation no longer does HLSL legalization.  That causes a difference in output, which fails the tests.

So we have to reorder the includes, but always force compilation via Shaderc to *disable* HLSL legalization. 

Assigning to @antiagainst to fix.

This is blocking all future changes, such as #458

antiagainst:
Fixes #459

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -32,6 +32,19 @@ if (IS_DIRECTORY ${SHADERC_SPIRV_HEADERS_DIR})
   add_subdirectory(${SHADERC_SPIRV_HEADERS_DIR} spirv-headers)
 endif()
 
+# Check SPIRV-Tools before glslang so that it is linked into glslang.
+# we control optimizations via glslang API calls directly.
+if (IS_DIRECTORY ${SHADERC_SPIRV_TOOLS_DIR})
+  if ("${SHADERC_SKIP_TESTS}")
+    # Also skip building tests in SPIRV-Tools.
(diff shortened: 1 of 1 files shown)


dneto0:
The first failures I found were due to a SPIRV-Headers update that changed capability dependencies.
https://github.com/KhronosGroup/SPIRV-Headers/commit/3a9bf0b12a5b1a0a4996b0066f1451345c293546 

This caused failures in SPIRV-Tools validation tests, since they check some of those dependencies.
Situation normal. I'll fix SPIRV-Tools tests.

antiagainst:
Bots are failing because of an inconsistency between SPIRV-Headers & SPIRV-Tools. I've uploaded a fix for it: https://github.com/KhronosGroup/SPIRV-Tools/pull/1555.

dneto0:
SPIRV-Tools has been updated.  Please amend and update the branch so that the bots restart.

antiagainst:
Force pushed a new commit.

antiagainst:
Okay. The remaining failure is the pip issue on macOS. I'll give another attempt to fix it in a following up pull request.

PR #461, opened 2018-05-11
Travis: download and install pip on macOS

antiagainst:


.travis.yml
```diff
@@ -56,11 +56,12 @@ before_install:
       git clone --depth=1 https://github.com/taka-no-me/android-cmake.git $HOME/android-cmake;
       export TOOLCHAIN_PATH=$HOME/android-cmake/android.toolchain.cmake;
     fi
-
-install:
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
-      sudo easy_install pip;
+      curl https://bootstrap.pypa.io/get-pip.py -o /tmp/get-pip.py;
```
dneto0: Do we know that /tmp from different jobs are isolated so that this will never collide?
It would be better if there were a TMP environment variable provided by Travis, but it seems not:
https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables 

I would rather get that file in the current directory:
```
curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py;
sudo python get-pip.py;
rm -f get-pip.py
```
I presume that the current directory is : always isolated from other builds, and writable.
Or

antiagainst: We are requiring sudo. IIRC, for sudo environments, we get full VM: https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

So /tmp should be isolated.

I tried to use the current directory for the project, then `add_copyright.py` checks yelled at me because `get-pip.py` does not have a copyright header. :(

dneto0:
Rebased and pushed into master as  be8e08797

PR #463, opened 2018-05-29
Fix the link order of libglslang and libHLSL

haasn:
libglslang depends on libHLSL, so the latter needs to be specified last.
This fixes an issue when trying to build shaderc against system-wide
versions of libglslang/libHLSL, rather than the in-tree versions from
third_party.

### DIFF
--- a/glslc/CMakeLists.txt
+++ b/glslc/CMakeLists.txt
@@ -18,7 +18,7 @@ add_library(glslc STATIC
 shaderc_default_compile_options(glslc)
 target_include_directories(glslc PUBLIC ${glslang_SOURCE_DIR})
 target_link_libraries(glslc PRIVATE glslang OSDependent OGLCompiler
-  HLSL glslang SPIRV ${CMAKE_THREAD_LIBS_INIT})
+  glslang SPIRV HLSL ${CMAKE_THREAD_LIBS_INIT})
 target_link_libraries(glslc PRIVATE shaderc_util shaderc)
 
 add_executable(glslc_exe src/main.cc)
--- a/libshaderc_util/CMakeLists.txt
+++ b/libshaderc_util/CMakeLists.txt
@@ -34,8 +34,8 @@ endif(SHADERC_ENABLE_INSTALL)
 
 find_package(Threads)
 target_link_libraries(shaderc_util PRIVATE
-  glslang OSDependent OGLCompiler HLSL glslang SPIRV
-  SPIRV-Tools-opt ${CMAKE_THREAD_LIBS_INIT})
+  glslang OSDependent OGLCompiler glslang HLSL SPIRV
+  SPIRV-Tools-opt SPIRV-Tools ${CMAKE_THREAD_LIBS_INIT})
 
 shaderc_add_tests(
   TEST_PREFIX shaderc_util


haasn:
Additionally, I found another error in the linker scripts: `libshaderc_util` also needs to depend on `SPIRV-Tools`. I've included this fix in the commit.

glslc/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@ add_library(glslc STATIC
 shaderc_default_compile_options(glslc)
 target_include_directories(glslc PUBLIC ${glslang_SOURCE_DIR})
 target_link_libraries(glslc PRIVATE glslang OSDependent OGLCompiler
-  HLSL glslang SPIRV ${CMAKE_THREAD_LIBS_INIT})
+  glslang SPIRV HLSL ${CMAKE_THREAD_LIBS_INIT})
```
dneto0: glslang was already listed before and after HLSL.
I don't know the details but that's usually necessary when there's a circular dependency between the two.   Are you sure that HLSL itself doesn't depend on glslang?

We may require a *second* mention of HLSL, something like:
 
    glslang ... HLSL .. glslang .. HLSL 
?

libshaderc_util/CMakeLists.txt
```diff
@@ -34,8 +34,8 @@ endif(SHADERC_ENABLE_INSTALL)
 
 find_package(Threads)
 target_link_libraries(shaderc_util PRIVATE
-  glslang OSDependent OGLCompiler HLSL glslang SPIRV
-  SPIRV-Tools-opt ${CMAKE_THREAD_LIBS_INIT})
+  glslang OSDependent OGLCompiler glslang HLSL SPIRV
+  SPIRV-Tools-opt SPIRV-Tools ${CMAKE_THREAD_LIBS_INIT})
```
dneto0: CMake already has a "PUBLIC" dependency from SPIRV-Tools-opt to SPIRV-Tools.  So it seems this is not necessary?

What error are you trying to fix here?

dneto0:
Please rebase your changes against latest master to rerun the bots.

Also, please sign the contributor's license agreement (CLA).  That's necessary before we can accept your CL.

glslc/CMakeLists.txt
haasn: It seems to work as-is for me already, but I suppose I can add the extra link to be on the safe side.

libshaderc_util/CMakeLists.txt
haasn: Correct me if I'm wrong, but those dependencies only do anything if you actually build SPIRV-Tools-opt / SPIRV-Tools using your own cmake scripts, from third_party. But in practice, I am not building libshaderc against the third_party versions - I am building them against the system-wide SPIRV-Tools installation. (As the commit message points out)

The concrete issue being solved here is https://github.com/sjnewbury/gentoo-gpu/issues/22, which prevents `libshaderc` from working on Gentoo. The gentoo packaging policy forbids the use of vendored `third_party`-style libraries in favor of system-wide versions of those libraries wherever avoidable, and this is such a case.

I can either ship the work-around from this commit as part of the gentoo package, or I can try submitting it upstream - which is what I tried here.

haasn:
> Also, please sign the contributor's license agreement (CLA). That's necessary before we can accept your CL.

If I give you free license to take over authorship of the commit, does that mean I don't have to sign the CLA? I looked at the relevant links but was confused as to what exactly you want me to do. I can copy/paste some statement, but I'm not about to download and fill in some form for a trivial commit.

dneto0:
I'd like @AWoloszyn 's opinion on how to better support building against already-installed support libraries.

AWoloszyn:
> I'd like @AWoloszyn 's opinion on how to better support building against already-installed support libraries.

Sorry about not getting around to this earlier, it fell off the end of my inbox.
My understanding of how one should generally handle external dependencies for CMake, is with find_package. 

In order for us to handle the correct dependencies from SPIRV-Tools-opt -> SPIRV-Tools we will have to export a CMake config file that describes the transitive dep. See:
https://stackoverflow.com/questions/46861504/cmake-transitive-dependency-is-not-found-via-find-package

Then users would not be required to know this transitive dependency.

Ideally something like this should be done in Glslang as well, so that users do not have to replicate the strange circular dependency behavior.

haasn:
> My understanding of how one should generally handle external dependencies for CMake, is with find_package.

The better way to handle external dependencies in general would be with pkg-config

AWoloszyn:
Correct me if I am wrong, but pkg-config is not supported (properly) on Windows, so we would have to maintain a second setup for deps for Windows as well.

With the cmake-based solution at least it should work everywhere.

dneto0:
> With the cmake-based solution at least it should work everywhere.

Doing something like LunarG for FindVulkan seems the right way forward.

See https://github.com/Kitware/CMake/blob/master/Modules/FindVulkan.cmake 
And https://cmake.org/cmake/help/v3.7/module/FindVulkan.html

dneto0:
this has gone stale.  closing.

PR #469, opened 2018-06-13
Add virtual dtor to classes with virtual functions

ProgramMax:
IncluderInterface has virtual functions but does not have a virtual
destructor. This class is derived from by FileIncluder w hich overrides
those functions.

Because there is an interface in use here, it is safe to assume some
container is storing IncluderInterface*. If the container instead held
FileIncluder* then the virtual functions wouldn't be needed.

This causes a problem since FileIncluder has member variables. The
destructor of FileIncluder knows to also destruct those member variables
(including deallocating their dynamically allocated memory).

But when IncluderInterface's destructor is called, it is not virtual and
will not call FileIncluder's destructor. So these member variables are
never destroyed and their dynamically allocated memory will be leaked.

In this case, FileIncluder stores a std::unordered_set<std::string>
which will be leaked.

This patch adds a virtual destructor to IncluderInterface to make sure
FileIncluder's destructor is called and this memory isn't leaked.

### DIFF
--- a/glslc/src/file_includer.cc
+++ b/glslc/src/file_includer.cc
@@ -25,6 +25,9 @@ shaderc_include_result* MakeErrorIncludeResult(const char* message) {
   return new shaderc_include_result{"", 0, message, strlen(message)};
 }
 
+FileIncluder::~FileIncluder() {
+}
+
 shaderc_include_result* FileIncluder::GetInclude(
     const char* requested_source, shaderc_include_type include_type,
     const char* requesting_source, size_t) {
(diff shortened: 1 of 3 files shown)


ProgramMax:
Oops. I was working off of Chromium's copy, which is stale.

It appears this has been fixed. A virtual destructor has been added to IncluderInterface. However, I am unsure if we want to implicitly override in FileIncluder. Right now, someone who looks at FileIncluder may not know that there is a virtual destructor.

This pull request now makes FileIncluder's destructor explicitly defaulted.

ProgramMax:
I see you have approved the changes. I'm not sure if I am supposed to commit (and how) or if someone else does.

What are the next steps?

dneto0:
Thanks for your patience.
I was waiting for the bots to finish, then got pulled to other work. 

Squashed and pushed into master as 30af9f9899aefd01866

PR #476, opened 2018-08-13
Create intermediate directories for output files

Leandros:
I noticed that `glslc` doesn't create intermediate directories (like other compilers do). I've added a helper function to do exactly that and added it to create the directories for the output & dependency files.


Leandros:
I think I broke outputting files with no paths. Will fix it.

Leandros:
Fixed the tests, however, there are a couple tests which assume that not creating intermediate directories is the correct behavior.

PR #479, opened 2018-09-19
Add support for Nvidia turing extensions

chaoc:


chaoc:
I signed it!

dneto0:
The bots failed because Glslang needed an update for these extensionsl.
Those have landed now.  That update causes Shaderc tests to fail.

So we have to update google/glslang in any case.
I have manually verified these changes work with updated SPIRV-Headers and SPIRV-Tools and Glslang.

PR #484, opened 2018-09-22
Add new Glslang files to Android.mk

dneto0:


dneto0:
I checked this fix locally with a run of ndk-build.  The rest of the build is not affected by this change, so I did not wait for the bots.  Merged early to assist LunarG release processes.

PR #485, opened 2018-09-24
Add BUILD.gn rules.

dj2:
This CL adds the necessary build configuration to allow building
shaderc within a Chromium checkout.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,95 @@
+# Copyright 2018 The Shaderc Authors. All rights reserved.
+# 
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#     http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


dj2:
@Kangz to verify by BUILD.gn is right-ish. I was able to successfully do a build of libshaderc from within the Chromium tree with this and the glslang BUILD.gn file.

BUILD.gn
```diff
@@ -0,0 +1,95 @@
...
+  # Executable needs an rpath to find our shared libraries on OSX
+  if (is_mac) {
+    ldflags = [
+      "-rpath",
+      "@executable_path/",
+    ]
+  }
+}
+
+shared_library("libshaderc") {
```
Kangz: I think that for Chrome we would like to have a dual `source_set` and `shared_library` model. See libdawn_native here: https://dawn.googlesource.com/dawn/+/master/BUILD.gn#313

This would help put shaderc in a component in Chrome.

dj2: Done

Kangz:
Still LGTM!

dj2:
@antiagainst ping, are you the right person to review this?

PR #489, opened 2018-09-27
Set the CMake C++ standard to C++11

alan-baker:
Force C++11 for builds.


antiagainst:
Good enough. Appveyor is just doing duplicated checks.

Issue #475, opened 2018-08-11
Cannot optimize with target vulkan1.1


danielmaciel:
Attempting to optimize using  --target-env=vulkan1.1 -O fails with this message:
shaderc: internal error: compilation succeeded but failed to optimize: Invalid SPIR-V binary version 1.3 for target environment SPIR-V 1.0 (under Vulkan 1.0 semantics).

I have attached a patch. It gives you an idea of what needs to be done. It is not super great but gets the job done.

[vulkan11opt.patch.txt](https://github.com/google/shaderc/files/2280049/vulkan11opt.patch.txt)

dneto0:
Thanks for the bug report.

We can only accept patches via merge request.  We use GitHub's code review flow, and build bots to check functionality across tools and environments.

(I see that our CONTRIBUTING.md document does not mention this!)

antiagainst:
I've created https://github.com/google/shaderc/pull/491, which should address this issue.

Selmar:
I've just built using the latest, as I had the same problem, but I still have it.

I saw the changes were only in libshaderutil, so I've simply set the third-party settings like this:
![image](https://user-images.githubusercontent.com/3010534/47260846-17ee3b00-d4c3-11e8-9c0d-2f2c1feef0f6.png)

PR #491, opened 2018-09-28
Pass environment version into SPIRV-Tools properly

antiagainst:
Otherwise the assembling/disassembling/optimizing may error out
because of target environment version mismatch.

Fixes https://github.com/google/shaderc/issues/475


dneto0:
Rebased and pushed into master as fc4ff0e56

PR #495, opened 2018-10-02
BUILD.gn: only use complete_static_lib on mac

Kangz:
PTAL @dneto0 @dj2


dneto0:
Rebased and pushed onto master as c9da865

PR #497, opened 2018-10-03
BUILD.gn: Avoid source_set depending on static_libraries

Kangz:
PTAL @dj2 @dneto0 this should be the last one of these PRs for now (at least to roll into Dawn, Chrome might need more)


Kangz:
Are you running tests with GN on kokoro? If not I'm not sure what could be causing the failure.

dneto0:
Test failures are due to https://github.com/KhronosGroup/glslang/issues/1521

PR #498, opened 2018-10-03
Use SHADERC_BUILD_SHARED to control static/shared library

antiagainst:
Previously we build both static and shared library at the same
time. Because of that, we cannot name both the static and shared
library as shaderc.

This commit stops compiling both static and shared library.
Instead we let the user choose now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,15 @@ cmake_minimum_required(VERSION 2.8.12)
 project(shaderc)
 enable_testing()
 
+set (CMAKE_CXX_STANDARD 11)
+
 message(STATUS "Shaderc: build type is \"${CMAKE_BUILD_TYPE}\".")
+
+# Note that we are not using or setting CMake BUILD_SHARED_LIBS here since that
+# also affects external projects and makes libshaderc.so to depend on shared
+# libraries compiled from external projects. We want libshaderc.so to depend
+# on only standard system libraries.
+option(SHADERC_BUILD_SHARED "Build Shaderc as shared library" OFF)
 
 option(SHADERC_SKIP_INSTALL "Skip installation" ${SHADERC_SKIP_INSTALL})
 if(NOT ${SHADERC_SKIP_INSTALL})
@@ -18,8 +26,6 @@ if(${SHADERC_ENABLE_TESTS})
 else()
   message(STATUS "Configuring Shaderc to avoid building tests.")
 endif()
-
-set (CMAKE_CXX_STANDARD 11)
 
 option(SHADERC_ENABLE_NV_EXTENSIONS "Enables support of Nvidia-specific extensions" ON)
 
--- a/libshaderc/CMakeLists.txt
+++ b/libshaderc/CMakeLists.txt
@@ -10,18 +10,19 @@ set(SHADERC_SOURCES
   src/shaderc_private.h
 )
 
-add_library(shaderc STATIC ${SHADERC_SOURCES})
+if (${SHADERC_BUILD_SHARED})
+  add_library(shaderc SHARED ${SHADERC_SOURCES})
+  target_compile_definitions(shaderc
+      PRIVATE SHADERC_IMPLEMENTATION
+      PUBLIC SHADERC_SHAREDLIB
+  )
+  set_target_properties(shaderc PROPERTIES SOVERSION 1)
+else()
+  add_library(shaderc STATIC ${SHADERC_SOURCES})
+endif (${SHADERC_BUILD_SHARED})
+
 shaderc_default_compile_options(shaderc)
 target_include_directories(shaderc PUBLIC include PRIVATE ${glslang_SOURCE_DIR})
-
-add_library(shaderc_shared SHARED ${SHADERC_SOURCES})
-shaderc_default_compile_options(shaderc_shared)
-target_include_directories(shaderc_shared PUBLIC include PRIVATE ${glslang_SOURCE_DIR})
-target_compile_definitions(shaderc_shared
-    PRIVATE SHADERC_IMPLEMENTATION
-    PUBLIC SHADERC_SHAREDLIB
-)
-set_target_properties(shaderc_shared PROPERTIES SOVERSION 1)
 
 if(SHADERC_ENABLE_INSTALL)
   install(
@@ -31,7 +32,7 @@ if(SHADERC_ENABLE_INSTALL)
     DESTINATION
       ${CMAKE_INSTALL_INCLUDEDIR}/shaderc)
 
-  install(TARGETS shaderc shaderc_shared
+  install(TARGETS shaderc
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
@@ -46,7 +47,6 @@ set(SHADERC_LIBS
 )
 
 target_link_libraries(shaderc PRIVATE ${SHADERC_LIBS})
-target_link_libraries(shaderc_shared PRIVATE ${SHADERC_LIBS})
 
 shaderc_add_tests(
   TEST_PREFIX shaderc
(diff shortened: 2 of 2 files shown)


dneto0:
This will break downstream projects?

antiagainst:
You mean renaming the shared library name or not building static and shared libraries together?

dneto0:
I mean renaming the library name.

antiagainst:
I checked the latest Vulkan SDK, only shaderc_combined.a is shipped. So won't be a problem for users replying on the SDK. Then for others that built Shaderc manually, I think they need to periodically accommodate the changes from Shaderc anyway?

We can create symbolic links if necessary.

antiagainst:
@dneto0: thoughts?

jeeb:
As a 3rd party stand-byer if this helps with the fact that:
1. A [project](https://github.com/mpv-player/mpv/blob/317d3ac26634d6719c2c70e3e88acbf0c108fd73/wscript#L732..L743) currently has to be on the look-out for shared and static shaderc with separate names.
2. Now there can be just a **single** .pc file instead of two (one for _shaderc-static_, one for _shaderc-shared_).

I am all for this.

haasn:
> Now there can be just a single .pc file instead of two (one for shaderc-static, one for shaderc-shared).

(disregarding that shaderc does not have a .pc file yet anyway)

eli-schwartz:
CMake has this builtin. Just don't specify whether it is STATIC or SHARED, and use the toggle described in https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html to allow CMake to automatically choose the preferred library type.

In addition to being less work, this means there is a common interface for this setting that people can utilize across many different CMake projects. You can add it as a default-off `option()` to make it user-visible, though.

Then just check the current value of it to determine whether to do the `*_combined` stuff.

antiagainst:
@eli-schwartz: As I commented in the source code:

```
# Note that we are not using or setting CMake BUILD_SHARED_LIBS here since that
# also affects external projects and makes libshaderc.so to depend on shared
# libraries compiled from external projects. We want libshaderc.so to depend
# on only standard system libraries.
```

eli-schwartz:
If a user requests to opt in to `BUILD_SHARED_LIBS`, why are you ignoring their strongly-worded request to use shared libs?

Also what happens if you statically compile libshaderc.so against symbols from spirv-tools or whatever, and then try to use the same symbols in a thirdparty program that links to both spirv-tools and libshaderc?

Attempting to compile a *shared library* using *static dependencies* seems rather silly.

- If you asked for non-default shared builds, then the logical assumption is you're trying to integrate into your system, for example, you're (a member of) a Linux distribution. (I'm a member of a linux distribution.) If https://github.com/google/shaderc is a project that *doesn't integrate into your system*, you are still going to have to delete the line https://github.com/google/shaderc/blob/eb743ec9cb31303e0abefeeb74886ab134dced6f/CMakeLists.txt#L64 and `rm -rf third_party`.

- If you're trying to avoid depending on system libraries, you are going to use a static library for libshaderc too, because you don't want to worry about installing two files. You want it to Just Work, and that means statically compiling libshaderc into your application.

dj2:
You don't need to remove the third_party, you can override where shaderc gets third_party from so it will use the same spirv-tools as you. This is done in https://github.com/google/amber/blob/master/third_party/CMakeLists.txt#L179 to share the third_party files.

dj2:
In which case, we may just want to add something like a USE_SYSTEM_LIBS flag to shaderc which avoids using its own third_party.

haasn:
> In which case, we may just want to add something like a USE_SYSTEM_LIBS flag to shaderc which avoids using its own third_party.

That would be a good idea. Note that [gentoo also opts to remove the third party entry](https://github.com/gentoo/gentoo/blob/master/media-libs/shaderc/shaderc-2017.2.ebuild#L48). In general, no distro is going compile shaderc against its vendored dependencies. You could make many distro maintainers' livee easier by just supporting this directly.

Incidentally, that means more people will hit #463. So you might want to start taking that PR seriously if you plan on allowing users to build against non-bundled libs. (And as expected, [gentoo also has to include this work-around](https://github.com/gentoo/gentoo/blob/master/media-libs/shaderc/files/shaderc-2017.2-fix-glslang-link-order.patch))

eli-schwartz:
While such an option would be useful, I still don't understand what the use case is for building a shared library and linking it against vendored static dependencies. It would make far more sense to have one option that toggled whether to build a fully static shaderc, or a fully shared shaderc.

The whole **point** of shared libraries is in order to share resources between many different projects. The shared library **format** is not much use if you don't, well, share it. I'm genuinely curious why someone afraid of relying on system copies of spirv-tools, glslang etc. would want to rely on system copies of shaderc either.

Seriously -- why does the entire computing world have such a hate-on for statically linking their applications? If your shared library is not tightly integrated into your operating system, it makes zero sense to use it at all. Static libraries are wonderful. They're the one and only, superior way to build programs that don't depend on fiddly external resources floating around outside your control.

Yes, this is me with my Linux distribution developer hat on.

antiagainst:
Hey @eli-schwartz:

Thanks a lot for your inputs on this!

> Attempting to compile a shared library using static dependencies seems rather silly.

Yep, I'm not a Linux distribution developer, but my intent was trying to help. This pull request is not landed, so it's just one possible way of achieving the purpose. However, I'm quite happy to explain why I did it this way. :-)

The reason is that I thought Glsang & SPIRV-Tools didn't have their shared library story sorted out yet; they have not concretized on their versioning schemes, etc. (It may be untrue, or it may have already changed now, or maybe it does not matter; you can correct me on this.) If we just compile multiple `.so`s and let it be, it may introduce problems regarding compatibility of different components. (We've seen quite a few failures because Glslang & SPIRV-Tools out of sync, e.g., using different different versions of SPIRV-Headers.) So I was trying to hide all that complexity behind and just provide a single `.so` and wait until all the subprojects sort their `.so` stories out. You can always convert to multiple `.so`s later and this provides a pragmatic way to ship asap I think?

Besides, this is the way `libshaderc_shared.so` used to work, which I think is proven to work for some users (Chrome?).

As for the point of using shared library to share things, it's relative, right? Even with shared libraries, you still compile a bunch of symbols into the same `.so`, not every component gets their dedicated `.so`. So if more than one binary link to `libshaderc.so`, they still share the same `libshaderc.so`.

Hope this helps. :-)

haasn:
While this is sort of off-topic, the main point of using shared libraries, to me, has always been the ability to make updates to shared libraries (as long as they don't break ABI) without needing to touch all of the dependents.

Consider the use case of something like `libjpeg.so`. If there is a critical security vulnerability in libpng, all it requires to fix it is to drop in the new .so (with the same ABI) and restart the affected programs. In an entirely statically linked system, I also need to recompile everything on the system that was linked against libjpeg - and unlike with dynamic linking, there's no easy way to find out which programs this affects. Plus, I may not even have the source code for the components I am trying to relink. With a shared approach, I can *still* swap out what it links against in this instance.

It's not just security vulnerabilities - if I want to change the behavior of a dependency, I can also easily do that in a way that affects the whole system. This is often the case for me as a developer. To give a silly but motivating example, I compile libjpeg.so with arithmetic decoding turned on. This means that everything which links against libjpeg.so dynamically is able to decode arithmetic JPEGs. But if a browser, say firefox, comes along and decides to use its own vendored libjpeg copy instead, it will not benefit from this change. Not a big deal in this instance but it can be annoying for more meaningful changes - especially if you're trying to test a patch that could potentially fix a problem you found in such a shared component.

Now, I realize that this reasoning does not apply that well to shaderc's vendored third party dependencies, because glslang & friends cannot be dynamically linked anyway. (They are only available as a weird collection of .a files). And that's probably a good thing, since I don't trust the glslang developers to bump their ABI version every time they break the ABI. (Heck, [they don't even bump their version headers when they make breaking changes to the API](https://github.com/KhronosGroup/glslang/issues/1538)). Actually, I don't really trust shaderc to do this either (and shaderc does not even *have* version headers, which makes it annoying to write forwards- and backwards-compatible code). But that's beside the point - the point is that even though glslang is linked against statically by shaderc, I **still want my distribution to control which version of glslang is on my system**. Imagine wanting to test out a new patch or a new version that could fix a problem in glslang. Right now, it's as simple as installing the new version of glslang and then re-installing shaderc. (The latter can even happen automatically). But if gentoo had not put in the effort to strip out the third_party from shaderc's build tools, I would first have to edit the shaderc build script to somehow point it at the new version of glslang.

Anyway, the bottom line is that (most) distributions **want one component per package**. No bundling three unrelated vendored dependencies inside the same package, that's simply not kosher. It makes things difficult to control for the user/maintainer, and unpredictable as to which version from which git repo is actually being used.

My 2 cents.

haasn:
> Linking to a private, vendored .so file means you don't get any of the benefits of shared libraries.

There is one case that still benefits: Say a closed-source video game links against a vendored, old/buggy version of a library like libsdl2.so (or, indeed, libshaderc.so). As a result of this bug, the game fails to work on my system. Thanks to dynamic linking, I can still fix the bug in the .so and swap it out without needing access to the video game's source code. Doing the same with a statically linked, stripped binary, is basically impossible.

antiagainst:
Thanks for all the great and detailed explanation! I see the points now.

I'm not married to this implementation; happy to drop this pull request if you guys think it's really troublesome. :-)

@Kangz: WDYT since you did the original shared library support?

Kangz:
No strong opinion, Dawn and Chrome and using the GN files now.

antiagainst:
Cool, I'll drop this pull request then. Thanks guys!

Issue #499, opened 2018-10-09
Newer SPIRV-Tools breaks hlsl compilation


baldurk:
SPIRV-Tools recently started running the validator on inputs before optimisation, see KhronosGroup/SPIRV-Tools#1799.

The validator runs with default options, which means that even with implied hlsl-offsets the validator uses glsl rules which disallows struct layouts like `cbuffer foo { float a; float3 b; }`.

The fix I think is in `SpirvToolsOptimize`, to add an explicit `spvtools::OptimizerOptions`. That would either let you configure validation to match the environment, or better yet just disable the validator. IMO the validator should **not** run by default on an internal pass since that's going to slow down compilation for everyone in a place the "shouldn't" have validation errors. Especially since it seems it was only added for the benefit of some fuzzing process that could opt-in to run validation itself before passing to spirv-opt.

The same problem happen with GLSL with opt-in hlsl-offsets - the option_dash_fhlsl_offsets.py test would hit this if it enabled optimisation, hlsl only hits it because of the required legalisation pass.

This affects the glslc in SDK 1.1.85.0.

dj2:
The spv_optimizer_options has a method : spvOptimizerOptionsSetRunValidator where you can disable the validator when running the optimizer. You can also set the validator options by calling: spvOptimizerOptionsSetValidatorOptions.

Just note, the optimizer isn't safe to run if there is invalid SPIRV. It will assert.

dneto0:
I agree the compiler should work out-of-the-box.  For HLSL compilation validation should be turned off or tuned to avoid the things we know the legalization steps will be working around.  When offsets are requested to be relaxed, those validation rules should be tuned down as well.

PR #500, opened 2018-10-09
Disable layout and relax logical pointer for optimizer validation

antiagainst:
These two validation exceptions are needed for HLSL compilation.

Fixes https://github.com/google/shaderc/issues/499

### DIFF
--- a/libshaderc/src/common_shaders_for_test.h
+++ b/libshaderc/src/common_shaders_for_test.h
@@ -320,6 +320,26 @@ const char kHlslWaveActiveSumeComputeShader[] =
        MyBuffer[id.x].result = WaveActiveSum(MyBuffer[id.x].val);
      })";
 
+const char kHlslMemLayoutResourceSelect[] =
+    R"(cbuffer Foo { float a; float3 b; }
+
+       Texture2D Tex;
+       SamplerState Sampler1;
+       SamplerState Sampler2;
+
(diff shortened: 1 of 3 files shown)


dneto0:
Bot failures:
 - macos-clang-debug is https://github.com/KhronosGroup/glslang/issues/1521  and https://github.com/KhronosGroup/glslang/issues/1522
- VisualStudio failures are googletest configuration problems? 

CMake Error:
  Error evaluating generator expression:
    $<IF:$<CONFIG:Debug>,gmockd,gmock>
  Expression did not evaluate to a known generator expression

antiagainst:
I cannot reproduce the failure too.

I suspect it's because https://github.com/google/googletest/commit/f5260ae757a681566d8f7d0558b8dc0551036506

dneto0:
I've confirmed the error is with that f5260ae commit

dneto0:
The $<IF: ... > syntax was introduced in CMake 3.8, but the Kokoro bots use CMake 3.7.
Ubuntu 18.04 LTS uses CMake 3.9

dneto0:
See https://cmake.org/cmake/help/v3.8/manual/cmake-generator-expressions.7.html

dneto0:
Ok.  #502 will avoid the googletest problem. That's independent.

PR #502, opened 2018-10-09
VS builds: Pin a older googletest to avoid around a CMake problem

dneto0:


dneto0:
Let's make sure the VS2017 build at least gets to running tests.

dneto0:
I filed https://github.com/google/googletest/issues/1892 to fix upstream

dneto0:
The VS2013 build had completed successfully.

PR #504, opened 2018-10-11
Unpin googletest. It has been fixed upstream

dneto0:


dneto0:
Rebased

dneto0:
VS2013 is expected to fail since googletest doesn't support it any more.
See also https://github.com/google/shaderc/issues/511 to disable that flow in our bots.

PR #510, opened 2018-11-12
Add effcee and re2 when building spirv-tools tests

dj2:
The effcee library was made a requirement for building the SPIRV-Tools test suite. When building that suite through shaderc we need to make sure the required dependencies are available.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,10 @@ build-*/
 compile_commands.json
 .ycm_extra_conf.py
 cscope.*
+third_party/effcee/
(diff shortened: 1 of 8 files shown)


dneto0:
I suspect the glslang tests will be fixed if we refresh google/glslang from upstream. I did that locally and it passes all tests.

dneto0:
I have refreshed google/glslang from KhronosGroup/glslang

dj2:
Failure looks unrelated, merging.

PR #513, opened 2018-11-15
Fixup shadowed variables

dj2:


dj2:
The failures look like unrelated glslang test failures.

PR #514, opened 2018-11-16
Build file updates.

dj2:
This CL adds a default to shaderc to build in Debug mode if the
build type is not provided. It also sets SPIRV-Headers to not build the
examples folder.


dj2:
Note, this CL depends on https://github.com/KhronosGroup/SPIRV-Headers/pull/82 landing first.

PR #515, opened 2018-11-19
Switch GLSLang to upstream

dj2:
This CL switches shaderc to use the KhronosGroup glslang instead of the
google glslang.


dneto0:
Thanks for revisiting this.  The interface between Shaderc and Glslang has been stable for a long time.  So the delayed integration has had more cost than benefit.

Now that Glslang is including SPIRV-Tools validation output, we end up being dependent on rapid updates to  our snapshot of Glslang anyway.  So the costs have become higher still.

I agree with this change in principle.  I'd like this MR to be rebased once the integration with Glslang top-of-tree goes green again; there's a bit of motion at the moment with return type of ArrayLength and possibly other things.  I expect to be able to land this within a week.

dj2:
We shouldn't be building tests on 2013 as googletest doesn't compile. For 2017, it looks like effcee and spriv-tools failures in the unit tests? (Other fault, whatever that means).

PR #518, opened 2018-11-21
Add support for NV_EXTENSIONS in #pragma shader_stage()

zeux:
This adds missing support for raytracing and task/mesh shaders to
shader_stage() pragma so that you can build these shaders without
specifying the stage via command line arguments.


zeux:
I signed it!

dneto0:
The bot failures are due to validation changes that require a Glslang update.

PR #520, opened 2018-11-27
Update to use Android.mk from glslang

dj2:


dj2:
Note, this depends on https://github.com/KhronosGroup/glslang/pull/1606 and https://github.com/KhronosGroup/SPIRV-Tools/pull/2126

PR #528, opened 2018-12-17
Switch GLSLang to upstream

dneto0:
This CL switches shaderc to use the KhronosGroup glslang instead of the
google glslang.

For getting source files out, switch to deps file.

Update Dockerfile to match.


dneto0:
This is a rebase and squash version of #515

dj2:
This was landed in other PR.

PR #531, opened 2018-12-21
[kokoro] Stop running tests on VS 2013 builds.

ehsannas:


kokoro/windows/build.bat
```diff
@@ -89,13 +89,15 @@ ninja check-copyright
 if %ERRORLEVEL% GEQ 1 exit /b %ERRORLEVEL%
 echo "Build Completed %DATE% %TIME%"
 
-:: #########################################
-:: Run the tests.
-:: #########################################
+:: ################################################
+:: Run the tests (We no longer run tests on VS2013)
+:: ################################################
+if NOT %VS_VERSION% == 2013 (
```
fjhenigman: So we're still going to build 2013 but not test it?  Why not wipe out 2013 completely, say it's not supported, delete the .bat and .cfg files?   Why leave an untested configuration hanging around?

dneto0: Yes, would love to do that. 
We've had this discussion with others in the ecosystem, and have pushed for what you suggest.  We get strong pushback.  I won't rehash the arguments here.

This is the compromise.

ehsannas:
Ouch:
```
T:\src\github\shaderc\third_party\googletest\googlemock\include\gmock/internal/gmock-port.h(61):

fatal error C1189: #error : 
"At least Visual C++ 2015 (14.0) is required to compile Google Mock."
```

ehsannas:
So we can't even compile googletest itself in 2013 anymore

dneto0:
Right, we have to completely skip compiling any gtest / gmock code for VS2013.

fjhenigman:
> > oh you already committed? you guys work cowboy style here, eh? :)
> 
> We can mark approval before bots are done. We don't want any new failures though, so I'll push back if there's a new failure.
> 
> (And yes, there has been noise at the margin with things changing upstream breaking things. Hopefully we'll squash that out.)

I saw "committed" on my screen and misunderstood.  It was just a commit on this branch, not merged into master.  Will take me a while to get used to github.  Seems very inferior to gerritt.

ehsannas:
VS 2013 bot seems to be happy now: builds everything except tests.

Issue #455, opened 2018-04-20
Add support for generating debug information


antiagainst:
Glslang can output `OpSource` and `OpLine`s since https://github.com/KhronosGroup/glslang/commit/44d2728e3d45bd064a4e6bf23d9c1793dc0da839.

We should hook that functionality up with `set_generate_debug_info` and `-g`:

https://github.com/google/shaderc/blob/5fd1b256480d4c3547015371b08c24ffe8ddba15/libshaderc_util/src/compiler.cc#L323

antiagainst:
Thanks, @aejsmith !

PR #532, opened 2018-12-28
Generate source text debug info if requested

aejsmith:
Pass the necessary flags through to glslang to get it to emit OpSource containing the shader source text if it was requested in the options.

Implements issue #455.

### DIFF
--- a/libshaderc/src/shaderc_cpp_test.cc
+++ b/libshaderc/src/shaderc_cpp_test.cc
@@ -458,21 +458,29 @@ TEST_F(CppInterface, ForcedVersionProfileRedundantProfileStd) {
 
 TEST_F(CppInterface, GenerateDebugInfoBinary) {
   options_.SetGenerateDebugInfo();
-  // The output binary should contain the name of the vector: debug_info_sample
-  // as char array.
-  EXPECT_THAT(CompilationOutput(kMinimalDebugInfoShader,
-                                shaderc_glsl_vertex_shader, options_),
-              HasSubstr("debug_info_sample"));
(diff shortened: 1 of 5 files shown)


aejsmith:
CLA signed.

libshaderc/src/shaderc_cpp_test.cc
```diff
@@ -458,21 +458,29 @@ TEST_F(CppInterface, ForcedVersionProfileRedundantProfileStd) {
...
-  // as char array.
-  EXPECT_THAT(CompilationOutput(kMinimalDebugInfoShader,
-                                shaderc_glsl_vertex_shader, options_),
-              HasSubstr("debug_info_sample"));
+  const std::string binary_output =
+      CompilationOutput(kMinimalDebugInfoShader,
+                        shaderc_glsl_vertex_shader, options_);
+  // The binary output should contain the name of the vector (debug_info_sample)
+  // null-terminated, as well as the whole original source.
+  std::string vector_name("debug_info_sample");
```
dneto0: Using an escaped null character would have been clearer, e.g.   "debug_info_sample\0"

aejsmith: Just doing `std::string vector_name("debug_info_sample\0")` wouldn't end up including that null terminator in the comparison since the string constructor would just treat that as normal end of string (if it's not clear - here I'm trying to specifically catch the OpName for debug_info_sample as the test would have previously done, rather than just possibly matching what's in the OpSource, which could happen without checking for the null terminator).

PR #533, opened 2019-01-03
Fix Windows test script return value.

fjhenigman:
Using %ERRORLEVEL% inside an IF...() block captures the variable's
value as the block is entered.  Use !ERRORLEVEL! so error handling
works as expected.
Use %ERRORLEVEL% NEQ 0 in case of negative values.
Ensure zero is returned if tests passed, regardless of cleanup errors.
Refactor cmake flags to reduce duplication.

### DIFF
--- a/kokoro/windows/build.bat
+++ b/kokoro/windows/build.bat
@@ -56,42 +56,44 @@ if "%KOKORO_GITHUB_COMMIT%." == "." (
   set BUILD_SHA=%KOKORO_GITHUB_COMMIT%
 )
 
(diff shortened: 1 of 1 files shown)


fjhenigman:
The failure on vs2017-debug is what we want.  It was previously failing and not reporting the failure.

ehsannas:
@dj2 yes, since spirv-tools and amber mostly copied these scripts, they should be fixed too.
@alan-baker , I believe this is the issue you observed in spirv-tools a couple of weeks ago.

ehsannas:
I have opened https://github.com/KhronosGroup/SPIRV-Tools/pull/2262

PR #534, opened 2019-01-03
Fixup source matching test

dj2:
This CL rolls GLSLang and reverts the test change made to make MacOS pass as it has
now been fixed upstream in GLSLang.


dj2:
Attempting to test GLslang fix before committing ....

PR #537, opened 2019-01-08
Roll RE2 Dep

dj2:


dj2:
This is just a test, when we land it it should point to the official re2 repo.

dj2:
Done, rolled to google/re2 revision.

PR #538, opened 2019-01-11
Android.mk: better fallback for SPIRV-Tools dir

dneto0:
This fixes building with ndk-build when the source tree is laid
out like in the Android NDK.


dneto0:
This MR only affects Android flows, and they've passed.

dneto0:
This fix is needed for NDK update and Vulkan SDK update.

PR #541, opened 2019-01-15
Remove use of deprecated googletest macro

s-perron:
INSTANTIATE_TEST_CASE_P has been deprecated. We need to use
INSTANTIATE_TEST_SUITE_P instead.


fjhenigman:
A rebase will fix the ndk-build.  You're missing fcb2d4e2ff2b22147712babbb8612f5157431239

dneto0:
To be clear, this PR or one like it will have to wait until the NDK updates its snapshot of googletest.  I don't know if that is even scheduled.

s-perron:
There is a CL to update the NDK. It is waiting for review.

s-perron:
The Andriod NDK has been updated.  This should be good to go.

PR #544, opened 2019-01-16
Fix MinGW: Format %lu must be unsigned long.

dneto0:


dneto0:
> I just learned %zu is the way to printf a size_t, but it's not supported on older ms compilers.

Yes, I can confirm it fails on older MS compilers, from past experience.  :-)

dneto0:
Rebased.

Issue #542, opened 2019-01-16
spvc compilation should be off by default


dneto0:
It's early to be requiring all Shaderc clients to build the spvc flow.

For example, we're not yet ready to publish the spvc tool into the Android NDK.  Doing _that_ requires quite a bit of review before it's permitted.

So please disable the spvc compilation by default.  The requirement to have the SPIRV-Cross source tree should be off by default (via the same option).

Suggesting new CMake variable:  SHADERC_BUILD_SPVC   it should default to OFF.
When OFF, it the build should work if the SPIRV-Cross source tree is absent.

(For an example, see SPIRV_BUILD_COMPRESSION in SPIRV-Tools)

PR #547, opened 2019-01-17
Disable spvc build by default.  (#542)

fjhenigman:
Make SHADERC_ENABLE_SPVC default to OFF then turn it on for all the bots.


CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ endif()
 
 option(SHADERC_SKIP_SPVC "Skip libshaderc_spvc" ${SHADERC_SKIP_SPVC})
 if(NOT ${SHADERC_SKIP_SPVC})
-  set(SHADERC_ENABLE_SPVC ON)
+  set(SHADERC_ENABLE_SPVC OFF)
```
dneto0: This doesn't work for me in a tree without the spirv-cross sources.
If I give none of the options SHADERC_ENABLE_SPVC or SHADERC_SKIP_SPVC then it fails with error:

CMake Error at third_party/shaderc/third_party/CMakeLists.txt:91 (message):
  SPIRV-Cross was not found - required for compilation

Just negating the setting isn't right.

PR #548, opened 2019-01-18
BUILD.gn: Make libraries Chromium-style "components"

Kangz:
Components are a concept defined in Chromium's BUILDCONFIG.gn that are
shared libraries (with the right rpath set up) or static_libraries
depending on whether is_component_build is defined.

This changes libshaderc and libshaderc_spvc to be components mostly so
that the right rpath options are set up automatically when they are
shared libraries.

Also fixes missing shaderc_util_public in the public_configs of both
libraries.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -58,18 +58,17 @@ source_set("shaderc_util_sources") {
 
 config("shaderc_public") {
   include_dirs = [ "libshaderc/include" ]
-  defines = [ "SHADERC_SHAREDLIB" ]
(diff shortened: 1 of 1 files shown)


Kangz:
Actually this fails on Linux, will update tomorrow.

dneto0:
fyi @fjhenigman

Kangz:
After more testing it seems like this *should* work on all platforms, in and outside of Chromium, with and without component builds.

PR #552, opened 2019-01-25
Flesh out for Dawn fuzzers.

fjhenigman:
Add enough functions to libshaderc_spvc so it can be used by Dawn fuzzers.

### DIFF
--- a/libshaderc_spvc/src/spvc.cc
+++ b/libshaderc_spvc/src/spvc.cc
@@ -15,10 +15,10 @@
 #include "shaderc/spvc.h"
 #include "libshaderc_util/exceptions.h"
 
-#include "spirv-tools/libspirv.hpp"
 #include "spirv-cross/spirv_glsl.hpp"
 #include "spirv-cross/spirv_hlsl.hpp"
 #include "spirv-cross/spirv_msl.hpp"
+#include "spirv-tools/libspirv.hpp"
 
 struct shaderc_spvc_compiler {};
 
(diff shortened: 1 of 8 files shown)


libshaderc_spvc/src/spvc.cc
```diff
@@ -89,6 +89,33 @@ void shaderc_spvc_compile_options_set_output_language_version(
...
+  options->glsl.vertex.fixup_clipspace = b;
+}
+
+void shaderc_spvc_compile_options_set_flip_vert_y(
+    shaderc_spvc_compile_options_t options, bool b) {
+  options->glsl.vertex.flip_vert_y = b;
+}
+
+// Fill options with given data.  Return amount of data used, or zero
+// if not enough data was given.
```
s-perron: These comments are not needed here.  They are on the prototype in the header file.

libshaderc_spvc/src/spvc.cc
```diff
@@ -89,6 +89,33 @@ void shaderc_spvc_compile_options_set_output_language_version(
...
+size_t shaderc_spvc_compile_options_set_for_fuzzing(
+    shaderc_spvc_compile_options_t options, const uint8_t *data, size_t size) {
+    if (!data || size < sizeof(*options))
+      return 0;
+
+    size_t used = std::min(size, sizeof(*options));
```
s-perron: The call to std::min is not needed here because you check that `size` is not smaller that `sizeof(*options)` above.

libshaderc_spvc/src/spvc.cc
```diff
@@ -105,17 +132,16 @@ void consume_validation_message(shaderc_spvc_compilation_result* result,
   result->messages.append(message);
   result->messages.append("\n");
 }
-}
 
-shaderc_spvc_compilation_result_t shaderc_spvc_compile_into_glsl(
-    const shaderc_spvc_compiler_t compiler, const uint32_t* source,
+shaderc_spvc_compilation_result_t validate_and_compile(spirv_cross::CompilerGLSL *compiler, const uint32_t* source,
```
s-perron: This function could use some comments to describe what it does.

PR #556, opened 2019-01-29
WIP Fix install.

### LINKED ISSUE #551: shaderc.h includes libshaderc_util headers, which are not installed (CMake)

fjhenigman:
A couple header files were not being installed.
Fixes #551.


fjhenigman:
> I find it confusing to have shaderc/visibility.h written in libshaderc_util/include/....
> I think I prefer them to moved into source tree at libshaderc/include/shaderc/visibility.h

I left them in libshaderc_util in case in future we wanted a switch to disable libshaderc, i.e. if someone wanted libshaderc_spvc but not libshaderc.  And all other shared stuff is in libshaderc_util.  Up to you.

dneto0:
> > I find it confusing to have shaderc/visibility.h written in libshaderc_util/include/....
> > I think I prefer them to moved into source tree at libshaderc/include/shaderc/visibility.h
> 
> I left them in libshaderc_util in case in future we wanted a switch to disable libshaderc, i.e. if someone wanted libshaderc_spvc but not libshaderc. And all other shared stuff is in libshaderc_util. Up to you.

I can buy that.  Right now libshaderc_spvc doesn't have a library or include dependency on libshaderc. So keeping the content in the libshaderc_util tree has merit.

dneto0:
Remaining two requests:

- please split "common.h" into env.h and status.h
- please fix the include guards.

dneto0:
Thanks for resolving.

PR #563, opened 2019-02-06
spvc is optional, off by default, in Android.mk

dneto0:
We need this until the NDK build has the SPIRV-Cross source.


dneto0:
I should explain:

- the main part of the Android NDK is where it builds the host tools, including shader tools like Shaderc's glslc, and SPIRV-Tools.  This part of the build invokes the shader-tools build via Cmake.  That already has the spvc build off by default.

- the second part of the Android NDK build is where it runs self tests, including doing per-ABI cross compiles using the shader-tools sources, using the ndk-build infrastructure (and hence the Android.mk).  

It's this second step which fails without this change.

PR #565, opened 2019-02-22
Add spvc command line tool and tests.

fjhenigman:
Add the command line tool 'spvc' which is a wrapper around spirv-cross.
It is tested with the same input and expectations as spirv-cross, except
we keep a binary version of each test shader here, and an explicit list
of test cases.


fjhenigman:
I forgot to take off the "WIP" - go ahead and review, I'll fix that later.

dj2:
I'm not sure if you can just commit the spirv-cross tests here due to license. You probably need to do something like access them from the spirv-cross checked out in third_party?

fjhenigman:
Dunno why it fails.  It works for me locally and it looks like I've uploaded the PR'd the right branch.  Will investigate further when I get a chance.

fjhenigman:
> I'm not sure if you can just commit the spirv-cross tests here due to license. You probably need to do something like access them from the spirv-cross checked out in third_party?  

I've compiled the spirv-cross tests and checked in the binaries here, to avoid versioning issues with spirv-as and glslang.  Do we need a lawyer to review?  We could do everything out of third_party/spirv-cross at test time, but the tests would take way longer.

dj2:
I'd vote against committing the compiled binaries. Compiling the GLSL shouldn't be too slow, you just need to call in to shaderc to build the spv file, don't need to write to disk, just pass the memory into your parser. Then you can just read the .glsl input files directly from the third_party. For the .spvasm files I think we already link spirv-tools? If so, you can do the assembly in memory pretty easy as well.

spvc/README.asciidoc
```diff
@@ -0,0 +1,61 @@
...
+
+==== `--validate=<target env>`
+
+`--validate` validates the input with the given target environment, which is one of:
+
+* vulkan1.0
+* vulkan1.1
+* opengl
+* opengl4.5
+* opengl_compate
```
dneto0: I think opengl_compat is not really working anymore (I think Glslang removed SPIR-V codegen support for it).  Let's not advertise it.

spvc/src/main.cc
```diff
@@ -0,0 +1,183 @@
...
+  auto status = result.GetCompilationStatus();
+  if (status == shaderc_compilation_status_validation_error) {
+    std::cerr << "validation failed:\n" << result.GetMessages() << std::endl;
+    return 1;
+  } else if (status == shaderc_compilation_status_success) {
+    const char* path = output_path.data();
+    if (path && strcmp(path, "-")) {
+      std::basic_ofstream<char>(path) << result.GetOutput();
+    } else {
+      std::cout << result.GetOutput();
```
dneto0: Do we need a binmode to accommodate Windows end-of-line translation?

dneto0:
> I'd vote against committing the compiled binaries. Compiling the GLSL shouldn't be too slow, you just need to call in to shaderc to build the spv file, don't need to write to disk, just pass the memory into your parser. Then you can just read the .glsl input files directly from the third_party. For the .spvasm files I think we already link spirv-tools? If so, you can do the assembly in memory pretty easy as well.

Yes, both Glslang and SPIRV-Tools are available to Shaderc.  So the test code can call into them.

I've been thinking also that checking in the binaries is going to make rolling deps forward less automatable.  (E.g. we would need to customize the regular Chrome deps rollers?)

dneto0:
The bots are showing some failures: I saw an apparent race on files in a VS2017 bot, and a more basic arguments failure on a Linux-based bot.

fjhenigman:
> > I'd vote against committing the compiled binaries. Compiling the GLSL shouldn't be too slow, you just need to call in to shaderc to build the spv file, don't need to write to disk, just pass the memory into your parser. Then you can just read the .glsl input files directly from the third_party. For the .spvasm files I think we already link spirv-tools? If so, you can do the assembly in memory pretty easy as well.
> 
> Yes, both Glslang and SPIRV-Tools are available to Shaderc. So the test code can call into them.
> 
> I've been thinking also that checking in the binaries is going to make rolling deps forward less automatable. (E.g. we would need to customize the regular Chrome deps rollers?)

I don't think so, not unless when they roll shaderc in chrome they also build and run all the shaderc tests.
I've sent you guys a design doc.  Let's move the debate there.

spvc/src/main.cc
fjhenigman: I don't think so, tests are passing on latest version.

fjhenigman:
I made a new PR.  It changed so much that made more sense than appending onto this one.

PR #569, opened 2019-03-01
Upate python scripts work with python3.

s-perron:
SPIR-V Tools has a python script that does not work with python3, so
cmake should look for python2 specifically.


dneto0:
Ugh.  This doesn't work when using a Python3.5 virtualenv.  That's the environment that Android NDK developers are using.   The symptoms are the same as before.

glslc/test/expect.py
```diff
@@ -550,16 +552,16 @@ def check_stdout_match(self, status):
...
                                'Expected:\n{ex}'.format(
                                    ac=status.stdout, ex=self.expected_stdout))
         else:
             if not self.expected_stdout.search(convert_to_unix_line_endings(
-                status.stdout)):
+                status.stdout.decode('utf8'))):
                 return False, ('Incorrect stdout output:\n{ac}\n'
                                'Expected to match regex:\n{ex}'.format(
-                                   ac=status.stdout,
+                                   ac=status.stdout.decode('uft8'),
```
dneto0: "utf8" instead.

PR #571, opened 2019-03-04
Add spvc command line tool and tests.

fjhenigman:
Initial version of command line tool 'spvc' which is a wrapper around
spirv-cross.  Testing is the same as for spirv-cross, but not all
pass yet.

### DIFF
--- a/spvc/.clang-format
+++ b/spvc/.clang-format
@@ -1,5 +1,0 @@
----
-# Use Google code formatting rules.
-Language: Cpp
-BasedOnStyle: Google
-...
--- a/spvc/.clang-format
+++ b/spvc/.clang-format
@@ -1,0 +1,5 @@
+---
+# Use Google code formatting rules.
+Language: Cpp
+BasedOnStyle: Google
+...
--- a/spvc/CMakeLists.txt
+++ b/spvc/CMakeLists.txt
@@ -1,0 +1,17 @@
+add_executable(spvc_exe src/main.cc)
+shaderc_default_compile_options(spvc_exe)
+target_include_directories(spvc_exe PRIVATE ${shaderc_SOURCE_DIR}/libshaderc/include ${spirv-tools_SOURCE_DIR}/include)
+set_target_properties(spvc_exe PROPERTIES OUTPUT_NAME spvc)
+target_link_libraries(spvc_exe PRIVATE shaderc_spvc shaderc_util)
+add_dependencies(spvc_exe build-version)
+
+shaderc_add_asciidoc(spvc_doc_README README)
+
+if(SHADERC_ENABLE_INSTALL)
+  install(TARGETS spvc_exe
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+endif(SHADERC_ENABLE_INSTALL)
+
+if(${SHADERC_ENABLE_TESTS})
+  add_subdirectory(test)
+endif()
--- a/spvc/README.asciidoc
+++ b/spvc/README.asciidoc
@@ -1,0 +1,60 @@
+= spvc Manual
+:toc:
+:toclevels: 3
+:numbered:
+:source-highlighter: pygments
+
+== Name
+
+`spvc` - A command-line SPIR-V to GLSL/HLSL/MSL compiler with
+Clang-compatible arguments.
+
+== Synopsis
+
+----
+spvc [--help]
+
+spvc [--version]
+     [--validate=<target env>]
+     [-o outfile]
+     shader...
+----
+
+== Description
+
+=== Input file languages
+
+Spvc accepts SPIR-V binary files as input.
+
+[[output-file-naming]]
+=== Output file naming
+
+If a name is specified via `-o`, the output file will be given that name.
+Otherwise the output file name is the same as the input file name except
+with a `.glsl,` `.hlsl` or `.msl` extension, as appropriate.
+
+== Command Line Options
+
+=== Overall Options
+
+==== `--help`
+
+`--help` tells the spvc compiler to display all available options and exit.
+
+==== `--version`
+
+`--version` displays version information.
+
+==== `--validate=<target env>`
+
+`--validate` validates the input with the given target environment, which is one of:
+
+* vulkan1.0
+* vulkan1.1
+* opengl
+* opengl4.5
+
+==== `-o`
+
+`-o` lets you specify the output file's name. It cannot be used when there are
+multiple files generated. A filename of `-` represents standard output.
--- a/spvc/README.asciidoc
+++ b/spvc/README.asciidoc
@@ -6,8 +6,7 @@
 
 == Name
 
-`spvc` - A command-line SPIR-V to GLSL/HLSL/MSL compiler with
-Clang-compatible arguments.
+`spvc` - A validating SPIR-V to GLSL/HLSL/MSL compiler with Clang-compatible options.
 
 == Synopsis
 
@@ -21,6 +20,9 @@ spvc [--version]
 ----
 
 == Description
+
+`spvc` is a wrapper around SPIRV-Cross.
+A SPIR-V module must pass validation before it is given to SPIRV-Cross for translation.
 
 === Input file languages
 
@@ -51,8 +53,8 @@ with a `.glsl,` `.hlsl` or `.msl` extension, as appropriate.
 
 * vulkan1.0
 * vulkan1.1
-* opengl
-* opengl4.5
+
+The default is `vulkan1.0`.
 
 ==== `-o`
 
--- a/spvc/src/main.cc
+++ b/spvc/src/main.cc
@@ -46,7 +46,7 @@ Options:
 )";
 }
 
-// TODO factor out
+// TODO(fjhenigman): factor out
 // Gets the option argument for the option at *index in argv in a way consistent
 // with clang/gcc. On success, returns true and writes the parsed argument into
 // *option_argument. Returns false if any errors occur. After calling this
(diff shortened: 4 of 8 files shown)


fjhenigman:
Differences from first PR: no shaders checked in, no list of tests with expectations.
Here I'm using spirv-tools and glslang from third_party in the spvc testing.
At the moment these happen to be the same revisions that spirv-cross has chosen for testing.
I don't think it's safe to assume they will always be the same.  Let me know if you think this should be addressed now or later.

spvc/.clang-format
```diff
@@ -0,0 +1,5 @@
+---
```
dj2: This should live in the root dir and all the source using the same formatting. It looks like shaderc isn't formatted yet?

spvc/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+add_executable(spvc_exe src/main.cc)
+shaderc_default_compile_options(spvc_exe)
+target_include_directories(spvc_exe PRIVATE ${shaderc_SOURCE_DIR}/libshaderc/include ${spirv-tools_SOURCE_DIR}/include)
+set_target_properties(spvc_exe PROPERTIES OUTPUT_NAME spvc)
+target_link_libraries(spvc_exe PRIVATE shaderc_spvc shaderc_util)
+add_dependencies(spvc_exe build-version)
+
+shaderc_add_asciidoc(spvc_doc_README README)
+
+if(SHADERC_ENABLE_INSTALL)
```
dj2: Does this need to be ${SHADERC_ENABLE_INSTALL} ?

spvc/README.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+
+`--version` displays version information.
+
+==== `--validate=<target env>`
+
+`--validate` validates the input with the given target environment, which is one of:
+
+* vulkan1.0
+* vulkan1.1
+* opengl
```
dj2: What version of opengl does this mean?

spvc/src/main.cc
```diff
@@ -0,0 +1,179 @@
...
+  --help                Display available options.
+  --version             Display compiler version information.
+  -o <output file>      '-' means standard output.
+  --validate=[<env>]    Validate SPIR-V source with given environment
+                        <env> is empty to skip validation, or one of the following:
+                        vulkan1.0 (default) vulkan1.1 opengl opengl4.5
+)";
+}
+
+// TODO factor out
```
dj2: TODO(fjhenigman): factor out

spvc/src/main.cc
```diff
@@ -0,0 +1,179 @@
...
+// consumed.
+bool GetOptionArgument(int argc, char** argv, int* index,
+                       const std::string& option,
+                       string_piece* option_argument) {
+  const string_piece arg = argv[*index];
+  assert(arg.starts_with(option));
+  if (arg.size() != option.size()) {
+    *option_argument = arg.substr(option.size());
+    return true;
+  } else {
```
dj2: else not needed after return

spvc/src/main.cc
```diff
@@ -0,0 +1,179 @@
...
+        options.SetTargetEnvironment(shaderc_target_env_vulkan,
+                                     shaderc_env_version_vulkan_1_0);
+      } else if (target_env == "vulkan1.1") {
+        options.SetTargetEnvironment(shaderc_target_env_vulkan,
+                                     shaderc_env_version_vulkan_1_1);
+      } else if (target_env == "opengl") {
+        options.SetTargetEnvironment(shaderc_target_env_opengl,
+                                     shaderc_env_version_vulkan_1_1);
+      } else if (target_env == "opengl4.5") {
+        options.SetTargetEnvironment(shaderc_target_env_opengl,
```
dj2: We don't differentiate opengl versions here?

spvc/src/main.cc
```diff
@@ -0,0 +1,179 @@
...
+                                     shaderc_env_version_vulkan_1_1);
+      } else {
+        std::cerr << "spvc: error: invalid value '" << target_env
+                  << "' in '--validate=" << target_env << "'" << std::endl;
+        return 1;
+      }
+    } else {
```
dj2: This can just be an else if

spvc/src/main.cc
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+
+  shaderc_spvc::CompilationResult result = compiler.CompileSpvToGlsl(
+      (const uint32_t*)input.data(), input.size(), options);
+  auto status = result.GetCompilationStatus();
+  if (status == shaderc_compilation_status_validation_error) {
+    std::cerr << "validation failed:\n" << result.GetMessages() << std::endl;
+    return 1;
+  } else if (status == shaderc_compilation_status_success) {
```
dj2: else not required since the if() returns.

spvc/test/run_spirv_cross_tests.py
```diff
@@ -0,0 +1,171 @@
...
+    ''' make sure directory exists to contain given file '''
+    dirpath = os.path.dirname(filepath)
+    try:
+        os.makedirs(dirpath)
+    except OSError as exc:
+        if not (exc.errno == errno.EEXIST and os.path.isdir(dirpath)):
+            raise
+
+def spirv_as(inp, out):
+    subprocess.check_call([spirv_as_path, '-o', out, inp], stdout=devnull)
```
dj2: Do we still get stderr if the compile fails?

spvc/test/run_spirv_cross_tests.py
```diff
@@ -0,0 +1,171 @@
...
+
+def spirv_opt(inp, out):
+    subprocess.check_call([spirv_opt_path, '-o', out, inp], stdout=devnull)
+
+def glslangValidator(inp, out):
+    subprocess.check_call([glslangValidator_path , '--target-env', 'vulkan1.1', '-V', '-o',
```
dj2: Does this target env not depend on how we're compiling? If we are using a vulkan1.0 env, shouldn't this be 1.0?

spvc/.clang-format
fjhenigman: shaderc has .clang-format files in glslc, libshaderc, libshaderc_spvc and libshaderc_util.  They are all the same.
I'll make a PR to have one at the root.

spvc/CMakeLists.txt
fjhenigman: No, and this shorter form is used all over shaderc.
I see I've been inconsistent however - I'll fix that.

spvc/README.asciidoc
fjhenigman: I'll remove that until it's covered by a test.

spvc/src/main.cc
fjhenigman: done

spvc/src/main.cc
fjhenigman: done

spvc/src/main.cc
fjhenigman: I'll remove until there's a test for this.

spvc/src/main.cc
fjhenigman: I think I like it better as is because the if else chain deals only with 'arg' and the inside code does what it needs to do for each specific arg.
You woudn't want "if (arg.starts_with("-o") && GetOptionArgument(....))" would you?

spvc/src/main.cc
fjhenigman: done

spvc/test/run_spirv_cross_tests.py
fjhenigman: yes

spvc/test/run_spirv_cross_tests.py
fjhenigman: spirv-cross always passes vulkan1.1 here so we do too

spvc/CMakeLists.txt
dj2: SHADERC_ENABLE_INSTALL and ${SHADERC_ENABLE_INSTALL} mean different things in cmake. I think it has to do with caching and options? I ran into issues with switftshader where it needed to switch to the ${} form.

It may work correctly with shaderc by itself, but embedding it into other projects is where I ran into issues with swiftshader.

fjhenigman: Actually only "if(SHADERC_ENABLE_INSTALL)" appears in shaderc at present.  If that's wrong, file a bug.

dj2: k

spvc/README.asciidoc
```diff
@@ -0,0 +1,60 @@
+= spvc Manual
+:toc:
+:toclevels: 3
+:numbered:
+:source-highlighter: pygments
+
+== Name
+
+`spvc` - A command-line SPIR-V to GLSL/HLSL/MSL compiler with
```
dneto0: I think there is no way to turn off validation.
And always-on validation was a key motivation for spvc in the first place.  So I'd like to put that up front. 

Suggest:

     spvc  - A validating SPIR-V to GLSL/HLSL/MSL compiler with Clang-compatible options

spvc/README.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+spvc [--help]
+
+spvc [--version]
+     [--validate=<target env>]
+     [-o outfile]
+     shader...
+----
+
+== Description
+
```
dneto0: I think it's useful here to describe the relationship with SPIRV-Cross.

Suggest

     spvc is a wrapper around SPIRV-Cross.   A SPIR-V module must pass validation before it is given to SPIRV-Cross for translation.

spvc/README.asciidoc
```diff
@@ -0,0 +1,60 @@
...
+==== `--validate=<target env>`
+
+`--validate` validates the input with the given target environment, which is one of:
+
+* vulkan1.0
+* vulkan1.1
+* opengl
+* opengl4.5
+
```
dneto0: Suggest adding:

      The default is `vulkan1.0`

Actually, this detail should also be stated in the C and C++ headers.  But that's for another CL.

spvc/README.asciidoc
fjhenigman: done.

spvc/README.asciidoc
fjhenigman: done

spvc/README.asciidoc
fjhenigman: done

PR #572, opened 2019-03-04
Fix spvc_c_smoke_test.

fjhenigman:
Free some memory we forgot to free.

### DIFF
--- a/libshaderc_spvc/src/spvc_c_smoke_test.c
+++ b/libshaderc_spvc/src/spvc_c_smoke_test.c
@@ -54,6 +54,7 @@ int main() {
   shaderc_spvc_compiler_t compiler;
   shaderc_spvc_compilation_result_t result;
   shaderc_spvc_compile_options_t options;
+  int ret_code = 0;
 
   compiler = shaderc_spvc_compiler_initialize();
   options = shaderc_spvc_compile_options_initialize();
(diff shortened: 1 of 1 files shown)


libshaderc_spvc/src/spvc_c_smoke_test.c
```diff
@@ -83,6 +86,9 @@ int main() {
     printf("failed to produce hlsl\n");
```
dj2: Does this also need to release?

libshaderc_spvc/src/spvc_c_smoke_test.c
```diff
@@ -93,12 +99,12 @@ int main() {
     printf("failed to produce msl\n");
```
dj2: ditto? and the one on 73 as well?

libshaderc_spvc/src/spvc_c_smoke_test.c
```diff
@@ -73,6 +73,9 @@ int main() {
     printf("failed to produce glsl\n");
     return -1;
   }
+  printf("success! %lu characters of glsl\n",
```
dj2: Can this just be %zu and we drop the cast to (unsigned long)?

fjhenigman: No, some old windows compiler chokes on %zu.

libshaderc_spvc/src/spvc_c_smoke_test.c
fjhenigman: *need* to, no because the test has failed.  I'd just as soon not complicate it, even if that is poor form.

libshaderc_spvc/src/spvc_c_smoke_test.c
fjhenigman: ditto

libshaderc_spvc/src/spvc_c_smoke_test.c
dj2: Well, it still leaks in this case, yea? Which means if we write a test which fails this on purpose we're going to get an ASAN leak bug from it aren't we?

fjhenigman: I don't see how that's possible, this test has no input or options and is hard coded to pass, but I'll add the free's, it's good form.

PR #578, opened 2019-03-18
Enable spvc test.

fjhenigman:
Update DEPS to work with spirv-cross tests.
Fix spvc return status.

Simplified the test script:
- no caching of compiled shaders
- all testing steps in one place
- removed unsupported spvc flags
Also clean up temporary files so they can't mess with subsequent testing.

### DIFF
--- a/spvc/test/run_spirv_cross_tests.py
+++ b/spvc/test/run_spirv_cross_tests.py
@@ -43,34 +43,39 @@ test_case_dirs = (
 test_count = 0
 pass_count = 0
 
+devnull = None
+not_used, tmpfile = tempfile.mkstemp()
+
+# Quietly run a command.  Throw exception on failure.
 def check_call(cmd):
     global devnull
-    if subprocess.call(cmd, stdout=devnull):
-        print('failed command: ', ' '.join(cmd))
-        sys.exit(1)
+    subprocess.check_call(cmd, stdout=devnull)
 
+# Run spirv-as.  Throw exception on failure.
 def spirv_as(inp, out, flags):
     check_call([spirv_as_path] + flags + ['-o', out, inp])
 
+# Run spirv-opt.  Throw exception on failure.
 def spirv_opt(inp, out, flags):
     check_call([spirv_opt_path] + flags + ['--skip-validation', '-O', '-o', out, inp])
 
+# Run glslangValidator as a compiler.  Throw exception on failure.
 def glslang_compile(inp, out, flags):
     check_call([glslangValidator_path] + flags + ['-o', out, inp])
 
+# Run spvc, return 'out' on success, None on failure.
 def spvc(inp, out, flags):
     global devnull
-    global test_count
-    test_count += 1
     cmd = [spvc_path] + flags + ['-o', out, '--validate=vulkan1.1', inp]
-    return subprocess.call(cmd, stdout=devnull)
+    return None if subprocess.call(cmd, stdout=devnull) else out
 
+# Compare result file to reference file and count matches.
 def check_reference(result, reference):
-    global spirv_cross_dir
     global pass_count
-    if result and filecmp.cmp(result, os.path.join(spirv_cross_dir, reference), False):
+    if filecmp.cmp(result, reference, False):
         pass_count += 1
 
+# Remove files and be quiet if they don't exist or can't be removed.
 def remove_files(*filenames):
     for i in filenames:
         try:
@@ -78,11 +83,17 @@ def remove_files(*filenames):
         except:
             pass
 
+# Test spvc producing GLSL the same way SPIRV-Cross is tested.
+# There are three steps: prepare input, convert to GLSL, check result.
 def test_glsl(shader, filename, optimize):
     global spirv_cross_dir, tmpfile
+    global test_count
     shader_path = os.path.join(spirv_cross_dir, shader)
 
-    # compile shader
+    # Prepare Vulkan binary.  The test input is either:
+    # - Vulkan text, assembled with spirv-as
+    # - GLSL, converted with glslang
+    # Optionally pass through spirv-opt.
     if '.asm.' in filename:
         flags = []
         if '.preserve.' in filename:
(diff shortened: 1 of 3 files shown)


spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,112 @@
...
-def mkdir_p(filepath):
-    ''' make sure directory exists to contain given file '''
-    dirpath = os.path.dirname(filepath)
-    try:
-        os.makedirs(dirpath)
-    except OSError as exc:
-        if not (exc.errno == errno.EEXIST and os.path.isdir(dirpath)):
-            raise
+def check_call(cmd):
+    global devnull
```
dneto0: I dislike backdoor declarations of global variables.  It was better to have devnull and tmpfile at global scope.

(I didn't even know Python could do this. Seems.... unpythonic?)

spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,112 @@
...
+    if subprocess.call(cmd, stdout=devnull):
+        print('failed command: ', ' '.join(cmd))
+        sys.exit(1)
```
dneto0: Either return an error or throw an exception.

spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,112 @@
...
-
-def mkdir_p(filepath):
-    ''' make sure directory exists to contain given file '''
-    dirpath = os.path.dirname(filepath)
-    try:
-        os.makedirs(dirpath)
-    except OSError as exc:
-        if not (exc.errno == errno.EEXIST and os.path.isdir(dirpath)):
-            raise
+def check_call(cmd):
```
dneto0: The methods should have method comments.

This became acute for me when reading the spvc function (since this version of the code has mixed responsibility: to run the command *and* increment the test count.)

spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,112 @@
...
-    else:
-        glslangValidator(source_path, binary_path)
-    if optimize:
-        spirv_opt(binary_path, binary_path)
-
-def do_test(binary_path, reference_path, flags):
-    global test_count, pass_count
+    global devnull
+    global test_count
     test_count += 1
```
dneto0: I would not modify the test counter inside this utility function.  That mixes responsibilities.

spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,112 @@
...
-    global test_count
-    test_count += 1
+def remove_files(*filenames):
+    for i in filenames:
+        try:
+            os.remove(i)
+        except:
+            pass
+
+def test_glsl(shader, filename, optimize):
```
dneto0: Very much needs a method comment.

This method name has glsl in it, but it's definitely handling SPIR-V assembly cases, etc.
Badly named method?

dneto0: This script or this method should describe that it's tracking the behaviours of SPIRV-Cross' own testing tactics.  Give the next person a place to look in SPIRV-Cross itself when there's new or changed behaviour to be replicated.

spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,112 @@
...
+    # run spvc
+    flags = []
+    output = output_vk = ""
+    if not '.nocompat.' in filename:
+        output = tmpfile + filename
+        if spvc(tmpfile, output, flags):
+            output = ""
+    if '.vk.' in filename or '.asm.' in filename:
+        output_vk = tmpfile + 'vk' + filename
+        if spvc(tmpfile, output_vk, flags):
```
dneto0: I can see why you are tempted to put the test counting in the low level utility function.  Probably because there's so much business logic in constructing the right command lines.

I'd approach it differently.  I'd construct the elaborated list of command lines, then execute them one by one, counting at that point.  That would also give an easy way to add a -n or --dry-run option to preview exactly what's going to execute before it's executed.

That also would resolve the usability problem where: when something goes wrong, it's annoying to have to guess how to reconstruct exactly one test case for iterative fix/debug.

spvc/test/run_spirv_cross_tests.py
fjhenigman: done

spvc/test/run_spirv_cross_tests.py
fjhenigman: Whoops, that was a debugging print I meant to remove.
This will call subprocess.check_call and throw an exception.

spvc/test/run_spirv_cross_tests.py
fjhenigman: done

spvc/test/run_spirv_cross_tests.py
fjhenigman: agreed, done

spvc/test/run_spirv_cross_tests.py
fjhenigman: I expanded the comments.  The function could be split up, but there's also something to be said for not having to page up and down looking for functions.  Instead I've tried to clearly label the three main steps with comments.

spvc/test/run_spirv_cross_tests.py
fjhenigman: It's a little more complicated to separate listing the steps and running through them.
If spvc fails you need to skip the comparison step.  The step could maybe be "spvc then compare if needed" but there's a property here I want to preserve, namely doing everything in the same order as spirv-cross.  That way I can log what both are doing and easily compare.  I had logging here temporarily and removed it - will add later as an official feature.  That also addresses the issue of knowing what went wrong (tail the log).  The logging in spirv-cross was similarly temporary and I'll try to land it there eventually.
Ideally some day we get rid of this duplication, but I'm proceeding cautiously on trying to change spirv-cross.
I think this is about as simple as possible at the moment, and that's a good thing.  I'd like to flesh out the MSL and HLSL testing a bit and see what complications that brings before deciding on another direction for this script.

PR #579, opened 2019-03-18
spvc: Add some flags.

fjhenigman:
This builds on https://github.com/google/shaderc/pull/578

spvc: Add some flags.
    
Implement the following spvc flags and update the test:
  --entry
  --version
  --remove-unused-variables
  --vulkan-semantics
  --separate-shader-objects
  --flatten-ubo
The old --version flag is now -v.

### DIFF
--- a/libshaderc_spvc/src/spvc.cc
+++ b/libshaderc_spvc/src/spvc.cc
@@ -32,6 +32,9 @@ struct shaderc_spvc_compilation_result {
 
 struct shaderc_spvc_compile_options {
   bool validate = true;
+  bool remove_unused_variables = false;
+  bool flatten_ubo = false;
+  std::string entry_point;
   spv_target_env target_env = SPV_ENV_VULKAN_1_0;
   spirv_cross::CompilerGLSL::Options glsl;
   spirv_cross::CompilerHLSL::Options hlsl;
@@ -39,7 +42,9 @@ struct shaderc_spvc_compile_options {
 };
 
 shaderc_spvc_compile_options_t shaderc_spvc_compile_options_initialize() {
-  return new (std::nothrow) shaderc_spvc_compile_options;
+  shaderc_spvc_compile_options_t options = new (std::nothrow) shaderc_spvc_compile_options;
+  if (options) options->glsl.version = 0;
+  return options;
 }
 
 shaderc_spvc_compile_options_t shaderc_spvc_compile_options_clone(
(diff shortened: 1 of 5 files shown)


libshaderc_spvc/src/spvc.cc
```diff
@@ -171,7 +202,88 @@ shaderc_spvc_compilation_result_t validate_and_compile(
...
+  if (!result) return nullptr;
+
   spirv_cross::CompilerGLSL compiler(source, source_len);
+
+  if (options->glsl.version == 0) {
+    // no version requested, was one detected in source?
+    options->glsl.version = compiler.get_common_options().version;
+    if (options->glsl.version == 0) {
+      // no version detected in source, use default
+      options->glsl.version = 450;
```
dj2: 450 should probably be a constant somewhere

libshaderc_spvc/src/spvc.cc
```diff
@@ -171,7 +202,88 @@ shaderc_spvc_compilation_result_t validate_and_compile(
...
+    uint32_t stage_count = 0;
+    for (auto &e : entry_points) {
+      if (e.name == options->entry_point) {
+        stage_count++;
+        model = e.execution_model;
+      }
+    }
+    if (stage_count != 1) {
+      result->status = shaderc_compilation_status_compilation_error;
+      if (stage_count)
```
dj2: Isn't this backwards? This should be if (stage_count == 0)?

libshaderc_spvc/src/spvc.cc
```diff
@@ -171,7 +202,88 @@ shaderc_spvc_compilation_result_t validate_and_compile(
...
+        result->messages = "There is no entry point with name: " + options->entry_point;
+      else
+        result->messages = "There is more than one entry point with name: " + options->entry_point + ". Use --stage.";
+      return result;
+    }
+  }
+  if (!options->entry_point.empty()) {
+    compiler.set_entry_point(options->entry_point, model);
+    }
```
dj2: formatting?

spvc/test/run_spirv_cross_tests.py
```diff
@@ -45,127 +43,135 @@
...
-    ''' make sure directory exists to contain given file '''
-    dirpath = os.path.dirname(filepath)
-    try:
-        os.makedirs(dirpath)
-    except OSError as exc:
-        if not (exc.errno == errno.EEXIST and os.path.isdir(dirpath)):
-            raise
+# Quietly run a command.  Throw exception on failure.
+def check_call(cmd):
+    global devnull
```
dj2: What does this do?

fjhenigman: You mean why the check_call() function?  No strong need for it now (just redirects stdout).  I expect to add logging here.

libshaderc_spvc/src/spvc.cc
fjhenigman: apologies, forgot to run clang-format

libshaderc_spvc/src/spvc.cc
fjhenigman: good catch

libshaderc_spvc/src/spvc.cc
fjhenigman: will do

spvc/test/run_spirv_cross_tests.py
dj2: Sorry, the global devnull thing? Haven't seen that before

fjhenigman: Just a handle to /dev/null which is opened/closed in main().
Yeah globals suck, but it doesn't seem much of an improvement to put this whole file into a class.

PR #584, opened 2019-03-29
WIP Zip and upload installed files.

fjhenigman:
WIP.  Want to see if this works, and still deciding how to package the builds.

At the end of the build do a "ninja install" then zip the result.
List the zip file in the kokoro configs so it gets uploaded.

### DIFF
--- a/kokoro/linux/build.sh
+++ b/kokoro/linux/build.sh
@@ -108,3 +108,6 @@ echo $(date): ctest completed.
 ninja install
 cd $(dirname $INSTALL_DIR)
 zip -r $KOKORO_ARTIFACTS_DIR/install.zip $(basename $INSTALL_DIR)
+
+# Test if gsutil to write to gcs
+gsutil cp /etc/passwd gs://shaderc/artifacts
--- a/kokoro/linux/build.sh
+++ b/kokoro/linux/build.sh
@@ -24,6 +24,7 @@ set -x
 
 BUILD_ROOT=$PWD
 SRC=$PWD/github/shaderc
+INSTALL_DIR="$SRC/install"
 CONFIG=$1
 COMPILER=$2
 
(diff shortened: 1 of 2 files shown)


kokoro/linux/build.sh
```diff
@@ -24,6 +24,7 @@ set -x
 
 BUILD_ROOT=$PWD
 SRC=$PWD/github/shaderc
+INSTALL_DIR="$SRC/install"
```
dneto0: Looks like this pollutes the source tree.  I'd put it beside the source tree, but under BUILD_ROOT.
Say, $BUILD_ROOT/install

fjhenigman: I put it beside the build dir.  We have:
$SRC/build
$SRC/install
If I move one I guess it makes sense to move both?

fjhenigman:
This worked, it uploaded the zip file to cloud storage, and I made the storage publicly readable.  However the files are not readily discoverable.  The web view of google storage does not generate directory listings, so you need to know the full url (which includes a unguessable time stamp) in order to download the build.  You can't just browse around and find the file you want.
One way to fix this is to add another build step which would update an index.html file in cloud storage.  This would be a fairly simple python script.  I'd have to figure out how to give the script permission to write to cloud storage but I think that's doable.  (The build upload happens automagically without us having to deal with permissions.)
There's a possible race condition if two builds (on the same builder) finish at the same time, but that seems highly unlikely and would be corrected by the next build that finishes.  If we're really worried about it, there's probably a way to prevent races.
Unless someone has a better idea I'll work on this.  @dneto0 @dj2 
I still need to check with @AWoloszyn how they get gapid releases to appear on github.  I suspect it's manual.

kokoro/linux/build.sh
dneto0: Ah, I didn't see the part later in the script where the build dir was created and then cd'd into.

What you have is ok then.  It makes an "install" dir beside the build dir.

dj2:
Going to close this for now as I don't think it's being worked on at the moment. Please re-open if this is still in progress.

PR #587, opened 2019-04-05
spvc: Add ability to use SPIRV-Tools parser.

fjhenigman:
Add --parser flag to select either SPIRV-Tools or SPIRV-Cross SPIR-V parser.
Test both options.

### DIFF
--- a/libshaderc_spvc/src/spvc.cc
+++ b/libshaderc_spvc/src/spvc.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <iterator>
+#include <sstream>
 
 #include "shaderc/spvc.h"
 #include "libshaderc_util/exceptions.h"
@@ -177,16 +178,16 @@ namespace {
 
 // Put the index-th operand of the instruction into 'out'
 template <typename Out>
-void GetOp(const spv_parsed_instruction_t* inst, unsigned index, Out& out) {
+void GetOp(const spv_parsed_instruction_t* inst, unsigned index, Out* out) {
   static_assert(sizeof(Out) == sizeof(uint32_t), "unexpected operand size");
   assert(index < inst->num_operands);
-  out = static_cast<Out>(inst->words[inst->operands[index].offset]);
+  *out = static_cast<Out>(inst->words[inst->operands[index].offset]);
 }
 
 // Specialized for string.
 template <>
 void GetOp(const spv_parsed_instruction_t* inst, unsigned index,
-           std::string& out) {
+           std::string* out) {
   assert(index < inst->num_operands);
   const spv_parsed_operand_t& op = inst->operands[index];
   assert(op.type == SPV_OPERAND_TYPE_LITERAL_STRING);
@@ -194,7 +195,7 @@ void GetOp(const spv_parsed_instruction_t* inst, unsigned index,
     for (unsigned i = 0; i < 4; ++i) {
       unsigned char c = (*p >> (i * 8)) & 0xff;
       if (!c) return;
-      out.push_back(c);
+      out->push_back(c);
     }
   }
 }
@@ -202,12 +203,25 @@ void GetOp(const spv_parsed_instruction_t* inst, unsigned index,
 // Specialized for ID list.
 template <>
 void GetOp(const spv_parsed_instruction_t* inst, unsigned index,
-           std::vector<uint32_t>& out) {
+           std::vector<uint32_t>* out) {
   if (index < inst->num_operands) {
     const spv_parsed_operand_t& op = inst->operands[index];
     assert(op.type == SPV_OPERAND_TYPE_ID);
     const uint32_t* p = inst->words + op.offset;
-    std::copy(p, p + inst->num_operands - index, std::back_inserter(out));
+    std::copy(p, p + inst->num_operands - index, std::back_inserter(*out));
+  }
+}
+
+// Specialized for ID list.
+template <>
+void GetOp(const spv_parsed_instruction_t* inst, unsigned index,
+           spirv_cross::SmallVector<uint32_t>* out) {
+  if (index < inst->num_operands) {
+    const spv_parsed_operand_t& op = inst->operands[index];
+    assert(op.type == SPV_OPERAND_TYPE_ID);
+    const uint32_t* p = inst->words + op.offset;
+    for (unsigned i=0; i<inst->num_operands - index;++i) out->push_back(p[i]);
+    //std::copy(p, p + inst->num_operands - index, std::back_inserter(*out));
   }
 }
 
@@ -217,14 +231,14 @@ void GetOpsRecur(const spv_parsed_instruction_t* inst, unsigned index) {}
 // Recursive template to do any number of GetOp().
 template <typename First, typename... Rest>
 void GetOpsRecur(const spv_parsed_instruction_t* inst, unsigned index,
-                 First& first, Rest&... rest) {
+                 First* first, Rest*... rest) {
   GetOp(inst, index, first);
   GetOpsRecur(inst, index + 1, rest...);
 }
 
 // Start the recursive template.
 template <typename... Outs>
-void GetOps(const spv_parsed_instruction_t* inst, Outs&... outs) {
+void GetOps(const spv_parsed_instruction_t* inst, Outs*... outs) {
   GetOpsRecur(inst, 0, outs...);
 }
 
@@ -275,7 +289,7 @@ struct Compiler {
       case spv::OpExtInstImport: {
         uint32_t id;
         std::string ext;
-        GetOps(inst, id, ext);
+        GetOps(inst, &id, &ext);
         if (ext == "GLSL.std.450")
           set<spirv_cross::SPIRExtension>(id, spirv_cross::SPIRExtension::GLSL);
         else if (ext == "SPV_AMD_shader_ballot")
@@ -303,13 +317,12 @@ struct Compiler {
         uint32_t id;
         std::string name;
         std::vector<uint32_t> interface_variables;
-        GetOps(inst, model, id, name, interface_variables);
+        GetOps(inst, &model, &id, &name, &interface_variables);
 
         auto itr = ir.entry_points.insert(
             std::make_pair(id, spirv_cross::SPIREntryPoint(id, model, name)));
         auto& e = itr.first->second;
-        std::copy(interface_variables.begin(), interface_variables.end(),
-                  back_inserter(e.interface_variables));
+        for (const auto& i : interface_variables) e.interface_variables.push_back(i);
         ir.set_name(id, e.name);
         if (!ir.default_entry_point) ir.default_entry_point = id;
       } break;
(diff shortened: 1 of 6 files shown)


dneto0:
VS 2015: 
T:\src\github\shaderc\libshaderc_spvc\src\spvc.cc(208): error C3861: 'back_inserter': identifier not found

fjhenigman:
You can see how much spirv-cross code we'll need to duplicate.  I did about 10 instruction cases so far, just  100 or so to go...

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+    assert(result);
+    result->messages.append(message);
+    result->messages.append("\n");
+  }
+
+  static spv_result_t parse_header(void* user_data, spv_endianness_t endian,
+                                   uint32_t magic, uint32_t version,
+                                   uint32_t generator, uint32_t id_bound,
+                                   uint32_t reserved) {
+    return ((Compiler*)user_data)
```
dj2: static_cast<Compiler*>() instead of c-style cast?

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+    return ((Compiler*)user_data)->parse_instruction(inst);
+  }
+
+  // Duplicate behavior of SPIRV-Cross.
+  spv_result_t parse_instruction(const spv_parsed_instruction_t* inst) {
+    switch (inst->opcode) {
+      case spv::OpExtInstImport: {
+        uint32_t id;
+        std::string ext;
+        GetOps(inst, id, ext);
```
dj2: I don't understand how this template stuff works. Why do we not just end up scribbling all over memory for the string case?

Would this be simpler to just have GetOp<uint32>, GetOp<spv::ExecutionModel> and GetStringOp methods that that you call on the individual lines above instead of using the GetOps template?

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+    ir.set_id_bounds(id_bound);
+    return SPV_SUCCESS;
+  }
+
+  static spv_result_t parse_instruction(void* user_data,
+                                        const spv_parsed_instruction_t* inst) {
+    return ((Compiler*)user_data)->parse_instruction(inst);
+  }
+
+  // Duplicate behavior of SPIRV-Cross.
```
dj2: I'd drop this or just make it 'Process SPIR-V file'

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+          result->messages.append(
+              "Cannot start a block before ending the current block.");
+          return SPV_REQUESTED_TERMINATION;
+        }
+
+        current_block = &set<spirv_cross::SPIRBlock>(inst->result_id);
+      } break;
+
+      case spv::OpReturn: {
+        if (!current_block) {
```
dj2: Should we move these blocks into a method per instruction? This method is going to get really long as it is.

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+        current_block = nullptr;
+      } break;
+
+      case spv::OpMemoryModel:
+        // Don't need to do anything.
+        break;
+
+      default:
+        std::ostringstream msg;
+        msg << "Instruction " << inst->opcode << " not supported yet.";
```
dj2: Can you use std::to_string(inst->opcode) instead of ostringstream?

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+        result->messages.append(msg.str());
+        return SPV_REQUESTED_TERMINATION;
+    }
+
+    return SPV_SUCCESS;
+  }
+
+  // Copied from SPIRV-Cross.
+  template <typename T, typename... P>
```
dj2: Can we get this added into the ir header in spirv-cross so we don't need to copy?

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+    auto& v = spirv_cross::variant_set<T>(ir.ids[id], std::forward<P>(args)...);
+    v.self = id;
+    return v;
+  }
+
+  bool ok() {
+    return result && result->status == shaderc_compilation_status_success;
+  }
+
+  ~Compiler() { delete cross; }
```
dj2: Make cross a unique_ptr, then the delete can be removed.

fjhenigman:
> I don't think we want to allow switching the parser long term. I think we should mark those API additions as temporary and once the spirv-tools parser is working we drop all the stuff related to using the spirv-cross parser.

I expect the ability to use the spirv-cross parser to be quite useful on an ongoing basis.  Rolling in new spirv-cross versions is likely to be painful because of the amount reimplemented here.  Being able to use its parser means I can tackle it in two steps: first get the tests updated (using the old parser), then try our parser and see what breaks.

libshaderc_spvc/src/spvc.cc
fjhenigman: yep

libshaderc_spvc/src/spvc.cc
fjhenigman: I would rather write this

 GetOps(inst, op1, op2, op3);

instead of this

 GetOp(inst, op1, 1);
 GetOp(inst, op2, 2);
 GetOp(inst, op3, 3);

or even worse

 GetOpBlah(inst, op1, 1);
 GetOpFoo(inst, op2, 2);
 GetOpExecutionModel(inst, op3, 3);

A pattern that will appear 100 times once all the instruction cases are filled in.

GetOp is specialized to extract characters from the spirv stream and put them in the given std::string.

libshaderc_spvc/src/spvc.cc
fjhenigman: ok

libshaderc_spvc/src/spvc.cc
fjhenigman: sure

libshaderc_spvc/src/spvc.cc
fjhenigman: I'll try that

libshaderc_spvc/src/spvc.cc
dj2: You're already doing:

```
uint32_t id;
std::string ext;
GetOps(inst, id, ext);
```
This would just mean:
```
uint32_t id = GetOp<uint32_t>(0);
std::string ext = GetString(1);
````
(or GetOp<const char*> if that's easier)

**Comment by dj2** [comment]
I'm fine with having the ability to do it, I just don't think we should expose it through our public API. It isn't a 'supported' use case for this code, our only supported method is with the spirv-tools parser.

File: libshaderc_spvc/src/spvc.cc (hunk 6 above)
Comment (fjhenigman) [reply libshaderc_spvc/src/spvc.cc]: It's already there but private.
My long term goal is of course to drive down the amount of reimplementation and duplication.
I'd like to proceed slowly with some changes helpful to spirv-cross before asking for stuff that's more for our benefit.

File: libshaderc_spvc/src/spvc.cc (hunk 7 above)
Comment (fjhenigman) [reply libshaderc_spvc/src/spvc.cc]: will do

File: libshaderc_spvc/src/spvc.cc (hunk 6 above)
Comment (dj2) [reply libshaderc_spvc/src/spvc.cc]: In that case, can we file a bug in shaderc to try to expose this in spirv-cross and link the bug from the comment above?

Comment (fjhenigman) [reply libshaderc_spvc/src/spvc.cc]: Will do once I figure out which other functions I'd like made public.

File: libshaderc_spvc/src/spvc.cc (hunk 2 above)
Comment (fjhenigman) [reply libshaderc_spvc/src/spvc.cc]: Still seems like avoidable tedium, writing in types and indices, hundreds of times, just to get rid of one tricky template.  Can I comment it better?
Care to weigh in @dneto0 ?

File: libshaderc_spvc/src/spvc.cc [hunk 8]
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+  }
+
+  static spv_result_t parse_instruction(void* user_data,
+                                        const spv_parsed_instruction_t* inst) {
+    return ((Compiler*)user_data)->parse_instruction(inst);
+  }
+
+  // Duplicate behavior of SPIRV-Cross.
+  spv_result_t parse_instruction(const spv_parsed_instruction_t* inst) {
+    switch (inst->opcode) {
```
dneto0: This method body is far far too long to be embedded in the class/struct declaration.  It's hard to see the class/struct definition from a high level view.  Break this out.

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+      case spv::OpEntryPoint: {
+        spv::ExecutionModel model;
+        uint32_t id;
+        std::string name;
+        std::vector<uint32_t> interface_variables;
+        GetOps(inst, model, id, name, interface_variables);
+
+        auto itr = ir.entry_points.insert(
+            std::make_pair(id, spirv_cross::SPIREntryPoint(id, model, name)));
+        auto& e = itr.first->second;
```
dneto0: What is e here?   itr.first->second doesn't give me a clue at all.  Use a better name.  Having auto doesn't help either.

libshaderc_spvc/src/spvc.cc
dneto0: This method is so short that duplicating it is not a big deal at all to me.   If SPIRV-Cross didn't already have it I imagine we would make our own anyway.

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+            break;
+        }
+      } break;
+
+      case spv::OpCapability: {
+        spv::Capability cap;
+        GetOps(inst, cap);
+        if (cap == spv::CapabilityKernel) {
+          result->messages.append("Kernel capability not supported.");
+          return SPV_REQUESTED_TERMINATION;
```
dneto0: If anything deserves shortening, I'd write a method to combine issuing a message and returning a failure status.
(See the use of diag in the validator.)

dneto0: This is failure, not just requested termination.. ??

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
+            GetOpsRecur(inst, 2, execution.output_vertices);
+            break;
+          default:
+            break;
+        }
+      } break;
+
+      case spv::OpCapability: {
+        spv::Capability cap;
+        GetOps(inst, cap);
```
dneto0: I very much do not like this non-const ref argument pattern, and the effects of the recursive template thing.

Also the recursive template GetOps construction won't deal well with variably-determined arguments, e.g. for image instructions (see the use of the bitmask word that controls image operands). There are other cases like that too, e.g. loop control, memory access.

Please take a look at SPIRV-Tools source/opt/ir_loader.h and ir_loader.cpp
That is a much more straightforward conversion from parsed instructions (from spvBinarParse) to an internal representation.

libshaderc_spvc/src/spvc.cc
```diff
@@ -164,74 +174,370 @@ void shaderc_spvc_compiler_release(shaderc_spvc_compiler_t compiler) {
...
-    // worked.
-    result->status = shaderc_compilation_status_success;
-  }
-  CATCH_IF_EXCEPTIONS_ENABLED(...) {
-    result->status = shaderc_compilation_status_compilation_error;
-    result->messages = "Compilation failed.  Partial source:";
-    result->messages.append(compiler->get_partial_source());
+// Specialized for ID list.
+template <>
+void GetOp(const spv_parsed_instruction_t* inst, unsigned index,
```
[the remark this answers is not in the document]
dneto0: If you want to have some automation for extracting parameters, I'd make an read-stream object that you can right-shift out of.
But so far I don't think it's warranted, given that there are multiple examples of clients of the spvBinaryParse that have not needed it (e.g. SPIRV-Tools optimizer ir_loader being one example, the validator another, and the disassembler a third.)

dneto0:
Closing.  This is being done in a different upcoming MR.

PR #595, opened 2019-04-15
Package and upload builds.

fjhenigman:
Add a step at the end of each build to install then package the
installation directory.
Name the package in the config of each build we want to upload.
Add a page that points to all the latest builds which can be downloaded.

### DIFF
--- a/kokoro/windows/build.bat
+++ b/kokoro/windows/build.bat
@@ -95,6 +95,7 @@ echo "Tests passed %DATE% %TIME%"
 :: ################################################
 :: Install and package.
 :: ################################################
+ninja install
 cd %SRC%
 zip -r install.zip install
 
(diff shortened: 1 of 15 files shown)


fjhenigman:
Ready for review.  You can see what the README will look like here:
https://github.com/fjhenigman/shaderc/tree/badge
I kept the three badges and linked them to what I thought would be the popular choices.
I added a separate downloads page (link just below the badges) that lists all available downloads.
The links are mostly broken until the continuous builders run but you can try the first badge on the README page.

kokoro/windows/build.bat
```diff
@@ -92,8 +92,15 @@ if %VS_VERSION% NEQ 2013 (
 )
 echo "Tests passed %DATE% %TIME%"
 
+:: ################################################
```
ehsannas: no `ninja install` here?

fjhenigman: will do

fjhenigman:
Note this will save every build, but only the latest of each kind are easily downloadable.
I made our google cloud storage publicly readable so anyone with an account can browse there to find older builds.
If we want to make older builds more easily accessible i.e. on a public web page, there's more work to do.
If we don't care about older builds, maybe we don't want to save them.  They're probably a drop in the bucket, but I haven't actually tried to estimate that.

PR #614, opened 2019-05-08
Roll glslang, spirv-tools, spirv-headers

dneto0:


dneto0:
Acknowledged.

fjhenigman:
Not at all surprising.  You kinda need to keep the following in sync:
{ spirv-tools, spirv-headers, glslang } as used in https://github.com/KhronosGroup/SPIRV-Cross/blob/master/checkout_glslang_spirv_tools.sh
with the same set as appears in https://github.com/google/shaderc/blob/master/DEPS

The alternative is to keep an extra set around in shaderc matching the ones spirv-cross uses for testing, just for running the spvc-spirv-cross tests.   Dan didn't want to do that but you can reevaluate which course is less painful.

dneto0:
Ok, I've updated again:
- takes yet a newer spirv-cross, which includes https://github.com/KhronosGroup/SPIRV-Cross/pull/970 which works around a buffer reference  issue in Glslang
- Number of passing tests goes from 1225 to 1246

Issue #612, opened 2019-05-08
Fix test breakage in shaderc_util_compiler


zoddicus:
CompilerTest.HlslFunctionality1Enabled in this test suite is failing, because the third_party/spirv-tools has not been rolled to pick up the SPIR-V 1.4 changes yet, but the test expectations were updated in #608 

Specifically OpDecorateStringGOOGLE was changed to OpDecorateString.

dneto0:
Rolling glslang, spirv-tools, spirv-headers via #614

dneto0:
When rolling forward SPIRV-Tools, SPIRV-Headers, and Glslang, that fails.

When I also roll forward SPIRV-Cross, I get a bunch of failures, including validation failures showing SPIR-V 1.4 modules failing to validate for Vulkan 1.1.
Many of those I can fix by forcing vulkan1.1 target environment for .asm. tests.

I still have a few left.

dneto0:
One of the errors looks like a bad test case in SPIRV-Cross:
https://github.com/KhronosGroup/SPIRV-Cross/issues/967

fjhenigman:
Did you run the spirv-cross tests (the actual spirv-cross tests, not the shaderc/spvc/spirv-cross tests)?
It seems unlikely (though anything is possible) that there's a spirv-cross rev where their tests don't pass.
If it works in spirv-cross it should work in shaderc/spvc/spirv-cross.  If not then it's a good bet https://github.com/google/shaderc/blob/master/spvc/test/run_spirv_cross_tests.py needs updating to correctly mimic the behavior of https://github.com/KhronosGroup/SPIRV-Cross/blob/master/test_shaders.py

dneto0:
Good point.  I just ran the SPIRV-Cross tests, when it's built with its own older versions of glslang, spirv-tools, and spirv-headers.  They pass, with a small handful of unrelated warnings (array of arrays, and precision).

The changelog on test_shaders.py is pretty slim lately and does not seem relevant.  So far the two issues @zoddicus  and I have identified are independent of SPIRV-Cross passing its own tests. That is, 
- the bad test case I identified above is legitimately invalid GLSL (with the extension) and Glslang is rightly rejecting it.
- When assembling .asm. cases, we should be explicitly specifying vulkan1.1 target environment (to get SPIR-V 1.3, rather than SPIR-V 1.4)

dneto0:
[...]

PR #615, opened 2019-05-08
Spvc tests: Assemble explicitly for Vulkan1.1 env

dneto0:
By default, SPIRV-Tools will assemble for the latest supported
version of SPIR-V.  With SPIR-V 1.4 arriving, that will generally
make modules which are not consumable by Vuklan 1.1.  So update
the Spvc tests to explicitly assemble for Vulkan 1.1.

### DIFF
--- a/spvc/test/run_spirv_cross_tests.py
+++ b/spvc/test/run_spirv_cross_tests.py
@@ -112,7 +112,7 @@ def compile_input_shader(shader, filename, optimize):
     global tmpfile
     shader_path = os.path.join(args.cross_dir, shader)
     if '.asm.' in filename:
-        flags = []
+        flags = ['--target-env', 'vulkan1.1']
         if '.preserve.' in filename:
             flags.append('--preserve-numeric-ids')
         spirv_as(shader_path, tmpfile, flags)


dneto0:
This corresponds to the SPIRV-Cross change https://github.com/KhronosGroup/SPIRV-Cross/pull/969

fjhenigman:
> I think the failure is acceptable, since it needs the deps roll to be fixed.

Your project now, you create revs that don't pass tests if you want.  :)

zoddicus:
> > I think the failure is acceptable, since it needs the deps roll to be fixed.
> 
> Your project now, you create revs that don't pass tests if you want. :)

The specific failure is coming from the changing the tests in anticipation of the uprev to SPIR-V 1.4. I am blocking my other PR until that is resolved, but I think this specific PR is part of getting everything rolled ahead and into a good state, so it makes sense to land with a know failure.

dneto0:
> Your project now, you create revs that don't pass tests if you want. :)

The trouble is there's a disconnect between SPIRV-Tools smoketest vs. Shaderc's self-test.

- SPIRV-Toosl smoke test builds Shaderc along with latest SPIRV-Tools and SPIRV-Headers.  Those latter are now defaulting to SPIR-V 1.4 and with the absorption of some SPV_GOOGLE_* extensions into core the default disassembly for certain tokens changes their string output. That's fixed by having merging #608 into Shaderc master
- Shaderc uses SPIRV-Tools and SPIRV-Headers pinned to older versions that don't support SPIR-V 1.4 yet.  This is causing the failures in #612 

Basically we sacrificed the Shaderc bots for the sake of the SPIRV-Tools bots.    This MR is part of a multi-step path to getting Shaderc back to stability, without re-breaking SPIRV-Tools bots.

PR #616, opened 2019-05-08
Fix spvc test logging

dneto0:
Use explicit list comprehension to map the command elements to strings.
Otherwise I get a "map object is not indexible" error when evaluating
cmd[0] two lines later.


fjhenigman:
yep, the way it was would only work in python 2

dneto0:
This is a net benefit.  Doesn't make the bot failures any worse, so I'm committing this.

PR #640, opened 2019-06-04
Start v2019.1 dev

dneto0:


dneto0:
I intend to tag 34c412f as  v2019.0

PR #695, opened 2019-06-12
Add invert-y option for glslc and libshaderc.

andrejnau:
Option invert-y for compiler is simple way for using DirectX shader code for Vulkan which uses a different coordinate system. Without the need to do it manually: ```gl_Position.y = -gl_Position.y``` or use [VK_KHR_MAINTENANCE1](https://www.saschawillems.de/blog/2019/03/29/flipping-the-vulkan-viewport/) extension.


andrejnau:
I signed it!

PR #725, opened 2019-06-18
Add -fnan-clamp

dneto0:
Tells the compiler to generate code for max and min builtins so that
when given a NaN operand, the builtin returns the other operand.
Similarly, the clamp builtin will favour the non-NaN operands, as
if clamp were implemented as a composition of max and min.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,10 @@
 Revision history for Shaderc
 
 v2019.1-dev 2019-06-04
- - Start v2019.1-dev
+ - Add -fnan-clamp: Generate code for max and min builtins so that,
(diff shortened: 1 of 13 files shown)


dneto0:
The ubuntu-clang-release bot failure is due to infra:

++ ./utils/git-sync-deps
fatal: unable to access 'https://github.com/google/googletest.git/': Failed to connect to github.com port 443: Connection timed out

dneto0:
Whoops.  Need to update DEPS to newer glslang that supports the option.

dneto0:
Rebased on top of a previous DEPS roll that included the necessary Glslang.

PR #733, opened 2019-06-21
BUILD.gn fixes for Fuchsia.

digit-google:
In order to allow the Fuchsia platform source tree to
use the newest shaderc upstream, this CL provides minor
fixes to its BUILD.gn. For the record, Fuchsia uses GN
with a different set of //build rules and configs.

- Add build_overrides/build.gni to let it defines
  |build_with_chromium| to false by default.

  Note that this is the standard location for this
  variable definition as used by Chromium and other
  projects that also use build_overrides/, see [1]
  for example.

- Scope all Chromium-specific statements under
  'if (build_with_chromium)' blocks.

  Note that building with Fuchsia requires disabling
  a small compiler warning due to one source file.

[1] https://github.com/KhronosGroup/SPIRV-Tools/blob/master/build_overrides/build.gni


digit-google:
Thanks, I fixed the newline issue. Let's wait now :)

build_overrides/build.gni
```diff
@@ -0,0 +1,18 @@
+# Copyright 2018 The Shaderc Authors. All rights reserved.
```
Kangz: nit: 2019

digit-google:
> I didn't see this in the PR, is it still applicable?

This was addressed in the second patch.

build_overrides/build.gni
digit-google: Done :)

digit-google:
> Regarding the lack of newline at eof: I suggest you use this PR to fix that.

Done!

zoddicus:
This appears to have broken the chromium roll. I think my original comment on the subject has disappeared into the ether.

An example of the failed roll:
https://chromium-review.googlesource.com/c/chromium/src/+/1676353

Seen error:
/b/s/w/ir/cache/builder/src/buildtools/linux64/gn gen //out/Release
  -> returned 1
ERROR at //third_party/shaderc/src/BUILD.gn:28:5: Undefined identifier.
    configs -= [ "//build/config/compiler:chromium_code" ]
    ^------
See //third_party/dawn/BUILD.gn:511:5: which caused the file to be included.
    "${dawn_shaderc_dir}:libshaderc",
    ^-------------------------------
GN gen failed: 1

# To recreate the file passed to argv[0] (input_path):
% cat > /b/s/w/ir/tmp/t/tmplAltXN.json <<EOF
{"additional_compile_targets": ["all"], "files": ["DEPS"], "test_targets": []}
EOF
%

digit-google:
I' m looking into it (I have a Chromium checkout on my machine)...

Le mar. 25 juin 2019 à 19:27, Ryan Harrison <notifications@github.com> a
écrit :

digit-google:
Fix should be in https://github.com/google/shaderc/pull/741

PR #741, opened 2019-06-25
Fix BUILD.gn for Chromium build.

digit-google:
Sorry about the noise. GN configs can be surprising :-(

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -23,13 +23,6 @@ config("shaderc_util_public") {
   include_dirs = [ "libshaderc_util/include" ]
 }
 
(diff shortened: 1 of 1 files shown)


digit-google:
I confirm that this works on my local Chromium checkout, however I would understand if you  would prefer confirmation from another Chrome reviewer.

zoddicus:
> I confirm that this works on my local Chromium checkout, however I would understand if you would prefer confirmation from another Chrome reviewer.

I think that is sufficient. Thanks for the quick turn around on this.

digit-google:
Thank you :)

PR #750, opened 2019-06-28
Add --target-spv option to set target SPIR-V version

dneto0:
Adds shaderc_compile_options_set_targte_spirv and associated enums.
Defaults to SPIR-V 1.0.

Passes that down into Glslang so we generate a module with the
specified version of SPIR-V.

Refactor how we map to Glslang target and client, and test that
better.

Setting spirv version is just an override.

Plumb through to glslc


dneto0:
This doesn't have integration tests for the specified SPIR-V version.  But I did test it manually on interesting cases.

PR #759, opened 2019-07-15
Explicitly use python3 for git-sync-deps

kiroma:
On few distributions Python is still being symlinked to Python 2 if it is installed.
Because this script only works on Python 3, we can simply ask for it directly.


dneto0:
The VS2017 failure looked like spurious linker problem.  Restarting the build.

PR #784, opened 2019-08-21
Force Glslang to support HLSL in its interface

dneto0:
Fixes a compilation problem introduced by recent
Glslang changes to optionally build into a very small
GLSL-only binary.

Also fix test cases to have set and binding layout qualifiers on
buffers when compiling for Vulkan.  This rule is now enforced by
Glslang.


zoddicus:
The failed bot looks like flake, triggered it re-running.

PR #795, opened 2019-09-03
Use agreed upon include paths for spirv-cross

Kangz:
All other projects using spirv-cross include it using <spirv_msl.hpp>
for example. Follow that convention so we can more easily integrate with
other projects.

### DIFF
--- a/libshaderc_spvc/src/spvc_private.h
+++ b/libshaderc_spvc/src/spvc_private.h
@@ -17,9 +17,9 @@
 #include <vector>
 
 #include "shaderc/spvc.h"
-#include "spirv-cross/spirv_glsl.hpp"
-#include "spirv-cross/spirv_hlsl.hpp"
-#include "spirv-cross/spirv_msl.hpp"
+#include <spirv_glsl.hpp>
+#include <spirv_hlsl.hpp>
+#include <spirv_msl.hpp>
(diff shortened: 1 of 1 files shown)


dj2:
The linter doesn't run for shaderc at the moment, but this would be a linter as includes need a path.

zoddicus:
> includes need a path

I feel like I am missing something here. Is this just a style guide type issue?

dj2:
Yea, style guide thing.

zoddicus:
> Yea, style guide thing.

k, this PR I think is intentionally breaking style to regularize with other usages. Would it be sufficient to document that? or is there no-lint guard we could put around it to avoid the linter potentially complaining in the future?

dj2:
I guess the question is, do we want to do this? This usage of spirv-cross is weird. What usage are we standardizing with?

Yes, if we turn on the linter we can ignore this with comments.

zoddicus:
> I guess the question is, do we want to do this? This usage of spirv-cross is weird. What usage are we standardizing with?
> 
> Yes, if we turn on the linter we can ignore this with comments.

I will defer to @Kangz to provide more details.

Kangz:
Quoting an internal chat with @dj2:

> @dj2: Bleh, k

Which I assume it a "I don't like it, but go ahead".

dj2:
lgtm

PR #802, opened 2019-09-07
Update all components

dneto0:
Use SPIRV-Tools stable commit from 2019-09-04, plus two commits
more, fixing SPIRV-Tools issue 2833


dneto0:
Internal verification passes, including NDK builds on Linux and Windows.

dneto0:
cc: @greg-lunarg

PR #807, opened 2019-09-17
[WIP] spvc: Update test script

sarahM0:
WIP, please do not review


zoddicus:
FYI, if you want to eliminate all of the upstream changes from being included in this, you could rebase your working branch onto an up to date master and do a force push, which will wipe out them out in the PR. This will also have a side effect of removing the history from the PR, which is why I wouldn't recommend it if people had already started doing reviews.

PR #822, opened 2019-10-02
Add support for targeting SPIR-V 1.5

dneto0:


dneto0:
vs2017-release and vs2017-debug are linker crashes in spirv-reduce and spirv-val.

Sigh

dneto0:
Resubmitting to kick off the bots again.

dneto0:
VS2107 debug build linker crashed again.

PR #827, opened 2019-10-03
Revert "Add spvc compiler to returned result"

zoddicus:
Reverts google/shaderc#821

This is needed because this PR exposed multiple issues related to spirv_cross inclusion. Specific exceptions can be included via header usage and issues with importing the reflect classes. I am going to roll this back, and investigate locally.


zoddicus:
TBRing

PR #833, opened 2019-10-08
Fix up use of private spirv-tools headers.

### LINKED ISSUE #832: Includes for spvcir need fixing
The includes for the spirir code are not generic enough.  See https://github.com/google/shaderc/blob/master/libshaderc_spvc/src/spvcir_pass.h#L19.  If SHADERC_SPIRV_TOOLS_DIR is set to a directory with a different name then the repo will not build.

This seems to work because the name of the directory that contains the spirv-tools repo is named "spirv-tools", and it is in the same directory as the spirv-cross source.

Note that we add `${SPIRV-Cross_SOURCE_DIR}/..` to the include path which is generally the entire `third_party` directory:

https://github.com/google/shaderc/blob/master/libshaderc_spvc/CMakeLists.txt#L25

This needs to be cleaned up.

s-perron:
Fixes #832

### DIFF
--- a/libshaderc_spvc/CMakeLists.txt
+++ b/libshaderc_spvc/CMakeLists.txt
@@ -22,7 +22,6 @@ target_include_directories(shaderc_spvc
   ${shaderc_SOURCE_DIR}/libshaderc_util/include
   ${spirv-tools_SOURCE_DIR}/include
   ${spirv-tools_SOURCE_DIR}/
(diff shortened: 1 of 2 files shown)


zoddicus:
Only one build remaining, and the non-debug version of it has passed already. Going to merge this now, since I have patches to fix builds I want to build on top of it.

PR #838, opened 2019-10-15
The glslang-testsuite is moved into glslang.

dj2:
Currently ShaderC does a full copy of the glslang testsuite during the build, rewrites the test script and then runs the tests as part of ctest. The code to run those tests has been added to the GLSLang CMake script so will be added to ctest automatically.

This CL removes the glslang-testsuite from shaderc and bumps the version of glslang so we will now use the version provided directly.


dj2:
Note, this requires the GLSLang side to land first.

s-perron:
I'm not clear on how the glslang tests will be run.  Were we running them twice before?  A little more explanation on what is happening would be good.

dj2:
Updated comment above. PTAL.

dneto0:
Hallelujah

Issue #651, opened 2019-06-05
Add support for non-spirv-cross tests to run_spirv_cross_tests.py


zoddicus:
Specifically allow adding new spirv-cross style tests to our repo. This will probably mean renaming the script :-)

zoddicus:
This will hopefully a good place to get started understanding how spvc works and the testing harness operated. Please reach out if you need help with this.

zoddicus:
spvc has been deprecated and removed.

PR #839, opened 2019-10-16
spvc parser for a trivial shader

sarahM0:
This is the first PR in a series of PRs for parsing and generating spiv-cross-ir using an opt pass.
fixes issue #651

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,7 +11,6 @@ message(STATUS "Shaderc: build type is \"${CMAKE_BUILD_TYPE}\".")
 
 option(SHADERC_ENABLE_SPVC "Enable libshaderc_spvc" OFF)
 option(SHADERC_ENABLE_SPVC_PARSER "Enable spvc parser \(WIP)" OFF)
-option(DISABLE_EXCEPTIONS "disable spirv-cross exceptions")
 
 option(SHADERC_SKIP_INSTALL "Skip installation" ${SHADERC_SKIP_INSTALL})
 if(NOT ${SHADERC_SKIP_INSTALL})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ message(STATUS "Shaderc: build type is \"${CMAKE_BUILD_TYPE}\".")
 
 option(SHADERC_ENABLE_SPVC "Enable libshaderc_spvc" OFF)
 option(SHADERC_ENABLE_SPVC_PARSER "Enable spvc parser \(WIP)" OFF)
+option(DISABLE_EXCEPTIONS "disable spirv-cross exceptions")
 
 option(SHADERC_SKIP_INSTALL "Skip installation" ${SHADERC_SKIP_INSTALL})
 if(NOT ${SHADERC_SKIP_INSTALL})
--- a/libshaderc_spvc/src/spvc_private.cc
+++ b/libshaderc_spvc/src/spvc_private.cc
@@ -300,32 +300,28 @@ shaderc_spvc_compilation_result_t generate_hlsl_shader(
 // compiler(SHADERC_ENABLE_SPVC_PARSER is OFF by default)
 // TODO (sarahM0): change the default to spvc IR generation when it's done
 #if SHADERC_ENABLE_SPVC_PARSER
-  spvtools::Optimizer opt(options->source_env);
-  opt.SetMessageConsumer(std::bind(
-      consume_spirv_tools_message, result, std::placeholders::_1,
-      std::placeholders::_2, std::placeholders::_3, std::placeholders::_4));
+  spirv_cross::ParsedIR ir;  // = new (std::nothrow) spirv_cross::ParsedIR();
+  {
+    spvtools::Optimizer opt(options->source_env);
+    opt.SetMessageConsumer(std::bind(
+        consume_spirv_tools_message, result, std::placeholders::_1,
+        std::placeholders::_2, std::placeholders::_3, std::placeholders::_4));
+    ir.spirv = std::vector<uint32_t>(source, source + source_len);
 
-  spirv_cross::ParsedIR* ir = new (std::nothrow) spirv_cross::ParsedIR();
-  if (!ir) {
-    result->messages.append("Unable to initialize SPIRV-Cross IR.\n");
-    return result;
+    opt.RegisterPass(
+        spvtools::Optimizer::PassToken(std::unique_ptr<spvtools::opt::Pass>(
+            reinterpret_cast<spvtools::opt::Pass*>(
+                new spvtools::opt::SpvcIrPass(&ir)))));
+
+    if (!opt.Run(source, source_len, &result->binary_output)) {
+      result->messages.append(
+          "Transformations between source and target "
+          "execution environments failed (spvc-ir-pass).\n");
+      result->status = shaderc_compilation_status_transformation_error;
+      return result;
+    }
   }
-  ir->spirv = std::vector<uint32_t>(source, source + source_len);
-
-  opt.RegisterPass(
-      spvtools::Optimizer::PassToken(std::unique_ptr<spvtools::opt::Pass>(
-          reinterpret_cast<spvtools::opt::Pass*>(
-              new spvtools::opt::SpvcIrPass(ir)))));
-
-  if (!opt.Run(source, source_len, &result->binary_output)) {
-    result->messages.append(
-        "Transformations between source and target "
-        "execution environments failed (spvc-ir-pass).\n");
-    result->status = shaderc_compilation_status_transformation_error;
-    return result;
-  }
-
-  compiler = new (std::nothrow) spirv_cross::CompilerHLSL(*ir);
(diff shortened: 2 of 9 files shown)


sarahM0:
@googlebot I fixed it.

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ message(STATUS "Shaderc: build type is \"${CMAKE_BUILD_TYPE}\".")
 
 option(SHADERC_ENABLE_SPVC "Enable libshaderc_spvc" OFF)
 option(SHADERC_ENABLE_SPVC_PARSER "Enable spvc parser \(WIP)" OFF)
+option(DISABLE_EXCEPTIONS "disable spirv-cross exceptions")
```
dneto0: This looks like it should go into third_party/CMakeLists.txt just before we include the spirv-cross subdirectory?

Otherwise the name looks too generic, applying to all code everywhere.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -14,25 +14,189 @@
 
 #include "spvcir_pass.h"
 
-#include <fstream>
+#include "source/opt/ir_context.h"
 
 namespace spvtools {
 namespace opt {
 
+SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR &ir) {
```
dneto0: Google style guide says that reference parameters should be const.
Since you'll be modifying this ir object, change the interface to make it pointer-to-spirv_cross::ParsedIR
That way it's clear that you'll be modifying it.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -14,25 +14,189 @@
...
+  ir_ = &ir;
+  current_function_ = nullptr;
+  current_block_ = nullptr;
+
+  auto s = ir_->spirv.data();
+  uint32_t bound = s[3];
```
dneto0: Suggest assert(ir_->spirv.size() > 3);

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -14,25 +14,189 @@
...
-  get_module()->ForEachInst([&oss, options](const Instruction* inst) {
-    oss << inst->PrettyPrint(options);
-    if (!IsTerminatorInst(inst->opcode())) {
-      oss << std::endl;
-    }
-  });
+  get_module()->ForEachInst(
+      [this](Instruction *inst) { genSpirvCrossIR(inst); });
+
+  assert(!current_function_ && "Function was not terminated.");
```
dneto0: It's ok to assert here because this class assumes the module passes validation. Suggest adding a comment to that effect.

libshaderc_spvc/src/spvcir_pass.h
```diff
@@ -25,16 +25,31 @@ namespace opt {
...
-  const char* name() const override { return "spvc-if-pass"; }
+  SpvcIrPass(spirv_cross::ParsedIR &ir);
+  const char *name() const override { return "spvc-if-pass"; }
   Status Process() override;
 
   IRContext::Analysis GetPreservedAnalyses() override {
     return IRContext::kAnalysisNone;
   }
 
+  void genSpirvCrossIR(Instruction *inst);
```
dneto0: Needs a method comment.
Style guide wants the method name to start with a capital.  (It's not a trivial accessor).

libshaderc_spvc/src/spvcir_pass.h
```diff
@@ -25,16 +25,31 @@ namespace opt {
...
+
  private:
-  spirv_cross::ParsedIR ir_;
+  spirv_cross::ParsedIR *ir_;
+  spirv_cross::SPIRFunction *current_function_ = nullptr;
+  spirv_cross::SPIRBlock *current_block_ = nullptr;
+
+  // copied from spirv-cross (class Parser)
```
dneto0: Say what this does:

// Adds an instruction to the SPIRV-Cross IR, for the given result Id and with given type and arguments.
// Returns the constructed instruction.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -14,25 +14,189 @@
...
+      break;
+    }
+
+    case SpvOpTypeFunction: {
+      auto id = inst->result_id();
+      auto result = inst->type_id();
+
+      // reading Parameter 0 Type, Parameter 1 Type, ...
+      auto &func = set<spirv_cross::SPIRFunctionPrototype>(id, result);
+      for (uint32_t i = 2; i < inst->NumInOperands(); ++i) {
```
dneto0: start i at 1 to skip over the function return type parameter.

This will only show itself to be a problem once you start having functions that have parameters. This code will skip the first parameter.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -14,25 +14,189 @@
...
+    case SpvOpTypeVoid: {
+      auto id = inst->result_id();
+      auto &type = set<spirv_cross::SPIRType>(id);
+      type.basetype = spirv_cross::SPIRType::Void;
+      break;
+    }
+
+    case SpvOpTypeFunction: {
+      auto id = inst->result_id();
+      auto result = inst->type_id();
```
dneto0: It's tricky, but the  "Return type" is not the type id.  You need `auto result = inst->GetSingleWordInOperand(0)` here.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -14,25 +14,189 @@
...
+      assert(!current_block_ &&
+             "Cannot start a block before ending the current block.");
+
+      current_block_ = &set<spirv_cross::SPIRBlock>(id);
+      break;
+    }
+
+    case SpvOpReturn: {
+      assert(current_block_ && "Trying to end a non-existing block.");
+      current_block_->terminator = spirv_cross::SPIRBlock::Return;
```
dneto0: Eventually I can see this being refactored into TerminateBlock( ... terminator type ...), which also resets current_block_ to nullptr.

But not yet, since you should only do that after having at least one more terminator case.

libshaderc_spvc/src/spvcir_pass.cc
sarahM0: Thanks for pointing this out. I think the same applies to SpvOpFunction, so I changed that too.

libshaderc_spvc/src/spvc_private.cc
```diff
@@ -305,8 +305,12 @@ shaderc_spvc_compilation_result_t generate_hlsl_shader(
       consume_spirv_tools_message, result, std::placeholders::_1,
       std::placeholders::_2, std::placeholders::_3, std::placeholders::_4));
 
-  spirv_cross::ParsedIR ir;
-  ir.spirv = std::vector<uint32_t>(source, source + source_len);
+  spirv_cross::ParsedIR* ir = new (std::nothrow) spirv_cross::ParsedIR();
```
dneto0: This ir object leaks now.
This part was better the way you first proposed it.

Note that registering the pass captures the pointer to the ir object, so you have to modify it to say `... spvtools::opt::SpvcIrPass(&ir)...`.    It's ok to have the pass have a pointer to the IR object because the pass will run while the IR object still exists on the stack.

libshaderc_spvc/src/spvc_private.cc
```diff
@@ -321,10 +325,9 @@ shaderc_spvc_compilation_result_t generate_hlsl_shader(
     return result;
   }
 
-  compiler.reset(new spirv_cross::CompilerHLSL(ir));
+  compiler = new (std::nothrow) spirv_cross::CompilerHLSL(*ir);
```
dneto0: I read down into spirv_cross::CompilerHLSL(const spirv_cross::ParsedIR&) and it looks like it makes a copy of the IR object.  You could use std::move here to transfer ownership of the contents of the ir object over to the CompilerHLSL. That might save some CPU time?? 

e.g.
```compiler = new (std::nothrow) spirv_cross::CompilerHLSL(std::move(*ir));```

Comment (dneto0) [reply libshaderc_spvc/src/spvc_private.cc]: This is an optional change. It's slightly tricky, and should be commented with why it's safe to do.

File: libshaderc_spvc/src/spvc_private.cc [hunk 12]
```diff
@@ -300,32 +300,28 @@ shaderc_spvc_compilation_result_t generate_hlsl_shader(
...
-          reinterpret_cast<spvtools::opt::Pass*>(
-              new spvtools::opt::SpvcIrPass(ir)))));
-
-  if (!opt.Run(source, source_len, &result->binary_output)) {
-    result->messages.append(
-        "Transformations between source and target "
-        "execution environments failed (spvc-ir-pass).\n");
-    result->status = shaderc_compilation_status_transformation_error;
-    return result;
+  spirv_cross::ParsedIR ir;  // = new (std::nothrow) spirv_cross::ParsedIR();
```
dneto0: Could (should) delete the dead-code-comment.

PR #842, opened 2019-10-18
spvc: add known_spvc_failures 

sarahM0:
Adding Known_spvc_failures file. 
In the upcoming PRs I will add new cases to SpvcIrPass and delete the newly supported tests from this file.


zoddicus:
> Should this new flag/test be part of the tests run by ctest?

or is that premature?

sarahM0:
This needs -DSHADERC_ENABLE_SPVC_PARSER=ON to be added to build which as you mention is a little to early. Down the road I will add another bot that builds with -DSHADERC_ENABLE_SPVC_PARSER=ON and does ctest with --run-spvc-test.

PR #845, opened 2019-10-21
spvc: add spvc parsing option to msl, glsl and vulkan compilers

sarahM0:


libshaderc_spvc/src/spvc_private.cc
```diff
@@ -332,27 +332,7 @@ shaderc_spvc_compilation_result_t generate_hlsl_shader(
...
-
-    if (!opt.Run(source, source_len, &binary_output)) {
-      result->messages.append(
-          "Transformations between source and target "
-          "execution environments failed (spvc-ir-pass).\n");
-      result->status = shaderc_compilation_status_transformation_error;
-      return result;
-    }
-  }
+  result = generate_spvcir(&ir, source, source_len, options, result);
```
zoddicus: This should check the result for an error status.

libshaderc_spvc/src/spvc_private.cc
```diff
@@ -390,27 +370,7 @@ shaderc_spvc_compilation_result_t generate_msl_shader(
...
```
zoddicus: ditto

libshaderc_spvc/src/spvc_private.cc
```diff
@@ -449,27 +409,7 @@ shaderc_spvc_compilation_result_t generate_vulkan_shader(
...
```
zoddicus: ditto

PR #847, opened 2019-10-23
spvc: parser enhancement

sarahM0:
Supports new cases in spvc parser.
Passing 304 out of 1500 tests (20%).


sarahM0:
Please do not review. Just realized there is a bug that I need to fix.

sarahM0:
Ready for review.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+      ptrbase.pointer_depth++;
+      ptrbase.storage =
+          static_cast<spv::StorageClass>(inst->GetSingleWordInOperand(0u));
+
+      if (ptrbase.storage == spv::StorageClassAtomicCounter)
+        ptrbase.basetype = spirv_cross::SPIRType::AtomicCounter;
+
+      ptrbase.parent_type = inst->GetSingleWordInOperand(1u);
+
+      // spirv-corss comment:
```
zoddicus: nit:cross

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -179,30 +637,116 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+          args[i] = inst->GetSingleWordInOperand(i + 1);
+        }
+        set<spirv_cross::SPIRConstantOp>(id, result_type, spec_op, args,
+                                         inst->NumOperandWords() - 3);
+      }
       break;
     }
 
-    default:
+    // TODO(sarahM0): These opcodes are processed in the generator.
```
zoddicus: There should probably be an issue on github tracking this.

sarahM0: issue #854

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -19,6 +19,43 @@
...
+  if (constant_type.pointer) {
+    auto &constant = set<spirv_cross::SPIRConstant>(id, type);
+    constant.make_null(constant_type);
+  } else if (!constant_type.array.empty()) {
+    assert(constant_type.parent_type &&
+           "constant type parent shouldn't be empty");
+    uint32_t parent_id = ir_->increase_bound_by(1);
+    make_constant_null(parent_id, constant_type.parent_type);
+
+    assert(constant_type.array_size_literal.back() &&
```
dneto0: I think this code only handles a one-dimensional array.   You're only checking the last dimension here.
From the SPIRV-Cross representation, it looks like constant_type.array_size_literal might be a long vector, with as many elements as number of nesting levels of the array.  That might be a fair assumption, but it should be stated in the method comment, and checked in the code (not just an assert).

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -19,6 +19,43 @@
...
+           "Array size of OpConstantNull must be a literal.");
+
+    spirv_cross::SmallVector<uint32_t> elements(constant_type.array.back());
+    for (uint32_t i = 0; i < constant_type.array.back(); i++)
+      elements[i] = parent_id;
+    set<spirv_cross::SPIRConstant>(id, type, elements.data(),
+                                   uint32_t(elements.size()), false);
+  } else if (!constant_type.member_types.empty()) {
+    uint32_t member_ids =
```
dneto0: Is this the struct case only?
Or does it cover all the non-array composites:  struct, vector, and matrix

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -83,28 +121,57 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+    case SpvOpExtInstImport: {
+      uint32_t id = inst->result_id();
+      std::string ext(
+          reinterpret_cast<const char *>(inst->GetInOperand(0u).words.data()));
+      if (ext == "GLSL.std.450")
+        set<spirv_cross::SPIRExtension>(id, spirv_cross::SPIRExtension::GLSL);
+      else if (ext == "DebugInfo")
+        set<spirv_cross::SPIRExtension>(
+            id, spirv_cross::SPIRExtension::SPV_debug_info);
+      else if (ext == "SPV_AMD_shader_ballot")
```
dneto0: It's unfortunate that SPIRV-Cross mixes the concepts of extended instruction set and extension.  But that's how it is, so this code must match that modeling.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+          execution.workgroup_size.y = inst->GetSingleWordInOperand(3u);
+          execution.workgroup_size.z = inst->GetSingleWordInOperand(4u);
+          break;
+
+        case SpvExecutionModeOutputVertices:
+          execution.output_vertices = inst->GetSingleWordInOperand(2u);
+          break;
+
+        default:
+          break;
```
dneto0: I would prefer this emits an error if there is at least one literal parameter, because that represents a case which is unhandled.  (E.g. this would occur if we ever anticipate using SPIR-V 1.4 and something like DenormPreserve)

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
     }
 
+    // opcode: 16
+    case SpvOpExecutionMode: {
+      auto &execution = ir_->entry_points[inst->GetSingleWordInOperand(0u)];
+      auto mode =
+          static_cast<SpvExecutionMode>(inst->GetSingleWordInOperand(1u));
+      execution.flags.set(mode);
+
+      switch (mode) {
```
dneto0: There are over 40 execution modes in the SPIR-V spec, so I expect eventually this will be better broken out into a helper function.

... Hm. But looking at the table in the SPIR-V spec, only a few have parameters, and only a subset of those are usable in Shader modules, and even fewer are (for now) usable in WebGPU.

So actually, it's perfectly fine to keep this inline with the rest of the code.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+        default:
+          break;
+      }
+      break;
+    }
+
+    // opcode: 5
+    case SpvOpName: {
+      uint32_t id = inst->GetSingleWordInOperand(0u);
+      ir_->set_name(id, reinterpret_cast<const char *>(
```
dneto0: We really should make a shortcut for this cast.... :-)

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+      // supported decorations are single uint, so merge decorate and
+      // decorate-id here.
+      // TODO(sarahM0): investigate if this limitation is acceptable.
+      uint32_t id = inst->GetSingleWordInOperand(0u);
+
+      auto decoration =
+          static_cast<spv::Decoration>(inst->GetSingleWordInOperand(1u));
+      if (inst->NumInOperands() > 2) {
+        ir_->meta[id].decoration_word_offset[decoration] =
+            uint32_t(inst->GetInOperand(2u).words.data() - ir_->spirv.data());
```
dneto0: This doesn't look right.

Unfortunately SPIRV-Cross doesn't document what decoration_word_offset means.

In any case, you're taking the difference between pointers in two different allocated buffers.  The spirv-opt Instruction holds onto Operand objects, and each Operand has its own independently allocated vector of words, and the words.data() is pointing at the start of that buffer.

I think you want something like:

      offset_ /* the word offset of this instruction */  + 2

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+
+    // opcode: 72
+    case SpvOpMemberDecorate: {
+      uint32_t id = inst->GetSingleWordInOperand(0u);
+      uint32_t member = inst->GetSingleWordInOperand(1u);
+      auto decoration =
+          static_cast<spv::Decoration>(inst->GetSingleWordInOperand(2u));
+      if (inst->NumInOperands() > 3)
+        ir_->set_member_decoration(id, member, decoration,
+                                   inst->GetSingleWordInOperand(3u));
```
dneto0: Weird that you don't have to set the meta.... offset thing?  Depends on what SPIRV-Cross expects. But please check.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,13 +194,241 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+        type.basetype = spirv_cross::SPIRType::Float;
+      else if (width == 16)
+        type.basetype = spirv_cross::SPIRType::Half;
+      else
+        assert("Unrecognized bit-width of floating point type.");
+      type.width = width;
+      break;
+    }
+
+    // opcode: 23
```
dneto0: ... stopping reviewing here...

libshaderc_spvc/src/spvcir_pass.h
```diff
@@ -40,6 +40,8 @@ class SpvcIrPass : public Pass {
   spirv_cross::ParsedIR *ir_;
   spirv_cross::SPIRFunction *current_function_ = nullptr;
   spirv_cross::SPIRBlock *current_block_ = nullptr;
+  spirv_cross::SmallVector<uint32_t> global_struct_cache;
+  uint32_t offset_ = 5;
```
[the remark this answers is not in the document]
sarahM0: It's a data structure to store type aliases. 
and according to a spirv-cross comment: This must be an ordered data structure so we always pick the same type aliases.

sarahM0:
This pull request will be broken into smaller ones including unit test for each supporting instruction.

PR #853, opened 2019-10-30
Add emit-line-directive to spvc

sarahM0:
Based on #847 
Supports new cases in spvc parser.
Passing 349 out of 1500 tests (23%).
Adding --emit-line-directive to options


sarahM0:
> This looks like this is based off of your other PR. Is it possible to base this off of master instead, so it doesn't include all those changes? Or is this PR a follow up to that PR?

It's a follow up to PR #847 .

sarahM0:
Closed because the branch that it's base of is closed (#847 ).

PR #857, opened 2019-10-31
Add spvcir unit test - OpCapability

sarahM0:


libshaderc_spvc/CMakeLists.txt
```diff
@@ -94,11 +94,14 @@ shaderc_add_tests(
   TEST_PREFIX shaderc
   LINK_LIBS shaderc_spvc ${SPVC_LIBS}
   INCLUDE_DIRS include ${shaderc_SOURCE_DIR}/libshaderc/include ${SPIRV-Cross_SOURCE_DIR}/..
+    ${SPIRV-Cross_SOURCE_DIR}/../effcee
```
dneto0: So far you aren't using these. I would avoid adding them until a test actually needs them.  (We went almost 3 years in SPIRV-Tools without adding this.)

libshaderc_spvc/src/spvcir_test.cc
```diff
@@ -0,0 +1,89 @@
...
+  ir.spirv =
+      std::vector<uint32_t>(binary.data(), binary.data() + binary.size());
+
+  auto result = SinglePassRunAndDisassemble<SpvcIrPass, spirv_cross::ParsedIR*>(
+      JoinAllInsts(input), /* skip_nop = */ true, /* do_validation = */ false,
+      &ir);
+
+  EXPECT_EQ(Pass::Status::SuccessWithoutChange, std::get<1>(result));
+  EXPECT_EQ(ir.declared_capabilities[0], spv::Capability::CapabilityShader);
+  EXPECT_EQ(ir.declared_capabilities[1],
```
dneto0: Good. This is exactly right.

libshaderc_spvc/CMakeLists.txt
dneto0: I'm told one of the fixtures headers uses effcee.  So this actually is needed.

dneto0:
Seems the test file is missing a link dependency.
Once that's fixed and bots are happy I'd be happy to see this accepted.

PR #872, opened 2019-11-06
Add support and unit test for OpTypeInt, OpTypePointer and OpConstant

sarahM0:


libshaderc_spvc/src/spvcir_test.cc
```diff
@@ -95,16 +95,16 @@ class SpvcIrParsingTest : public PassTest<::testing::Test> {
...
-                   "%1 = OpFunction %2 None %3",
-                   "%4 = OpLabel",
-                   "OpReturn",
-                   "OpFunctionEnd"};
+              "%2 = OpTypeVoid",
+              "%3 = OpTypeFunction %2",
+              "%1 = OpFunction %2 None %3",
+              "%4 = OpLabel",
+              "OpReturn",
+              "OpFunctionEnd"};
```
zoddicus: nit, the convention I have seen for spirv asm has been to have the opcode vertically alligned., so the '%4 = ' sticks out on the left.

If it is clang-format that is moving the lines around, you can use // clang-format off and // clang-format on before and after the region you don't want it to muck around with

sarahM0: Thanks!

PR #873, opened 2019-11-07
Add spvc parser support and unit tests for OpVariable, OpDecorate, OpExtInstImport, … …

sarahM0:


libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,6 +162,36 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+      // supported decorations are single uint, so merge decorate and
+      // decorate-id here.
+      // TODO(sarahM0): investigate if this limitation is acceptable.
+      uint32_t id = inst->GetSingleWordInOperand(0u);
+
+      auto decoration =
+          static_cast<spv::Decoration>(inst->GetSingleWordInOperand(1u));
+      if (inst->NumInOperands() > 2) {
+        // instruction offset + length = offset_ + 1 + inst->NumOpreandWords()
+        assert(offset_ + 1 + inst->NumOperandWords() < ir_->spirv.size() &&
```
zoddicus: Is this condition possible for a validly parsed shader or does it being violated indicate a bug in the parsing?

If it is possible for a valid shader to violate this assert, then there should be some sort of guard logic down below, since on non-debug builds the assert will be compiled out. If it is just a sanity check on the parser, then it is fine to leave as an assert

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -160,6 +225,37 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+    case SpvOpVariable: {
+      uint32_t type = inst->type_id();
+      uint32_t id = inst->result_id();
+      auto storage =
+          static_cast<spv::StorageClass>(inst->GetSingleWordInOperand(0u));
+      uint32_t initializer =
+          inst->NumInOperands() == 2 ? inst->GetSingleWordInOperand(1u) : 0;
+
+      if (storage == spv::StorageClassFunction) {
+        assert(current_function_ && "No function currently in scope");
```
zoddicus: Same question as above about asserts.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -257,9 +353,30 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
       break;
     }
 
-    default:
+    case SpvOpStore: {
+      assert(current_block_ && "Currently no block to insert opcode.");
```
zoddicus: ditto

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -257,9 +353,30 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+      spirv_cross::Instruction instr = {};
+      instr.op = inst->opcode();
+      instr.count = inst->NumOperandWords() + 1;
+      instr.offset = offset_ + 1;
+      instr.length = instr.count - 1;
+
+      assert(
```
zoddicus: ditto

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -257,9 +353,30 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+          instr.count != 0 &&
+          "SPIR-V instructions cannot consume 0 words. Invalid SPIR-V file.");
+      assert(offset_ <= ir_->spirv.size() &&
```
zoddicus: ditto

libshaderc_spvc/src/spvcir_pass.cc
sarahM0: It's a sanity check (having safe guards for valid shaders violations is very good point and I'm going to keep it mind).

sarahM0:
> High level comment:
> I suspect in the long term we are going to want to have some sort of mechanism to return success/failure out of GenerateSpirvCrossIR, so that if we hit a case like an unsupported instruction or value for an operand the processing can be halted there and the failure propegated out to spvc to halt the entire process.
> 
> Related, there are a lot of asserts in the code so far and in this PR. I think with a proper reporting mechanism some of those could be converted to error returns.

Fair point. I changed the mechanism to return Pass::Status with proper error message. Please take a second look (if you could think of better names for CheckAndSetErrorMessage function, please let me know).

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -97,6 +120,43 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+            id, spirv_cross::SPIRExtension::SPV_AMD_gcn_shader);
+      else {
+        // spirv-cross comment:
+        // Other SPIR-V extensions which have ExtInstrs are currently not
+        // supported.
+        // TODO(sarahM0): figure out which ones are not supported and try to add
+        // them.
+        CheckAndSetErrorMessage(
+            false,
+            "SpvcIrPass: Error while parsing OpExtInstImport, SPIRV extension "
```
zoddicus: Should have the unsupported extension string in the error message.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -23,25 +23,46 @@ SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR *ir) {
...
 
   auto s = ir_->spirv.data();
-  assert(ir_->spirv.size() > 3 && "spirv data is too small");
+  CheckAndSetErrorMessage(ir_->spirv.size() > 3,
+                          "SpvcIrPass: spirv data is too small");
   uint32_t bound = s[3];
   ir_->set_id_bounds(bound);
 }
 
+void SpvcIrPass::CheckAndSetErrorMessage(bool condition, std::string message) {
```
zoddicus: Maybe CheckConditionAndSetErrorMessage

zoddicus: Also maybe also return a bool that is condition, so that you can write thing like 

`if (!CheckConditonAndSetErrorMessage(foo, "I prefer bar over foo") return;`

zoddicus: Also probably work putting a comment on this method to say what it is trying to do at a high-level.

`// Check boolean condition. If false record the error message, and set pass status to failure to stop processing.`

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -23,25 +23,46 @@ SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR *ir) {
...
+    status_ = Status::Failure;
+    if (consumer()) {
+      consumer()(SPV_MSG_ERROR, 0, {0, 0, 0}, message.c_str());
+    } else {
+      assert(condition && message.c_str());
+    }
+  }
+}
+
 Pass::Status SpvcIrPass::Process() {
```
zoddicus: If you don't change the constructor, you need to check that status_ isn't Failure at the top of this function.

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -23,25 +23,46 @@ SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR *ir) {
   ir_ = ir;
   current_function_ = nullptr;
   current_block_ = nullptr;
+  status_ = Status::SuccessWithoutChange;
 
   auto s = ir_->spirv.data();
```
zoddicus: I preferential to not doing any work/processing in the constructor, I would move this to beginning of Process().

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -84,8 +105,10 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
 
     case SpvOpCapability: {
       auto cap = inst->GetSingleWordOperand(0u);
-      assert(cap != spv::CapabilityKernel &&
-             "Kernel capability not supported.");
+
+      CheckAndSetErrorMessage(cap != spv::CapabilityKernel,
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -127,6 +187,37 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+      // supported decorations are single uint, so merge decorate and
+      // decorate-id here.
+      // TODO(sarahM0): investigate if this limitation is acceptable.
+      uint32_t id = inst->GetSingleWordInOperand(0u);
+
+      auto decoration =
+          static_cast<spv::Decoration>(inst->GetSingleWordInOperand(1u));
+      if (inst->NumInOperands() > 2) {
+        // instruction offset + length = offset_ + 1 + inst->NumOpreandWords()
+        CheckAndSetErrorMessage(
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -153,13 +244,48 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
       auto id = inst->result_id();
       auto type = inst->GetSingleWordInOperand(1u);
 
-      assert(!current_function_ &&
-             "Must end a function before starting a new one!");
+      CheckAndSetErrorMessage(
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -153,13 +244,48 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+    case SpvOpVariable: {
+      uint32_t type = inst->type_id();
+      uint32_t id = inst->result_id();
+      auto storage =
+          static_cast<spv::StorageClass>(inst->GetSingleWordInOperand(0u));
+      uint32_t initializer =
+          inst->NumInOperands() == 2 ? inst->GetSingleWordInOperand(1u) : 0;
+
+      if (storage == spv::StorageClassFunction) {
+        CheckAndSetErrorMessage(current_function_,
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -220,7 +346,9 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
       // Blocks
     case SpvOpLabel: {
       // OpLabel always starts a block.
-      assert(current_function_ && "Blocks cannot exist outside functions!");
+      CheckAndSetErrorMessage(current_function_,
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -229,18 +357,22 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
       if (!current_function_->entry_block) {
         current_function_->entry_block = id;
       }
-      assert(!current_block_ &&
-             "Cannot start a block before ending the current block.");
+      CheckAndSetErrorMessage(!current_block_,
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -229,18 +357,22 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
       current_block_ = &set<spirv_cross::SPIRBlock>(id);
       break;
     }
 
     case SpvOpReturn: {
-      assert(current_block_ && "Trying to end a non-existing block.");
-      // TODO (sarahM0): refactor this into TerminateBlock( ... terminator type
-      // ...), which also resets current_block_ to nullptr. Once having one more
-      // terminator case.
+      CheckAndSetErrorMessage(current_block_,
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -249,17 +381,44 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
 
     case SpvOpFunctionEnd: {
       // Very specific error message, but seems to come up quite often.
-      assert(!current_block_ &&
-             "Cannot end a function before ending the current block.\n"
-             "Likely cause: If this SPIR-V was created from glslang HLSL, "
-             "make sure the entry point is valid.");
+      CheckAndSetErrorMessage(
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -249,17 +381,44 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+          "function before ending the current block.\n"
+          "Likely cause: If this SPIR-V was created from glslang HLSL, "
+          "make sure the entry point is valid.");
       current_function_ = nullptr;
       break;
     }
 
-    default:
+    case SpvOpStore: {
+      CheckAndSetErrorMessage(
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -249,17 +381,44 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+          current_block_, "SpvcIrPass: Currently no block to insert opcode.");
+      spirv_cross::Instruction instr = {};
+      instr.op = inst->opcode();
+      instr.count = inst->NumOperandWords() + 1;
+      instr.offset = offset_ + 1;
+      instr.length = instr.count - 1;
+
+      CheckAndSetErrorMessage(instr.count != 0,
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -249,17 +381,44 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+                              "SpvcIrPass: SPIR-V instructions cannot consume "
+                              "0 words. Invalid SPIR-V file.");
+      CheckAndSetErrorMessage(
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -249,17 +381,44 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
...
+          offset_ <= ir_->spirv.size(),
+          "SpvcIrPass: SPIR-V instruction goes out of bounds.");
+      current_block_->ops.push_back(instr);
       break;
+    }
+
+    default: {
+      CheckAndSetErrorMessage(
```
zoddicus: I would suggest returning immediately here if the condition is false

libshaderc_spvc/src/spvcir_test.cc
```diff
@@ -374,6 +411,92 @@ TEST_F(SpvcIrParsingTest, SpvOpTypePointer) {
...
+
+  auto spir_variable = maybe_get<spirv_cross::SPIRVariable>(17, &ir);
+  ASSERT_NE(spir_variable, nullptr);
+  EXPECT_EQ(spir_variable->basetype, static_cast<spirv_cross::TypeID>(16));
+  EXPECT_EQ(spir_variable->storage, spv::StorageClass::StorageClassOutput);
+  EXPECT_EQ(spir_variable->initializer, static_cast<spirv_cross::TypeID>(0));
+  EXPECT_EQ(spir_variable->basevariable, static_cast<spirv_cross::TypeID>(0));
+}
+
+TEST_F(SpvcIrParsingTest, OpDecrateInstruction) {
```
zoddicus: nit: Decorate

libshaderc_spvc/src/spvcir_test.cc
```diff
@@ -374,6 +411,92 @@ TEST_F(SpvcIrParsingTest, SpvOpTypePointer) {
...
+  auto spir_decoration = ir.find_meta(17);
+  ASSERT_NE(spir_decoration, nullptr);
+  auto& work_offsets = spir_decoration->decoration_word_offset;
+  auto itr = work_offsets.find(spv::DecorationLocation);
+  EXPECT_NE(itr, end(work_offsets));
+  auto& work_offset = itr->second;
+  EXPECT_EQ(work_offset, 28);
+}
+
+TEST_F(SpvcIrParsingTest, OpDecrateInstruction2) {
```
zoddicus: nit: Decorate

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -24,38 +24,43 @@ SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR *ir) {
...
+bool SpvcIrPass::CheckConditionAndSetErrorMessage(bool condition,
+                                                  std::string message) {
   if (!condition) {
     status_ = Status::Failure;
     if (consumer()) {
       consumer()(SPV_MSG_ERROR, 0, {0, 0, 0}, message.c_str());
     } else {
       assert(condition && message.c_str());
     }
+    return false;
```
zoddicus: Instead of the return false; } else { return true;}, you could just unconditional return condition;

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -24,38 +24,43 @@ SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR *ir) {
...
+  ir_->set_id_bounds(bound);
   get_module()->ForEachInst(
       [this](Instruction *inst) {
         if (status_ != Status::SuccessWithoutChange) return;
         GenerateSpirvCrossIR(inst);
       },
       true);
 
-  CheckAndSetErrorMessage(
+  CheckConditionAndSetErrorMessage(
```
zoddicus: nit: missed an immediate return

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -24,38 +24,43 @@ SpvcIrPass::SpvcIrPass(spirv_cross::ParsedIR *ir) {
...
       !current_block_,
       "SpvcIrPass: Error at the end of parsing, block was not terminated.");
 
-  CheckAndSetErrorMessage(
+  CheckConditionAndSetErrorMessage(
```
zoddicus: nit: missed an immediate return

libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -149,10 +156,11 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
         // supported.
         // TODO(sarahM0): figure out which ones are not supported and try to add
         // them.
-        CheckAndSetErrorMessage(
-            false,
-            "SpvcIrPass: Error while parsing OpExtInstImport, SPIRV extension "
-            "not supported");
+        CheckConditionAndSetErrorMessage(false,
```
zoddicus: nit: missed an immediate return

PR #886, opened 2019-11-13
Add spvc parser support for all OpConstant.* and OpSpecConstant.*

sarahM0:
Add support for:

- OpConstantFalse
- OpConstantComposite
- OpConstantNull
- OpSpecConstantTrue
- OpSpecConstantFalse
- OpSpecConstantComposite
- OpSpecConstantOp
- OpExtInst
- OpFunctionParameter

### DIFF
--- a/libshaderc_spvc/src/spvcir_pass.cc
+++ b/libshaderc_spvc/src/spvcir_pass.cc
@@ -19,6 +19,47 @@
 namespace spvtools {
 namespace opt {
 
+void SpvcIrPass::make_constant_null(uint32_t id, uint32_t type) {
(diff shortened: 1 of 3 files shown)


libshaderc_spvc/src/spvcir_pass.cc
```diff
@@ -639,6 +862,7 @@ void SpvcIrPass::GenerateSpirvCrossIR(Instruction *inst) {
       break;
     }
 
+    case SpvOpFAdd:
```
zoddicus: Is the case fall through here intentional? If so please add a comment mentioning that.

PR #888, opened 2019-11-14
Android.mk: Avoid polluting NDK build's test tree.

dneto0:
When installing shaderc headers, do that only as a dependency
of libshaderc_combined.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -28,8 +28,6 @@ endef
 
 define gen_libshaderc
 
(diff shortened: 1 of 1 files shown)


dneto0:
Rebased.
Keeping the LGTM.
This was also LGTM's by the NDK team internally.

dneto0:
This was submitted in the NDK internally.
The macos bots are affected by a Python2-Python3 issue.

PR #917, opened 2019-11-29
[WIP] spvc: adding the missing spirv-cross flags

sarahM0:
WIP, do not review.

### DIFF
--- a/spvc/src/main.cc
+++ b/spvc/src/main.cc
@@ -12,25 +12,52 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stdlib.h>
+
+#include <algorithm>
+#include <cassert>
+#include <cstdio>
+#include <cstring>
 #include <fstream>
+#include <functional>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


spvc/src/main.cc
```diff
@@ -98,8 +125,8 @@ const char kBuildVersion[] = ""
     // TODO(bug 653): #include "build-version.inc"
     ;
 
-bool ReadFile(const std::string& path, std::vector<uint32_t>* out) {
-  FILE* file =
+bool ReadFile(const std::string &path, std::vector<uint32_t> *out) {
```
zoddicus: nit: & should live with the type here not the variable name

sarahM0:
For now this task has a low priority (spirv-cross supports more than 60 flags, however Dawn doesn't use any of them). Will be opened again if we find out any of those flags are necessary.

sarahM0:
For now this task has a low priority (spirv-cross supports more than 60 flags, however Dawn doesn't use any of them). Will be opened again if we find out any of those flags are necessary.

Issue #919, opened 2019-12-02
Add missing APIs for HLSL support in Dawn


zoddicus:
Dawn needs the equivalent of the following methods in the API to support HLSL:
set_decoration()

PR #924, opened 2019-12-03
Add missing APIs for HLSL support in Dawn - set_decoration

sarahM0:
fixs #919

### DIFF
--- a/libshaderc_spvc/include/spvc/spvc.h
+++ b/libshaderc_spvc/include/spvc/spvc.h
@@ -28,6 +28,31 @@ extern "C" {
 #include "shaderc/visibility.h"
 
 typedef enum {
+  SHADERC_SPVC_DECORATION_SPECID,
+  SHADERC_SPVC_DECORATION_BLOCK,
+  SHADERC_SPVC_DECORATION_ROWMAJOR,
+  SHADERC_SPVC_DECORATION_COLMAJOR,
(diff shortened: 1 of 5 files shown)


sarahM0:
I'm passing the parameters as uint32_t then casting them into what they should be (static_cast<spirv_cross::ID>(id) and static_cast<spv::Decoration>(decoration)). 
1 - Dawn invokes this function and does not have access to spirv_cross namespace.
2- spvc.hpp does not have access to spv namespace.

Please let me know what you think.

zoddicus:
I am fine with passing the ID as a uint32_t, since at its core that is all an ID is.

spv::Decorations comes from spirv.hpp in spirv-cross, which in turn is copied from spirv-headers. I think we should be using spv::Decorations in the API, instead of an uint32_t.

If spirv-cross was just using vulkan-headers as a dependency instead of copying in the data, I would just say use vulkan-headers. But because it is copying it in, so it may diverge. So we need to expose the version in spirv-cross to the use. I am not sure if we want to expose the actual spirv-cross header file or copy in the data we care about... Let me think on this for a moment.

zoddicus:
For posterity, talked with dj2 and sarahM0 offline. A third option was proposed, having spvc version of spv::Decoration, that only had the values we support, and the internal implementation translates from that in the API to the call to spirv_cross.

sarahM0:
Thanks for your comment @zoddicus .
Ready for review.

PR #931, opened 2019-12-05
Don't assume locations of spirv-tools, effcee, and RE2 are located

dneto0:
Use ${<project>_BINARY_DIR} for built headers for spirv-tools.
Use ${<project>_SOURCE_DIR} for source headers for effcee and RE2.

CMake generates those variables automatically.


dneto0:
My workspace places source trees in unusual places, i.e. different from how the build bots do it.
So I need this fix/generalization to make Shaderc build.

PR #934, opened 2019-12-10
Add missing APIs for GLSL in Dawn

### LINKED ISSUE #921: Add missing APIs for GLSL in Dawn
Dawn needs the equivalent of the following methods in the API to support GLSL:
build_combined_image_samplers()
get_combined_image_samplers()
get_decoration()
set_name()
unset_decoration()

sarahM0:
fix #921

### DIFF
--- a/libshaderc_spvc/include/spvc/spvc.h
+++ b/libshaderc_spvc/include/spvc/spvc.h
@@ -310,6 +310,38 @@ SHADERC_EXPORT shaderc_spvc_status
 shaderc_spvc_compile_shader(const shaderc_spvc_context_t context,
                             shaderc_spvc_compilation_result_t result);
 
+// Get spirv_cross decoration (added for GLSL API support in Dawn)
+// Given an id and a decoration, result is sent out through |argument|
+// if |id| does not exist, returns an error message.
+SHADERC_EXPORT shaderc_spvc_status shaderc_spvc_get_decoration(
+    const shaderc_spvc_context_t context, uint32_t id,
+    shaderc_spvc_decoration decoration, uint32_t* argument);
+
+// Unset spirv_cross decoration (added for GLSL API support in Dawn)
+// Given an id and a decoration. Assuming |id| is valid.
+SHADERC_EXPORT shaderc_spvc_status
+shaderc_spvc_unset_decoration(const shaderc_spvc_context_t context, uint32_t id,
+                              shaderc_spvc_decoration decoration);
+
+// set |name| on a given |id| (added for GLSL API support in Dawn)
+// Assuming |id| is valid.
+SHADERC_EXPORT void shaderc_spvc_set_name(const shaderc_spvc_context_t context,
+                                          uint32_t id, const char* name);
+
+// spirv-cross comment:
+// Analyzes all separate image and samplers used from the currently selected
+// entry point, and re-routes them all to a combined image sampler instead.
+// (added for GLSL API support in Dawn)
+SHADERC_EXPORT void shaderc_spvc_build_combined_image_samplers(
+    const shaderc_spvc_context_t context);
+
+// For each combined_image_sampler calls the provided callback function |f|,
+// passing in three arguments read the combined_image_sampler
+// (added for GLSL API support in Dawn)
+SHADERC_EXPORT void shaderc_spvc_for_each_combined_image_sampler(
+    const shaderc_spvc_context_t context,
+    void (*f)(uint32_t*, uint32_t*, uint32_t*));
+
 // Set spirv_cross decoration (added for HLSL support in Dawn)
 // Given an id, decoration and argument, the decoration flag on the id is set
 // Assuming id is valid.
--- a/libshaderc_spvc/include/spvc/spvc.h
+++ b/libshaderc_spvc/include/spvc/spvc.h
@@ -312,7 +312,7 @@ shaderc_spvc_compile_shader(const shaderc_spvc_context_t context,
 
 // Get spirv_cross decoration (added for GLSL API support in Dawn)
 // Given an id and a decoration, result is sent out through |argument|
-// if |id| does not exist, returns an error message.
+// if |id| does not exist, returns an error.
 SHADERC_EXPORT shaderc_spvc_status shaderc_spvc_get_decoration(
     const shaderc_spvc_context_t context, uint32_t id,
     shaderc_spvc_decoration decoration, uint32_t* argument);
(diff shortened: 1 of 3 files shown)


libshaderc_spvc/include/spvc/spvc.h
```diff
@@ -310,6 +310,38 @@ SHADERC_EXPORT shaderc_spvc_status
 shaderc_spvc_compile_shader(const shaderc_spvc_context_t context,
                             shaderc_spvc_compilation_result_t result);
 
+// Get spirv_cross decoration (added for GLSL API support in Dawn)
+// Given an id and a decoration, result is sent out through |argument|
+// if |id| does not exist, returns an error message.
```
zoddicus: nit: returns an error.

libshaderc_spvc/include/spvc/spvc.h
```diff
@@ -310,6 +310,38 @@ SHADERC_EXPORT shaderc_spvc_status
...
+    const shaderc_spvc_context_t context, uint32_t id,
+    shaderc_spvc_decoration decoration, uint32_t* argument);
+
+// Unset spirv_cross decoration (added for GLSL API support in Dawn)
+// Given an id and a decoration. Assuming |id| is valid.
+SHADERC_EXPORT shaderc_spvc_status
+shaderc_spvc_unset_decoration(const shaderc_spvc_context_t context, uint32_t id,
+                              shaderc_spvc_decoration decoration);
+
+// set |name| on a given |id| (added for GLSL API support in Dawn)
```
zoddicus: nit: capitalize set

libshaderc_spvc/include/spvc/spvc.hpp
```diff
@@ -18,6 +18,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <functional>
```
zoddicus: nit: I don't think this is needed anymore, since you are no longer using std::function

libshaderc_spvc/include/spvc/spvc.hpp
```diff
@@ -361,6 +362,45 @@ class Context {
...
+  // if |id| does not exist, returns an error message.
+  shaderc_spvc_status GetDecoration(uint32_t id,
+                                    shaderc_spvc_decoration decoration,
+                                    uint32_t* argument){
+    return shaderc_spvc_get_decoration(context_.get(), id, decoration, argument);
+  }
+
+  // Unset spirv_cross decoration (added for GLSL API support in Dawn).
+  // Given an id and a decoration. Assuming id is valid.
+  shaderc_spvc_status UnSetDecoration(uint32_t id,
```
zoddicus: nit: Unset instead of UnSet

libshaderc_spvc/include/spvc/spvc.hpp
```diff
@@ -361,6 +362,45 @@ class Context {
     return shaderc_spvc_set_decoration(context_.get(), id, decoration, argument);
   }
 
+  // Get spirv_cross decoration (added for GLSL API support in Dawn).
+  // Given an id and a decoration, result is sent out through |argument|
+  // if |id| does not exist, returns an error message.
+  shaderc_spvc_status GetDecoration(uint32_t id,
+                                    shaderc_spvc_decoration decoration,
+                                    uint32_t* argument){
```
zoddicus: nit: ) { instead of ){. Possibly need to run clang format, since this is occuring below?

libshaderc_spvc/include/spvc/spvc.hpp
```diff
@@ -361,6 +362,45 @@ class Context {
...
+  }
+
+  // Unset spirv_cross decoration (added for GLSL API support in Dawn).
+  // Given an id and a decoration. Assuming id is valid.
+  shaderc_spvc_status UnSetDecoration(uint32_t id,
+      shaderc_spvc_decoration decoration){
+    return shaderc_spvc_unset_decoration(context_.get(), id, decoration);
+  }
+
+// For each combined_image_sampler calls the provided callback function |f|,
```
zoddicus: nit: For each combined_image_sampler, instead of For each combined_image_sampler

libshaderc_spvc/src/spvc.cc
```diff
@@ -338,6 +338,71 @@ shaderc_spvc_status shaderc_spvc_set_decoration(
   return status;
 }
 
+shaderc_spvc_status shaderc_spvc_get_decoration(
+    const shaderc_spvc_context_t context, uint32_t id,
+    shaderc_spvc_decoration decoration, uint32_t* argument_ptr) {
+  uint32_t& argument = *argument_ptr;
```
zoddicus: You should test that argument_ptr is not null.

zoddicus: Also no need to create the LHS argument reference, you can use *argument_ptr later to get and set values.

libshaderc_spvc/include/spvc/spvc.h
```diff
@@ -310,6 +310,38 @@ SHADERC_EXPORT shaderc_spvc_status
...
+// (added for GLSL API support in Dawn)
+SHADERC_EXPORT void shaderc_spvc_build_combined_image_samplers(
+    const shaderc_spvc_context_t context);
+
+// For each combined_image_sampler calls the provided callback function |f|,
+// passing in three arguments read the combined_image_sampler
+// (added for GLSL API support in Dawn)
+SHADERC_EXPORT void shaderc_spvc_for_each_combined_image_sampler(
+    const shaderc_spvc_context_t context,
+    void (*f)(uint32_t*, uint32_t*, uint32_t*));
```
zoddicus: The uint32_t* values I think can just be uint32_t values. Passing ints by value vs reference should be the sme memory usage, and prevents someone doing something silly like writing back a value. I guess technically they should be const uint32_t to make it clear that they are read only values.

libshaderc_spvc/include/spvc/spvc.hpp
sarahM0: For some reason the format short-cut that I have sat-up in my IDE always ignores this file!

zoddicus: Possibly the fact it is a .hpp?

libshaderc_spvc/src/spvc.cc
sarahM0: Yeah. It's a habit intending to make the code look nicer.

zoddicus: I prefer using pointers over references when possible, since the * syntax makes it clear that you writing out of the value out of scope.

libshaderc_spvc/src/spvc.cc
```diff
@@ -381,13 +380,13 @@ shaderc_spvc_status shaderc_spvc_unset_decoration(
 
 inline void shaderc_spvc_for_each_combined_image_sampler(
     const shaderc_spvc_context_t context,
-    void (*f)(uint32_t*, uint32_t*, uint32_t*)) {
+    void (*f)(uint32_t, uint32_t, uint32_t)) {
   for (const auto& combined :
        context->cross_compiler->get_combined_image_samplers()) {
     uint32_t x = combined.sampler_id;
```
zoddicus: nit: There is no need have these temp variables. You can write f(combined.sampler_id, combined.image_id, combined.combined_id);

sarahM0: Yes, that's much better (sorry about these nits, they're left over from from other approaches I tried).

PR #973, opened 2020-01-24
Handle new Glslang profile enum in switch

dneto0:


dneto0:
making an MR ahead of local testing.  I need to see the bots pass.

zoddicus:
Does this need a DEPS roll for gslang also?

dneto0:
> Does this need a DEPS roll for gslang also?

Evidently. :-)

PR #976, opened 2020-01-28
Make a source_set for libshaderc_spvc

kainino0x:
This allows Dawn to include it as a source_set instead of a component,
so Dawn's complete_static_lib setting applies properly:
https://dawn-review.googlesource.com/c/dawn/+/15481

@zoddicus PTAL


kainino0x:
Is this ready to merge? (I can't merge it.)

PR #982, opened 2020-02-02
Fix SPIRV includes location

eclipseo:
SPIRV includes have been moved under glslang/ in the latest version.

Signed-off-by: Robert-André Mauchin <zebob.m@gmail.com>

### DIFF
--- a/libshaderc/src/shaderc.cc
+++ b/libshaderc/src/shaderc.cc
@@ -20,7 +20,7 @@
 #include <sstream>
 #include <vector>
 
-#include "SPIRV/spirv.hpp"
+#include "glslang/SPIRV/spirv.hpp"
 
 #include "libshaderc_util/compiler.h"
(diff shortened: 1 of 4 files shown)


dj2:
This should be using spirv-headers instead of GLSLang for the spirv.hpp file. So, instead of `glslang/SPIRV/` it should change to `spirv/`

dneto0:
I agree Shaderc should be using headers from SPIRV-Headers.

See also https://github.com/KhronosGroup/glslang/issues/1964 for header confusion for Glslang, particularly on case-insensitive filesystems.

dneto0:
I'm reopening, because there is some work to do to properly switch over to using spirv.hpp from SPIRV-Headers.  I'm doing that now.

dneto0:
I think https://github.com/google/shaderc/pull/992 is the right fix.  It gets spirv.hpp from SPIRV-Headers

PR #991, opened 2020-03-02
DEPS: Update SPIRV-Tools

dneto0:
Avoid Python distutils.dir_util


zoddicus:
This I believe was done in https://github.com/google/shaderc/pull/990

PR #992, opened 2020-03-02
Get spirv.hpp from SPIRV-Headers, not Glslang

dneto0:


dneto0:
I haven't looked into the GN build config.

PR #1003, opened 2020-03-17
Remove NV_EXTENSIONS conditionals

dgkoch:
This has been removed from glslang a while ago and is no longer needed.


dgkoch:
cc @dneto0 @s-perron @zoddicus 

This is a precursor for deps to include KHR_ray_tracing support (see https://github.com/KhronosGroup/Vulkan-Docs/issues/1205)

PR #1006, opened 2020-03-19
Update all tools: support provisional ray-tracing

dneto0:


dneto0:
Tested with NDK build on Linux and Windows.

PR #1008, opened 2020-03-20
Extract texture dimension for storage textures

### LINKED ISSUE #1007: Missing the extraction of texture dimension on storage textures
Current SPVC [does not extract the texture dimension on the storage textures](https://github.com/google/shaderc/blob/master/libshaderc_spvc/src/spvc.cc#L981), which is needed in the validations on the storage textures inside Dawn.

@kainino0x  @Kangz @zoddicus PTAL, thanks!

Jiawei-Shao:
Fixes #1007


Kangz:
This is simple enough that I think I can review it. I checked there were no tests added in past and similar commits to SPVC.

PR #1011, opened 2020-03-23
SPVC: Record if a storage texture is declared as multisampled texture

Jiawei-Shao:


Jiawei-Shao:
@Kangz @zoddicus sorry that I missed this before. PTAL, thanks!

Kangz:
Aren't storage textures disallowed from being multisampled? Can't hurt to set the info though.

Kangz:
I don't think we have decided where validation corresponding to the "Shader Interface" [section of Vulkan](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#interfaces) should live. Either way sounds fine for now.

Jiawei-Shao:
@Kangz @zoddicus 

According to https://github.com/KhronosGroup/Vulkan-Docs/issues/1015:
SPIR-V shader modules can declare the ```StorageImageMultisample``` capability, can declare ```OpTypeImage``` with ```Sampled```=2 (storage image) and ```MS```=1 (multisample), and can pass a ```Sample``` operand to ```OpImageRead``` and ```OpImageWrite``` to read/write a specific sample.

And I have tried that SPIRV-Cross allows declaring an ```image2DMS``` in the GLSL shader as multisampled storage images. Multisampled storage textures cannot be supported on D3D12, so we should report an error when a multisampled storage image (```image2DMS```) is declared in the shader.

Jiawei-Shao:
The unit test added in this patch (https://dawn-review.googlesource.com/c/dawn/+/17600) cannot pass without this fix in shaderc.

zoddicus:
@Jiawei-Shao It takes ~4 hours for patches to propagate through Chromium infra, you will have to roll the  shaderc DEP value in your patch once that is done.

Jiawei-Shao:
@zoddicus this commit has been rolled into Chromium. Thanks for your help!

PR #1075, opened 2020-05-28
Update glslang, spirv-headers, spirv-tools

dneto0:
Update SPIRV-Tools to v2020.3


dneto0:
Verified with NDK build on Linux and Windows.

PR #1084, opened 2020-06-12
Add minimum buffer size for dawn reflection

iraiter2:
In dawn, we need to validate buffer sizes for uniform and storage buffers to make sure bound buffers are large enough. [https://github.com/gpuweb/gpuweb/pull/678](https://github.com/gpuweb/gpuweb/pull/678) decided that we should require for runtime arrays at least 1 element, so that we can clamp indices. Luckily, there is a function that does this already. This patch adds to the bindings reflection info the minimum buffer size so we can use it in dawn.

The relevant dawn CL is [https://dawn-review.googlesource.com/c/dawn/+/22421](https://dawn-review.googlesource.com/c/dawn/+/22421), which does the same thing using spirv-cross in ShaderModule.cpp.

The variable name can be changed if another one makes more sense?

### DIFF
--- a/libshaderc_spvc/include/spvc/spvc.h
+++ b/libshaderc_spvc/include/spvc/spvc.h
@@ -242,6 +242,7 @@ typedef struct {
   shaderc_spvc_texture_format_type texture_component_type;
   bool multisampled;
   shaderc_spvc_storage_texture_format storage_texture_format;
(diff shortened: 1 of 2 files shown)


iraiter2:
@googlebot I signed it!

PR #1090, opened 2020-06-26
Add CMake support for buildint Tint in third_party

dneto0:
This is off by default, and not announced in the README.

Requires some Tint changes in order to build:
https://dawn-review.googlesource.com/c/tint/+/24020


zoddicus:
Going to re-run the bots, since they look like infra failure.

PR #1101, opened 2020-07-10
Improve how GN finds auto-generated SPIRV-Tools headers

hujiajie:
SPIRV-Tools and Shaderc are no longer enforced to be placed side by side, because priori of the SPIRV-Tools path can be avoided now. This would allow projects using GN to have a directory hierarchy different from Chromium and Dawn.


zoddicus:
Reflexively added kokoro:run, but this patch doesn't need it since those bots don't test the BUILD.gn builds. I have manually tested this patch.

PR #1108, opened 2020-07-17
Mingw - use `posix` toolchain flavor

ben-clayton:
The default is now the `win32` flavor, which is not built with GLIBC threading support in the standard ubuntu package build. This results in `std::mutex` is undefined errors.

Fixes mingw Kokoro continuous builds.


ben-clayton:
Does the Android NDK build use CMake / kokoro bots?
If not, we should be fine.

PR #1109, opened 2020-07-20
Not for review: Tickle kokoro bots

ben-clayton:


ben-clayton:
New Windows Kokoro VM instances appear to be working.

PR #1110, opened 2020-07-20
glslc: Add -mfmt=wgsl output format option

dneto0:
This option is valid only when compiling source, rather than just
preprocessing.

Only works when glslc has been built with WGSL output support,
depending on the "tint" third party library.
Build with CMake option SHADERC_ENABLE_WSGL_OUTPUT=ON


dneto0:
cc: @afd

dneto0:
TODO: update the README and online help once this stabilizes

PR #1119, opened 2020-07-23
Update components

dneto0:
- Glslang
  - including SPV_EXT_shader_atomic_float
- SPIRV-Tools v2020.4
  - including SPV_KHR_terminate_invocation
- SPIRV-Headers
- Shaderc v2020.2
- Effcee v2019.1


dneto0:
Verified with an NDK build on Linux and Windows.

PR #1139, opened 2020-09-28
known-good:  Update tooling

dneto0:
spirv-headers - top of tree 2020-09-28
glslang - top of tree 2020-09-28
spirv-tools - v2020.5
shaderc - v2020.3

effcee - still at v2019.1


dneto0:
Tested on Linux, and as part of an Android NDK build on Linux and Windows.

PR #1141, opened 2020-10-25
Handle alias libraries in shaderc_combined to avoid duplicated symbols

antiagainst:
When getting the transitive libraries needed for shaderc_combined,
we just deduplicate according to their names. Aliased libraries can
have different names. They can all be packed into shaderc_combined,
which will cause duplicated symbols.

I hit this in https://github.com/google/shaderc-rs, because now in
SPIRV-Tools we have defined the SPIRV-Tools library to alias to
SPIRV-Tools-static.


zoddicus (approved):
lgtm

PR #1147, opened 2020-11-11
WIP:  BEFORE suspiciously stack-hungry change in SPIRV-Tool

dneto0:
I just want to see the bots run.


dneto0:
This is no longer needed for testing.

PR #1148, opened 2020-11-11
WIP: AFTER suspiciously stack-hungry validator change in SPIRV-Tools

dneto0:
I just want to see the bots run.


dneto0:
I'm bisecting the failure using #1147

PR #1151, opened 2020-11-12
copyright check: Exclude .venv in source tree

### LINKED ISSUE #1137: Build fails if python venv is a project local directory
## What
Build fails with a plethora of copyright messages such as 
`./.venv/lib/python3.8/site-packages/cmake/data/share/cmake-3.18/Modules/CMakeCompilerABI.h  has no copyright message.`

## Where
Debian Linux X86_64, Python 3.8.5, Virtual environment inside `$PROJECT_DIR/.venv`

## Why
Apparently the script add_copyright.py tries to indiscriminately check for copyright header inside all directories and files

## Potential Fix
Within the same script, the function filtered_descendants can be modified to detect virtual environment
```
def filtered_descendants(glob):
    """Returns glob-matching filenames under the current directory, but skips
    some irrelevant paths."""
    filter_list = ['third_party', 'external', 'build*', 'out*', 'CompilerIdCXX']
    venv_dir = os.environ.get('VIRTUAL_ENV', None)
    if venv_dir:
        filter_list.append(os.path.basename(venv_dir))
    return find('.', glob, filter_list)
```

dneto0:
Fixes #1137

### DIFF
--- a/utils/add_copyright.py
+++ b/utils/add_copyright.py
@@ -62,7 +62,7 @@ def filtered_descendants(glob):
     """Returns glob-matching filenames under the current directory, but skips
     some irrelevant paths."""
     return find('.', glob, ['third_party', 'external', 'build*', 'out*',
-                            'CompilerIdCXX'])
+                            'CompilerIdCXX', '.venv'])
 
 
 def skip(line):


dneto0:
I went with this fix, rather than migrating fully to check-license because check-license would add yet another project dependency.

zoddicus (approved):

PR #1157, opened 2020-12-03
Roll dependencies, supporting Vulkan raytracing

dneto0:


dneto0:
Tested with the Android NDK build on Linux and Windows.
Tested standalone on Linux.

PR #1169, opened 2021-01-29
Improve file+line parsing from Glslang messages

### LINKED ISSUE #1153: Inconsistent error message when compiling files with colons in their names
With files `foo` and `f:o` both with this invalid content:

```
float3 main(){
  return a;
}
```

Observe that the error messages are formatted differently:

```
$ glslc -fshader-stage=frag -x hlsl foo -o foo.spv
foo:2: error: 'a' : unknown variable
foo:2: error: 'return' : type does not match, or is not convertible to, the function's return type
2 errors generated.

$ glslc -fshader-stage=frag -x hlsl f:o -o foo.spv
f:o: error: f:o:2: 'a' : unknown variable
f:o: error: f:o:2: 'return' : type does not match, or is not convertible to, the function's return type
2 errors generated.
```

In the former, the line number appears before `error` and in the latter it appears after `error` and the filename has been prepended to each line.

dneto0:
Fixes #1153

### DIFF
--- a/libshaderc_util/src/message.cc
+++ b/libshaderc_util/src/message.cc
@@ -46,7 +46,7 @@ MessageType DeduceMessageType(string_piece* message) {
 
 // Deduces a location specification from the given message. A location
 // specification is of the form "<source-name>:<line-number>:" and a trailing
-// spaace.  If the deduction is successful, returns true and updates source_name
+// space.  If the deduction is successful, returns true and updates source_name
 // and line_number to the deduced source name and line numer respectively. The
 // prefix standing for the location specification in message is skipped.
 // Otherwise, returns false and keeps all parameters untouched.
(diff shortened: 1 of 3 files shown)


libshaderc_util/src/message.cc
```diff
@@ -45,41 +45,72 @@ MessageType DeduceMessageType(string_piece* message) {
 }
 
 // Deduces a location specification from the given message. A location
-// specification is of the form "<source-name>:<line-number>:". If the deduction
-// is successful, returns true and updates source_name and line_number to the
-// deduced source name and line numer respectively. The prefix standing for the
-// location specification in message is skipped. Otherwise, returns false and
-// keeps all parameters untouched.
+// specification is of the form "<source-name>:<line-number>:" and a trailing
+// spaace.  If the deduction is successful, returns true and updates source_name
```
antiagainst, line 49: typo: spaace

antiagainst (approved):

zoddicus (approved):

Issue #303, opened 2017-02-13
Publish Blaze/Bazel build files


abergmeier-dsfishlabs:
Since this is a Google project I am somewhat sure that internally, you have BUILD files for building all this. Please make them available to the public.

dneto0:
Sorry, I don't know of any Blaze/BUILD files for Shaderc.
I'd be happy to accept a contribution in this area, but would rely on the community to verify its correctness.

abergmeier-dsfishlabs:
If shaderc is used internally at google, how is it integrated with the rest of Blaze based projects, then?

dneto0:
You seem to be making certain assumptions, without evidence.

My team maintains Shaderc in the service of the graphics community, with a focus on Vulkan.  I'm glad you're interested in it.   We also ship Shaderc in the Android NDK, for use by Android developers specifically.

Again, I'd gladly accept contributions in this area.

abergmeier-dsfishlabs:
> You seem to be making certain assumptions

Sorry. Last information I had was that ANDROID NDK is being built with Bazel by now.

dneto0:
The Android NDK build is described in https://android.googlesource.com/platform/ndk/+/master/README.md

If you trace down through the NDK's checkbuild.py script, you'll see the "shader-tools" module. That's what builds Shaderc.  It invokes the cmake flow to build host executables.  Shaderc is also delivered as source libraries with accompanying Android.mk files for integration directly into Android applications.

dnovillo:
I'll be working on this in the coming days.

dadrian:
I started creating Bazel build definitions for personal use and encountered the following issue: https://github.com/bazelbuild/bazel/issues/11672

Basically, Bazel mis-includes files if their name collides with a Windows SDK file, and both the local file and SDK file are needed to build. For shaderc, this means that `io.h` in `libshaderc_util` doesn't get included properly on Windows, and as a result, the targets fail to build.

This might be able to be addressed within Bazel, but an alternative solution would be to just rename `io.h` to something else, e.g. `io_shaderc.h`. Would the maintainers of shaderc be open to such a patch? If so, I'd be happy to provide it, and provide my BUILD file.

My Bazel BUILD file is available [here](https://github.com/bazelbuild/bazel/files/4848626/shader.build.txt).

PR #1173, opened 2021-03-04
Rename io.h to io_shaderc.h

dadrian:
This makes libshaderc possible to compile on Windows under Bazel, due to https://github.com/bazelbuild/bazel/issues/11672.


dadrian:
I tested this against my own BUILD file, see https://github.com/google/shaderc/issues/303#issuecomment-785495366. Once I clean that happy, happy to make a follow-on PR to land the BUILD file. I don't think I have all components / features building with it yet.

zoddicus:
@dneto0 This seems reasonable to me, thoughts?

dneto0 (approved):
LGTM.  That library is an internal inteface of Shaderc (despite being visible to glslc).

PR #1179, opened 2021-04-07
Added a way of setting an implicit set to the compiler options.

AlvaroBarua:


AlvaroBarua:
@googlebot I signed it!

zoddicus:
I just noticed that this never was reviewed.

@dneto0 when you get back to the office can you take a look at it?

AlvaroBarua:
Hi, just remembered I added this pull request. I was wondering if it was reviewed?

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -328,6 +328,23 @@ class Compiler {
...
+    for (auto stage : stages()) {
+      SetImplicitSetForStage(stage, set);
+    }
+  }
+
+  // Sets an implicit set for the given GLSL shader stage if no set has been provided in
+  // the qualifier
+  void SetImplicitSetForStage(Stage stage, uint32_t set)
+  {
+    hlsl_explicit_bindings_[static_cast<int>(stage)].clear();
```
dneto0, line 344: This interferes with the operation of SetHlslRegisterSetAndBindingForStage

Also, glslang only operates on these options in triples.
See https://github.com/KhronosGroup/glslang/blob/b2d2c9dd3d77c6adadff37309aa58a181bafc5f9/glslang/HLSL/hlslParseHelper.cpp#L6388

dneto0 (changes requested):
I am skeptical that this works.

Also, there are no tests.

Issue #1180, opened 2021-04-15
Feature Request: expose glslang's option to combine separate HLSL Texture*D and SamplerState objects


Rob2309:
A while ago, GLSLang added functionality to automatically combine separate Textures and Samplers into SampledImage SPIRV objects (See this [issue](https://github.com/KhronosGroup/glslang/issues/591) and this [PR](https://github.com/KhronosGroup/glslang/pull/856)). 

It would be very useful if this functionality is exposed in shaderc, as this would allow compiling HLSL shaders into an optimal SPIRV assembly without using SPIRV-cross to combine the texture and sampler objects.

I could implement this myself, if anyone can tell me what exactly I have to do to comply with section 4b) of the Apache License, which states that any modified files have to carry prominent notice. Does this apply to pull requests as well?

PR #1181, opened 2021-04-19
Exposed GLSLang's option to combine textures and samplers automatically.

Rob2309:
This pull request implements #1180 by exposing the GLSLang option TShader::setTextureSamplerTransformMode(). I am not quite sure how to name the respective functions, as I don't think setTextureSamplerTransformMode() really describes what the function does, so naming ideas are welcome.

The "MODIFIED BY  Robin Quint" notices are there to comply with Apache §4b, if that is not required, I can remove them quickly.

### DIFF
--- a/glslc/src/main.cc
+++ b/glslc/src/main.cc
@@ -1,4 +1,5 @@
 // Copyright 2015 The Shaderc Authors. All rights reserved.
+// MODIFIED BY Robin Quint
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -307,6 +308,8 @@ int main(int argc, char** argv) {
       }
     } else if (arg == "-fauto-bind-uniforms") {
       compiler.options().SetAutoBindUniforms(true);
+    } else if (arg == "-fupgrade-textures") {
+      compiler.options().SetUpgradeTextures(true);
     } else if (arg == "-fauto-map-locations") {
       compiler.options().SetAutoMapLocations(true);
     } else if (arg == "-fhlsl-iomap") {
--- a/libshaderc/include/shaderc/shaderc.h
+++ b/libshaderc/include/shaderc/shaderc.h
@@ -1,4 +1,5 @@
 // Copyright 2015 The Shaderc Authors. All rights reserved.
+// MODIFIED BY Robin Quint
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -406,6 +407,9 @@ SHADERC_EXPORT void shaderc_compile_options_set_limit(
 // that aren't already explicitly bound in the shader source.
 SHADERC_EXPORT void shaderc_compile_options_set_auto_bind_uniforms(
     shaderc_compile_options_t options, bool auto_bind);
+
+SHADERC_EXPORT void shaderc_compile_options_set_upgrade_textures(
+    shaderc_compile_options_t options, bool upgrade);
 
 // Sets whether the compiler should use HLSL IO mapping rules for bindings.
 // Defaults to false.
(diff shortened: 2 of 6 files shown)


Rob2309:
@googlebot I signed it!

glslc/src/main.cc
```diff
@@ -1,4 +1,5 @@
 // Copyright 2015 The Shaderc Authors. All rights reserved.
+// MODIFIED BY Robin Quint
```
dneto0, line 2: this isn't necessary, please remove.
By committing to this repo you become part of the collective of "Shaderc Authors".
If you like you can add yourself to the CONTRIBUTORS file in the root of the source tree.  But authorship is determined by the git history.

glslc/src/main.cc
```diff
@@ -307,6 +308,8 @@ int main(int argc, char** argv) {
       }
     } else if (arg == "-fauto-bind-uniforms") {
       compiler.options().SetAutoBindUniforms(true);
+    } else if (arg == "-fupgrade-textures") {
```
dneto0, line 311: I found the original commit of this feature in Glslang.
https://github.com/KhronosGroup/glslang/pull/856

And I see you're adding this to the glslangValidator command line via https://github.com/KhronosGroup/glslang/pull/2616 

Can I suggest -fauto-combined-image-sampler instead?  That would also be more in line with what I suggested for glslangValidator.

Also, please rename the other methods and state variables to match.

dneto0: Also, please update the glslc help string. I suggest:


  -fauto-sampled-textures    Removes sampler variables and converts existing textures to combined image-sampler

libshaderc/include/shaderc/shaderc.h
```diff
@@ -407,6 +408,9 @@ SHADERC_EXPORT void shaderc_compile_options_set_limit(
 SHADERC_EXPORT void shaderc_compile_options_set_auto_bind_uniforms(
     shaderc_compile_options_t options, bool auto_bind);
 
+SHADERC_EXPORT void shaderc_compile_options_set_upgrade_textures(
```
dneto0, line 411: This needs a method comment.

libshaderc/include/shaderc/shaderc.hpp
```diff
@@ -274,6 +275,10 @@ class CompileOptions {
     shaderc_compile_options_set_auto_bind_uniforms(options_, auto_bind);
   }
 
+  void SetUpgradeTextures(bool upgrade) {
```
dneto0, line 278: this needs a method comment.

libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -277,6 +279,8 @@ class Compiler {
   // uniform variables that don't have explicit bindings.
   void SetAutoBindUniforms(bool auto_bind) { auto_bind_uniforms_ = auto_bind; }
 
+  void SetUpgradeTextures(bool upgrade) { upgrade_textures_ = upgrade; }
```
dneto0, line 282: This needs a method comment.

dneto0 (changes requested):
Also, please add a glslc-based test to prove the command line option works.
I suggestion glslc/test/option_fauto_combined_image_sampler.py
You can follow a pattern like in option_fauto_map_ocations.py, where you can match against an expected substring in the disassembled output. 
E.g. see FAutoMapLocationsGeneratesLocationsCheckOutput0

Rob2309:
Hopefully everything is resolved now :) 
All functions are renamed and added comments where appropriate. The unnecessary "MODIFIED BY" clauses were removed. 
I am not quite sure about the test though, since the "glslc_tests" always fail for me with `ctest -C Release`, even before adding any testcases, so I am not sure whether my test cases actually work. Is there any file where more verbose test output is shown?

dneto0:
Thanks for the update!

> I am not quite sure about the test though, since the "glslc_tests" always fail for me with ctest -C Release

Hm.  The glslc_tests are all file based and there's logic somewhere for finding the source files.  I wonder if it's broken due to an unexpected relative path.  
It's a small enough thing that I can add the test later.

PR #1193, opened 2021-05-22
glslc: main.cc: update help: debug info disclaimer

pavel-the-best:
As it appears, the option -g actually has some effect now.


pavel-the-best:
@googlebot I signed it!

zoddicus (approved):

zoddicus (changes requested):
Looking at the Kokoro failures, it is looks like there is a test, HelpParameters, that is dooing some sort of string matching against this. Can you please update those tests?

pavel-the-best:
Done

pavel-the-best:
@googlebot I signed it!

zoddicus (approved):

PR #1194, opened 2021-05-27
Add explicit copy constructor for string_piece

ianloic:
Implicit copy constructors are deprecated and can cause build warnings
or errors with some compilers.


zoddicus:
I think the license check failing in unrelated, so will approve and land

zoddicus (approved):

PR #1197, opened 2021-06-02
Fix shaderc util test to pass the correct shader_stage

sriramm-nv:
Found this bug while browsing through the test code. It is passing incorrect shader kind for the test using vertex shader.


zoddicus:
This looks good to me, could you please sign the CLA?

sriramm-nv:
@googlebot I signed it!

PR #1211, opened 2021-08-08
Standardised git-sync-deps.py

TheMindVirus:
The edited command before changes is incompatible with Windows. There is no file extension to identify the type of script it is running and no explicit invocation with python. There is no path and version number to be specified at the top of the file or in the command because this can differ between machines. It either works or it doesn't. If it doesn't, the machine is not tolerant to new standards and needs an update or the repository repackaging.


zoddicus (approved):
lgtm, will merge after kokoro has run.

zoddicus (changes requested):
It is essential to change the file to ./utils/git-sync-dep.py, or is just adding python to the README.md sufficient?

It looks like our Windows kokoro bot is happy with using `python utils\git-sync-deps`

There are a number of places, scripts and configs where the file name would need to be updated:
Dockerfile:RUN ./utils/git-sync-deps
kokoro/android-release/build.sh:./utils/git-sync-deps
kokoro/linux/build-docker.sh:./utils/git-sync-deps
kokoro/macos/build.sh:./utils/git-sync-deps
kokoro/ndk-build/build.sh:./utils/git-sync-deps
kokoro/windows/build.bat:python utils\git-sync-deps
license-checker.cfg:                    "utils/git-sync-deps",
license-checker.cfg:            { "include": [ "utils/git-sync-deps" ] }

zoddicus:
Looks like the bots are failing with `ImportError: No module named builtins`, which is normally from trying to run python3 code in the python2, so probably there is some path issues going on for the bots.

I am not really an expert in the Kokoro bots, so I don't really want to go mucking about and potentially break them.

@dneto0 when you are back, can you take a look at this?

zoddicus:
or possibly @ben-clayton

ben-clayton:
You can try running the python script with `python3`

TheMindVirus:
Please may I draw your attention to this line of configuration:
https://github.com/google/shaderc/blob/fadb0edb247a1daa74f9a206a27e9a1c0417ce49/kokoro/windows/build.bat#L24
```
:: Force usage of python 3.6.
set PATH=C:\python36;%PATH%
```
We are on Python 3.9.6 now. Unless it is absolutely necessary to make this change every billionth minor version,
I suggest avoiding the forcing of version numbers in commands for software packages across the board.

This is so that the package doesn't automatically break when a parent, sibling or dependent repository decides to commit.
If intolerances with versions must be dealt with urgently, then it may on occasion be necessary to "hold back" versions.

To hold back versions, the command must still remain the same while the version number is changed (or indeed kept the same)
in a metadata file. This will come to the aid of a manual installer or also an artificially intelligent version scanning system.

PR #1216, opened 2021-09-20
build: add switch to disable copyright check

lunarpapillo:
A Vulkan SDK build of shaderc fails on the copyright check:

    C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microsoft.
    CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with co
    de 1. [C:\j\msdk\build\shaderc\repo\src\dbuild32\check-copyright.
    vcxproj]

Analysis in Visual Studio reveals that the problem is:

    20>  .\dbuild\include\glslang\build_info.h  has no copyright mess
    age.
    20>C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V140\Microso
    ft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with
    code 1.

"build_info.h" does indeed have a copyright message:

    // Copyright (C) 2020 The Khronos Group Inc.

but this fails the copyright check, which is looking solely for:

    Copyright <NNNN> The Shaderc Authors. All rights reserved.

These changes add a SHADERC_SKIP_COPYRIGHT_CHECK option that can be
passed at generation time to disable the copyright check and allows
the build to succeed.

It turns out CMake 3.10 ordered the shaderc build in such a way that
the copyright check occurred at the end of the build, while CMake 3.19
puts the copyright check at the beginning of the build.  So the SDK
*could* be built with CMake 3.10 if the build error was ignored, but
cannot be built with CMake 3.19 even if the build error is ignored.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,16 @@ else()
   message(STATUS "Configuring Shaderc to avoid building examples.")
 endif()
 
+option(SHADERC_SKIP_COPYRIGHT_CHECK "Skip copyright check" ${SHADERC_SKIP_COPYRIGHT_CHECK})
+if(NOT ${SHADERC_SKIP_COPYRIGHT_CHECK})
+  set(SHADERC_ENABLE_COPYRIGHT_CHECK ON)
+endif()
+if(${SHADERC_ENABLE_COPYRIGHT_CHECK})
+  message(STATUS "Configuring Shaderc to check copyrights.")
+else()
+  message(STATUS "Configuring Shaderc to avoid checking copyrights.")
+endif()
+
 option(SHADERC_ENABLE_WERROR_COMPILE "Enable passing -Werror to compiler, if available" ON)
 
 set (CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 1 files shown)


lunarpapillo:
@googlebot I signed it!

lunarpapillo:
@dneto0 Hi, David - @greg-lunarg suggested I petition you for a review, as I seem to be unable to request reviewers on my own...

CMakeLists.txt
```diff
@@ -50,6 +50,16 @@ else()
   message(STATUS "Configuring Shaderc to avoid building examples.")
 endif()
 
+option(SHADERC_SKIP_COPYRIGHT_CHECK "Skip copyright check" ${SHADERC_SKIP_COPYRIGHT_CHECK})
+if(NOT ${SHADERC_SKIP_COPYRIGHT_CHECK})
```
dneto0, line 54: Thanks for keeping the check in the default flow.
So I'm ok with doing this.

Another fix would be to update the filtering parameters in https://github.com/google/shaderc/blob/main/utils/add_copyright.py#L64

dneto0 (approved):

CMakeLists.txt
lunarpapillo: Thanks for your time!

There doesn't seem to be an easy way that wouldn't modify the whole algorithm in a way that could have unwanted side effects, or am I missing something?  The current `find()` function cannot filter out particular filenames (it filters out directory names).  To block `.\build\include\glslang\build_info.h` from the check, I'd either have to add the whole `glslang` directory to the `skip_glob_list` parameter (which might be too much), or add a new parameter that allows removal of filenames that match a pattern (which seems to be overkill for a single file, but is feasible, and wouldn't likely have tremendous side effects).

I'm happy to re-implement with a filtering algorithm modification if you prefer.

dneto0: It should have filtered out `\build\` because that matches `build*`.  Is the offending file in `dbuild` ?

In any case, I'm still ok landing this, and will do so.

One other thing comes to mind, which is I usually configure builds to avoid writing into the same tree as the source tree.  But that's not always so easy to control with IDEs, so I won't be fussy about that.

lunarpapillo: > It should have filtered out \build\ because that matches build*. Is the offending file in dbuild ?

Ah, yes, that's it - I missed that detail.  Yes, the failed copyright message was indeed coming from `dbuild`.  So it would have been easy to add 'dbuild*` to the list of ignored directories... I apologize for my errant analysis.

> In any case, I'm still ok landing this, and will do so.

Thank you!

PR #1217, opened 2021-10-12
Fix for missing copyright in glslang

rofrol:
Fix similar to https://github.com/google/shaderc/commit/10530a08bdc5a3fb627746f64bc7e0a8b18819f8

### DIFF
--- a/utils/add_copyright.py
+++ b/utils/add_copyright.py
@@ -62,7 +62,7 @@ def filtered_descendants(glob):
     """Returns glob-matching filenames under the current directory, but skips
     some irrelevant paths."""
     return find('.', glob, ['third_party', 'external', 'build*', 'out*',
-                            'CompilerIdCXX', '.venv'])
+                            'CompilerIdCXX', '.venv', 'glslang'])
 
 
 def skip(line):


utils/add_copyright.py
```diff
@@ -62,7 +62,7 @@ def filtered_descendants(glob):
     """Returns glob-matching filenames under the current directory, but skips
     some irrelevant paths."""
     return find('.', glob, ['third_party', 'external', 'build*', 'out*',
-                            'CompilerIdCXX', '.venv'])
+                            'CompilerIdCXX', '.venv', 'glslang'])
```
dneto0, line 65: The glslang sources should be under third_party, and so they should be covered already.
Have I misunderstood the problem?

dneto0: Ok, I see.
When I build Shaderc in the source tree, then the copyright check fails because Glslang's build generates two header files that of course don't have the Shaderc copyright.

To reproduce:
```bash
git clone --depth 1 https://github.com/google/shaderc
cd shaderc
python utils/git-sync-deps
# Build in the source tree. This is NOT RECOMMENDED
cmake -GNinja .
ninja glslangValidator
ninja check-copyright
```

Then I get an error like this:
```
$ ninja check-copyright
[1/1] Check copyright
FAILED: CMakeFiles/check-copyright /shaderc/CMakeFiles/check-copyright 
python3 utils/add_copyright.py --check
./include/glslang/build_info.h  has no copyright message.
./include/glslang/glsl_intrinsic_header.h  has no copyright message.
ninja: build stopped: subcommand failed.
```

The patch fixes the problem.

Note: When building in the source tree, the copyright check will take a lot longer because it will scan all the object files too.  I strongly recommend building outside the source tree, or in a subdirectory whose name starts with 'build' or 'out'

dneto0 (approved):

dneto0:
Thanks for the contribution!

PR #1226, opened 2021-11-11
Update deps

dneto0:
Shaderc v2021.3
Glslang GitHub master 2021-11-11
SPIRV-Tools v2021.4
SPIRV-Headers GitHub master 2021-11-11


dneto0:
Tested with Android NDK build, and on Linux

PR #1227, opened 2021-11-19
Fix link with NDKr23+

bog-dan-ro:
NDKr23+ removed all but as GNU binutils


dneto0:
I apologize for not getting to this.
I implemented #1234 instead, which uses the recommended Android.mk variable names TARGET_AR and TARGET_STRIP, as recommended by the NDK maintainers.

PR #1230, opened 2021-12-29
Update DEPS to those supporting SPIR-V 1.6

dneto0:
Shaderc itself does not yet support SPIR-V 1.6


dneto0:
(Trying to see if the 'core' file problem in certain Kokoro bots is only exercised by the Shaderc changes.)

dneto0:
Investigating locally via Docker, I find the shaderc_util_file_finder_test is dumping core. ?!

dneto0:
The FileFinderDeathTest.EmptyFilename is the one dumping core.

PR #1243, opened 2022-04-06
Update SPIRV-Tools, SPIRV-Headers, Glslang

dneto0:
Each to top of tree on GitHub


dneto0:
Tested on an NDK build. It passed.

PR #1251, opened 2022-06-20
Add support for 16-bit HLSL types

pablode:
glslang allows the use of 16-bit types in HLSL shaders with the `EShMsgHlslEnable16BitTypes` flag or the `--hlsl-enable-16bit-types` glslangValidator argument.

This PR exposes the functionality as a compile option in the shaderc header and adds a corresponding command-line argument.

PR #1253, opened 2022-06-29
Add support for 16b-bit types in HLSL (with tests)

dneto0:
Builds on #1251


dneto0:
Needs #1255 as well

pablode:
Right - should have updated the docs, too. Good job with the test!

PR #1261, opened 2022-08-11
update deps, finalize v2022.2

dneto0:


dneto0:
The CI-android-arm and CI-android-x86 failures look like this:

```
CMake Error at third_party/glslang/glslang/OSDependent/Unix/CMakeLists.txt:56 (install):
  install TARGETS given no ARCHIVE DESTINATION for static library target
  "OSDependent".

```

Due to: https://github.com/KhronosGroup/glslang/pull/2989 

The bots are using CMake 3.7.1
Cmake 3.23 works fine with this.

dneto0:
The android bots are not using the same VM as the others. I'm trying to update the config now.

PR #1262, opened 2022-08-12
Update Shaderc, Glslang, SPIRV-Tools

dneto0:
Shaderc v2022.2
Glslang GitHub master 2022-08-11
SPIRV-Headers GitHub master 2022-08-11
SPIRV-Tools v2022.3 plus one patch


dneto0:
Verified with Shaderc bots, and Android NDK build.

PR #1267, opened 2022-09-16
Add vcpkg installation instructions

FrankXie05:
`shaderc` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `shaderc` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `shaderc`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/shaderc/portfile.cmake). We try to keep the library maintained as close as possible to the original library. 😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,20 @@ test.vert
 /code $ glslc -c -o - test.vert | spirv-dis
 ```
 
+### Building and installing Shaderc using vcpkg 
+
+Alternatively, you can build and install shaderc using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+```sh
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
+./bootstrap-vcpkg.sh
+./vcpkg integrate install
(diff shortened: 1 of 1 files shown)


iaomw:
CMake Error at /home/iaomw/Github/vcpkg/scripts/buildsystems/vcpkg.cmake:834 (_find_package):
  Found package configuration file:

    /home/iaomw/Github/vcpkg/installed/x64-linux/share/unofficial-shaderc/unofficial-shaderc-config.cmake

  but it set unofficial-shaderc_FOUND to FALSE so package
  "unofficial-shaderc" is considered to be NOT FOUND.  Reason given by
  package:

  The following imported targets are referenced, but are missing:
  unofficial::shaderc_util::shaderc_util

PR #1272, opened 2022-10-12
Finalize v2022.3

dneto0:


dneto0:
This passes building in the NDK.

PR #1276, opened 2022-11-02
CMakeLists.txt: drop OSDependent/OGLCompiler from lists of glslang libraries

kanavin:
glslang no longer installs them separately, and all needed functionality has been merged into glslang shared library itself.

This wasn't a problem previously as they were still provided, as static libraries but in latest glslang they no longer are: https://github.com/KhronosGroup/glslang/commit/7cd519511c32d7e86d901c7ed231cb84c652d18d


dneto0 (approved):

juan-lunarg:
@dneto0 this can now be closed as well.

Issue #1278, opened 2022-11-13
How to build Shaderc from source ?


jiapei100:
I tried to build Shaderc from source, but obtained the following ERROR messages.

```console
[ 85%] Update build-version.inc in the Shaderc build directory (if necessary).
/usr/bin/python3 ....../shaderc/utils/update_build_version.py ....../shaderc ....../shaderc/build/build-version.inc
usage: ....../shaderc/utils/update_build_version.py <shaderc-dir> <spirv-tools-dir> <glslang-dir> <output-file>
make[2]: *** [CMakeFiles/build-version.dir/build.make:74: CMakeFiles/build-version] Error 1
make[2]: Leaving directory '....../shaderc/build'
make[1]: *** [CMakeFiles/Makefile2:257: CMakeFiles/build-version.dir/all] Error 2
make[1]: Leaving directory '....../shaderc/build'
make: *** [Makefile:149: all] Error 2
```

Did anybody meet the same issue?

jiapei100:
Problem solved: [https://github.com/google/shaderc/pull/1279](https://github.com/google/shaderc/pull/1279)

ben-clayton:
`usage: ....../shaderc/utils/update_build_version.py <shaderc-dir> <spirv-tools-dir> <glslang-dir> <output-file>`

Is suggesting that `update_build_version.py` is unhappy with the arguments. It looks like you've trimmed the paths with `......`, so I have to speculate here - but do you have whitespaces in your directory paths?

PR #1279, opened 2022-11-13
In order to build build-version.inc

jiapei100:
Hmmm...
Deal with build-version.inc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,11 @@ endif()
 project(shaderc)
 enable_testing()
 
+
+set(CMAKE_CURRENT_BINARY_DIR ".")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,11 @@ endif()
 project(shaderc)
 enable_testing()
 
+
+set(CMAKE_CURRENT_BINARY_DIR ".")
+set(spirv-tools_SOURCE_DIR "YourFolderTo-SPIRV-Tools")
```
ben-clayton, line 27: We don't want to have people have to modify `CMakeLists.txt` in order to build.

ben-clayton (changes requested):
This isn't a good fix for the issue you reported in #1278.

We should figure out why the `update_build_version.py` script isn't happy with your build environment. I'll follow up on #1278.

PR #1280, opened 2022-11-22
Add 'BUNDLE DESTINATION' for target install

antiagainst:
CMake would need a `BUNDLE DESTINATION` for installing MACOSX_BUNDLE targets. It falls back to use `RUNTIME DESTINATION` as per the `OLD` behavior of CMP0006
(https://cmake.org/cmake/help/latest/policy/CMP0006.html), but that's deprecated and will be removed in future versions.


antiagainst:
For more context: I'm seeing errors when compiling shaderc for shaderc-rs: https://github.com/google/shaderc-rs/actions/runs/3331419724/jobs/5898768267. This commit should address it.

dneto0 (approved):

PR #1283, opened 2022-11-30
Update DEPS

dneto0:
Move to latest SPIRV-Headers, SPIRV-Tools, Glslang


dneto0:
This passed everything but the gcov run. gcc 7 has been removed from the testing image.

gcc-10 does not work:
https://github.com/google/shaderc/issues/1284

Issue #1284, opened 2022-11-30
gcc coverage build is failing


dneto0:
The Kokoro coverage bot tries to use gcc 7, but that has been removed from the VM.
I tried gcc-10 and that still fails. 

Consider it broken.

PR #1287, opened 2022-12-05
Remove GCC-based code coverage builds

dneto0:
Fixed: #1284


dneto0:
Requires a corresponding change within Google to disable the corresponding Kokoro jobs.

PR #1294, opened 2023-01-16
Update shaderc

dneto0:


dneto0:
This passes the Android NDK build

PR #1295, opened 2023-01-17
Update dependencies

dneto0:
Shaderc v2023.1
Glslang GitHub 2023-01-16 top of tree
SPIRV-Tools v2023.1
SPIRV-Headers GitHub 2023-01-16 top of tree


dneto0:
Verified with an Android NDK build.

PR #1298, opened 2023-01-18
Update DEPS, finalize v2021.2, start V2023.3

dneto0:


dneto0:
Passed testing with the Android NDK build.

Issue #1300, opened 2023-01-23
[Feature Request] Add flag to preserve bindings


kuhar:
When compiling with optimizations enabled, `glslc` removes unused bindings. This is [legal in GLSL](https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)#Memory_layout), but has an undesirable (to me) side-effect that is effectively breaks the shader's ABI.

This can be worked around by disabling optimizations or adding a superficial use of otherwise unused buffers that spirv-opt can't elide (e.g., read element `[0]` and store in an output buffer). The latter workaround is fairly brittle as future optimizations may break it. Another possibility would be to inspect the compiled shaders and check which bindings are still present in the final `.spv` file, but this would have some runtime penalty and potentially hide bugs in the application's bindings.  

I'd like to propose a new `glslc` flag matching spirv-opt's `--preserve-bindings` as a more robust solution.

kuhar:
cc: @alan-baker, @antiagainst

PR #1301, opened 2023-01-23
Add option to preserve bindings

kuhar:
The new option `-fpreserve-bindings` matches the spirv-option `-preserve-bindings`.

Fixes: https://github.com/google/shaderc/issues/1300

### DIFF
--- a/libshaderc_util/src/spirv_tools_wrapper.cc
+++ b/libshaderc_util/src/spirv_tools_wrapper.cc
@@ -17,6 +17,7 @@
 #include <algorithm>
 #include <sstream>
 
+#include "spirv-tools/libspirv.hpp"
 #include "spirv-tools/optimizer.hpp"
 
 namespace shaderc_util {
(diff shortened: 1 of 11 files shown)


libshaderc_util/src/spirv_tools_wrapper.cc
```diff
@@ -108,6 +109,7 @@ bool SpirvToolsAssemble(Compiler::TargetEnv env,
 bool SpirvToolsOptimize(Compiler::TargetEnv env,
                         Compiler::TargetEnvVersion version,
                         const std::vector<PassId>& enabled_passes,
+                        spvtools::OptimizerOptions& optimizer_options,
```
dneto0, line 112: In the past we've stayed away from non-const references as params.
But that rule is gone.
Fortunately the caller destroys the enclosing object pretty quickly, so this seems ok.

dneto0 (approved):
Thanks for threading all this through the various layers!

PR #1313, opened 2023-03-17
kokoro: Remove VS2015 scripts, add VS2019 scripts

dneto0:


dneto0:
Need to configure Kokoro production configs to remove vs2015 first.
Then land this,
then enable kokoro production configs to add vs2019.

PR #1315, opened 2023-03-17
Add VS2019 download links, remove VS2015 download links

dneto0:


dneto0:
Need to wait for the Kokoro production configs to be updated.

PR #1323, opened 2023-04-21
Android.mk: Use $(file ..) instead of raw @echo. Update to NDK 21d

### LINKED ISSUE #1321: llvm-ar.exe error "unknown command" when building via Android.mk
From https://github.com/android/ndk/issues/1815

dneto0:
Fixes building of combined.ar on Windows.

Fixes #1321

Additionally:
- Use generate-file-dir macro This saves calling 'mkdir -p' repeatedly for each file.
- Simplify header generation logic
- Avoid a useless eval. Invoke the combined lib rules on TARGET_OUT directly.

Update NDK dependency to Android NDK 21d
- The ndk-build is based on GNU make 4. We need that to support the $(file ...) function

### DIFF
--- a/kokoro/ndk-build/build.sh
+++ b/kokoro/ndk-build/build.sh
@@ -30,9 +30,9 @@ unzip -q ninja-linux.zip
 export PATH="$PWD:$PATH"
 
 # Get Android NDK.
-wget -q https://dl.google.com/android/repository/android-ndk-r18b-linux-x86_64.zip
-unzip -q android-ndk-r18b-linux-x86_64.zip
-export ANDROID_NDK=$PWD/android-ndk-r18b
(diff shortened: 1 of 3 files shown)


kokoro/ndk-build/build.sh
```diff
@@ -30,9 +30,9 @@ unzip -q ninja-linux.zip
 export PATH="$PWD:$PATH"
 
 # Get Android NDK.
-wget -q https://dl.google.com/android/repository/android-ndk-r18b-linux-x86_64.zip
-unzip -q android-ndk-r18b-linux-x86_64.zip
-export ANDROID_NDK=$PWD/android-ndk-r18b
+wget -q https://dl.google.com/android/repository/android-ndk-r21d-linux-x86_64.zip
```
dneto0, line 33: NDK 21d is used by the CI flows in related projects; it's built into the docker image and will save this download.

PR #1329, opened 2023-05-25
Add abseil as a dep

s-perron:
The latest version of RE2 depends on Abseil. This add Abseil as a dep
for shaderc, and updates RE2 to make sure it works.

This is needed to fix the spirv-tools smoke test.


dneto0 (approved):
Thank you!

I'll do the update to add abseil to the known-good.json

PR #1331, opened 2023-05-26
Add Abseil

dneto0:


dneto0:
tested by running update_shaderc_source.py, then building the code that ends up in src/

PR #1339, opened 2023-06-27
Fix MSVC runtime library linking in CMake

### LINKED ISSUE #1337: MSVC Static/Dynamic Runtime set incorrectly.
With CMake 3.15 and onwards using policy CMP0091, the -MD option is added. This is a problem because the [current code](https://github.com/google/shaderc/blame/4dc596ddc2702092c670e828745dc3e0338d83c1/CMakeLists.txt#L115) is looking for /MD, meaning that shaderc is being built with a dynamically linked CRT instead of the intended statically linked CRT.

The way to set the CRT with CMake 3.15 and above is `set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")`. This will change the default from dynamic CRT to static CRT.

I am willing to make a PR to fix this, but wanted to document that this is in fact an first.

charles-lunarg:
3.15 no longer appends /MD to the C_FLAGS and CXX_FLAGS by default when using the policy CMP0091 with the new behavior. This breaks any existing CMake code which searched for "/MD" to replace it with "/MT", as the flags no longer contain /MD. The fix is to use the new property MSVC_RUNTIME_LIBRARY so that all targets are configured correctly.

Fixes #1337

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,11 +14,6 @@
 
 cmake_minimum_required(VERSION 3.17.2)
 
-if (POLICY CMP00091)
-  # Enable MSVC Runtime Library Property
-  cmake_policy(SET CMP0091 NEW)
-endif()
-
 project(shaderc)
 enable_testing()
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -14,10 +14,8 @@
 
 cmake_minimum_required(VERSION 3.17.2)
 
-if (POLICY CMP00091)
-  # Enable MSVC Runtime Library Property
-  cmake_policy(SET CMP0091 NEW)
-endif()
+# Enable MSVC Runtime Library Property
+cmake_policy(SET CMP0091 NEW)
```
juan-lunarg, line 18:
```suggestion
```

juan-lunarg (changes requested):

CMakeLists.txt
charles-lunarg: Yep, setting cmake_minimum_required() to a version at 3.15 or above makes setting the policy manually redundant. Removing since its no longer necessary.

charles-lunarg:
Argh it appears there is another change necessary here: https://github.com/google/shaderc/blame/e31c4c2e41544d63d90be28c46e4a4793a624240/cmake/setup_build.cmake#L34

Seems that its only applicable to the tests, but stands to reason that the `gtest_force_shared_crt` option should match whatever `SHADERC_ENABLE_SHARED_CRT` was set to.

dneto0 (approved):

dneto0:
Landing this.  The  VS2017 break is for a different issue. #1345, which is independent.
The MINGW things is a misconfiguration for something different.

PR #1341, opened 2023-07-10
Use android ndk r25

s-perron:
We currently use R21 of the Android NDK for our tests. There have been
to LTS release since that one, and we do not expect people to use it
anymore. Also, it contains Python 2.7, not Python3. The python scripts
in SPIR-V Tools expect Python 3, so we have to update.

We chose the latest LTS release.

### DIFF
--- a/kokoro/android-release/build-docker.sh
+++ b/kokoro/android-release/build-docker.sh
@@ -27,7 +27,7 @@ set -x
 
 using cmake-3.17.2
 using ninja-1.10.0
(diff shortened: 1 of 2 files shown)


s-perron:
@dneto0 I just updated the setttings to require a review before merging. This should stop auto-merge from merging before there is a review approving the PR. Let me know if this needs to change. I can make necessary changes.

dneto0 (commented):
LGTM

Issue #1350, opened 2023-07-19
work around bug in CMake 3.22.1 bug in CMP0128 policy, for compiling SPIRV-Tools tests


dneto0:
This bug in the OLD implementation in CMake 3.22.1 is causing SPIR-V Tools test code to compile with tailing -std=c++11 when it really needs -std=c++17 to take effect.

A fix is 
cmake_policy(SET CMP0128 NEW)

The bug is fixed in CMake 3.22.2

When the bug is present, I get a build.ninja lines like this:

build third_party/spirv-tools/test/CMakeFiles/test_spirv_unit_tests.dir/binary_destroy_test.cpp.o: CXX_COMPILER__test_spirv_unit_tests_Release /build/ndk/external/shaderc/spirv-tools/test/binary_destroy_test.cpp || cmake_object_order_depends_target_test_spirv_unit_tests
  DEFINES = -DSPIRV_COLOR_TERMINAL -DSPIRV_LINUX -DSPIRV_TIMER_ENABLED
  DEP_FILE = third_party/spirv-tools/test/CMakeFiles/test_spirv_unit_tests.dir/binary_destroy_test.cpp.o.d
[...]

PR #1351, opened 2023-07-19
Set cmake_policy CMP0128 to NEW

dneto0:
Work around a problem in CMake 3.22.1 in setting -std=c++17. Instead -std=c++11 is in effect for targets in test/*, but those targets require C++17.

Fixes compiling SPIRV-Tools test targets in with certain Clang and CMake combinations.  Fixes a problem when trying to refresh shader-tools in the Android NDK.

Fixes: #1350

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,16 @@ cmake_minimum_required(VERSION 3.17.2)
 project(shaderc)
 enable_testing()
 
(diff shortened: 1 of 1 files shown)


dneto0:
This didn't work locally for me.
cmake_policy is apparently scoped to the current project.  So this can only be worked around in SPIRV-Tools, where the problem is triggered.

PR #1353, opened 2023-07-19
Update DEPS

dneto0:
SPIRV-Tools v2023.5-rc1 plus PR 5334
Glslang near top of tree 2023-07-19
SPIRV-Headers top of tree 2023-07-18
Shaderc v2023.5


dneto0:
Tested with the Android NDK build.
One item not fixed, which is https://github.com/KhronosGroup/SPIRV-Tools/issues/5340
That only affects CMake 3.22.1 with certain compilers, and only when building SPIRV-Tools tests.

PR #1358, opened 2023-08-03
Add a workflow to automatically roll DEPS

sudonatalie:
Create a GitHub action that syncs all git dependencies, rolls all entries in DEPS to tip-of-tree and commits, then uploads a new GitHub pull request with a trigger to automatically run kokoro tests.

The roll-deps script has been updated to not return an error when there are no new commits for a repository.

Both the autoroll.yml script and changes to utils/roll-deps are modeled after the scripts in KhronosGroup/SPIRV-Tools.

### DIFF
--- a/.github/workflows/autoroll.yml
+++ b/.github/workflows/autoroll.yml
@@ -1,0 +1,66 @@
+# Copyright 2023 The Shaderc Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,s either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


.github/workflows/autoroll.yml
```diff
@@ -0,0 +1,66 @@
...
+            echo "changed=false" >> $GITHUB_OUTPUT
+          else
+            echo "changed=true" >> $GITHUB_OUTPUT
+          fi
+        id: update_dependencies
+      - name: Push changes and create PR
+        if: steps.update_dependencies.outputs.changed == 'true'
+        run: |
+          git push --force --set-upstream origin roll_deps
+          gh pr create --label 'kokoro:run' --base main -f
```
s-perron, line 64: I have been wanting to fix this in the spirv-tools side, but I have not got to it yet. If a PR is already open, this step fail, and reports the entire workflow as failing. Is there a way to ignore the return code from this command?

s-perron (approved):
Thanks. This looks good. There is one issue I have seen in spirv-tools. Do you know how we could fix that?

.github/workflows/autoroll.yml
s-perron: Can we add `|| true`? We might miss some errors, but it is better than too many false errors.

```suggestion
          gh pr create --label 'kokoro:run' --base main -f || true
```

sudonatalie: I tested this out on my fork and it looks like it works

Issue #1343, opened 2023-07-17
adapt linux-mingw build for clang with mingw


dneto0:
The CI-linux-mingw bot is broken. 
We have a MinGW bot to support the flow used in the build of the Android NDK itself.
But the NDK has switched to cross-compiles with clang.  Our build is not adapted for that.

dneto0:
The MinGW bot is broken because:
- spirv-tools tools/objdump/objdump.cpp uses `#include <filesystem>` which is a C++17 feature
- but the Kokoro bot MinGW toolchain is the default one on that version of ubuntu which is derived from GCC 7.3, and it doesn't fully support C++17. The best it can muster is c++1z

dneto0:
The NDK build uses a prebuilt Clang compiler with runtime libraries from MinGW v4.8, i.e. aligning with features in GCC 4.8.
That's too old and not fixable by me.

The MinGW build doesn't have enough benefit to justify maintaining its bots.  I'll remove them instead.

PR #1360, opened 2023-08-08
Finalize v2023.6

dneto0:


dneto0:
The MinGW builds have never worked, and I've now determined they are not useful / don't reflect the thing they're trying to gate (i.e. the NDK use of mingw as a cross compiler). See #1343

PR #1356, opened 2023-07-25
cmake: Remove redundant -std=c++11 flag

dneto0:
This is already covered by the CMAKE_CXX_STANDARD directive

PR #1362, opened 2023-08-09
Update README.md: Need a C++17 compiler

dneto0:


TheLavaBlock:
It would be great if the remaining parts could be changed as well:
- https://github.com/google/shaderc/pull/1370
- https://github.com/google/shaderc/pull/1356

PR #1363, opened 2023-08-09
Remove redundant gtest compile flags for MinGW

dneto0:
When compiling with MinGW, we no longer need to use -Dgtest_disable_pthreads=ON.

See https://github.com/google/googletest/pull/856
and https://github.com/google/googletest/pull/721


dneto0:
I verified this build manually.   The kokoro bots don't have a MinGW build that supports C++17 so the bots are disabled (and therefore don't exercise this path).

ben-clayton (approved):

PR #1364, opened 2023-08-10
Adapt test to different error message from SPIRV-Tools

dneto0:
Update SPIRV-Tools

The SPIRV-Tools message has changed, for the case when a capability was used that is not appropriate to the current target environment.


dneto0:
Error message changed due to https://github.com/KhronosGroup/SPIRV-Tools/pull/5370

alan-baker (approved):

Issue #1367, opened 2023-09-06
Ninja Multi-Config generator fails - Evaluation file to be written multiple times with different content


kroppt:
I was trying to use this project and ran into what appears to be a bug in the CMake configuration. When using Ninja Multi-Config as the generator, the project errors at CMake configure. I'm using Linux, so this is the only multi-config generator available.

Minimal, reproducible example:
```bash
git clone https://github.com/google/shaderc.git
cd shaderc
./utils/git-sync-deps
cmake -B build -G "Ninja Multi-Config"
```

Output:
```none
Cloning into 'shaderc'...
remote: Enumerating objects: 7168, done.
remote: Counting objects: 100% (767/767), done.
remote: Compressing objects: 100% (348/348), done.
remote: Total 7168 (delta 468), reused 567 (delta 393), pack-reused 6401
Receiving objects: 100% (7168/7168), 2.45 MiB | 12.33 MiB/s, done.
Resolving deltas: 100% (4858/4858), done.
/home/tyler/code/shaderc/third_pa... @ 19b4aa87af25cb4ee779a071409732f34bfc305c
/home/tyler/code/shaderc/third_pa... @ c9cba76063cf4235c1a15dd14a24a4ef8d623761
/home/tyler/code/shaderc/third_pa... @ 124a9665e464ef98b8b718d572d5f329311061eb
/home/tyler/code/shaderc/third_pa... @ c541e7c11044b1e0303103ef8a47d7a9632c479b
/home/tyler/code/shaderc/third_pa... @ 8e3da01b45806fbacbb9e6fce9c5f9ae49f60e42
/home/tyler/code/shaderc/third_pa... @ 5be22f98733c674d532598454ae729253bc53e82
/home/tyler/code/shaderc/third_pa... @ 76b52ebf77833908dc4c0dd6c70a9c357ac720bd
-- The C compiler identification is GNU 13.2.1
-- The CXX compiler identification is GNU 13.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- No build type selected, default to Debug
-- Shaderc: build type is "Debug".
[...]

Issue #473, opened 2018-07-31
cmake/utils.cmake doesn't properly quote arguments to "ar"


linuxturtle:
When utils.cmake builds a directive file for "ar" as part of building libshaderc/libshaderc_combined.a, it fails to properly quote the arguments to "ar", thus causing "ar" to fail if the build directory happens to contain any characters which might be construed to mean something special on the command line.  Here's an example from me trying to build a debian package:

```
[313/319] cd "/build/shaderc-2018.0~dev+20/debian/build/libshaderc" && /usr/bin/ar -M < 
shaderc_combined.ar
FAILED: libshaderc/libshaderc_combined.a 
cd "/build/shaderc-2018.0~dev+20/debian/build/libshaderc" && /usr/bin/ar -M < shaderc_combined.ar
/usr/bin/ar: /build/shaderc-2018.0~dev: No such file or directory
+Syntax error in archive script, line 1

```
Here is a patch to fix the problem:

```
diff --git a/src/cmake/utils.cmake b/src/cmake/utils.cmake
index ed3c733..519d28f 100644
--- a/src/cmake/utils.cmake
+++ b/src/cmake/utils.cmake
@@ -186,9 +186,9 @@ function(shaderc_combine_static_lib new_target target)
       DEPENDS ${all_libs}
       COMMAND libtool -static -o ${libname} ${lib_target_list})
   else()
-    string(REPLACE ";" "> \naddlib $<TARGET_FILE:" temp_string "${all_libs}")
+    string(REPLACE ";" "> '\naddlib '$<TARGET_FILE:" temp_string "${all_libs}")
     set(start_of_file
-      "create ${libname}\naddlib $<TARGET_FILE:${temp_string}>")
+      "create '${libname}'\naddlib '$<TARGET_FILE:${temp_string}>'")
     set(build_script_file "${start_of_file}\nsave\nend\n")
 
     file(GENERATE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${new_target}.ar"

```

linuxturtle:
Hmm, I should point out that this patch doesn't touch the "MSVC" or "APPLE" versions of output.  I assume they'd need similar changes, but am not familiar with either environment.

dneto0:
Thanks for the good bug report and suggested solution.   I think this flow is used on OSX but not on Windows.  OSX uses Bash as well, so this solution will work.

dneto0:
Actually, I see the MSVC case there too.

[...]

PR #1368, opened 2023-09-07
CMake: Use TARGET_OBJECTS to replace archive acrobatics

kroppt:
Rather than trying to manually merge static libraries into a combined static library, we can instead just let CMake do it for us.

This solves:
- https://github.com/google/shaderc/issues/1367
- https://github.com/google/shaderc/issues/473

I've successfully tested this with the following generators:
- Visual Studio 17 2022
- Unix Makefiles
- Ninja
- Ninja Multi-Config
- CodeBlocks - Ninja
- CodeBlocks - Unix Makefiles
- CodeLite - Ninja
- CodeLite - Unix Makefiles
- Eclipse CDT4 - Ninja
- Eclipse CDT4 - Unix Makefiles
- Kate - Ninja
- Kate - Unix Makefiles
- Sublime Text 2 - Ninja
- Sublime Text 2 - Unix Makefiles

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -173,50 +173,10 @@ endmacro()
 # Combines the static library "target" with all of its transitive static
 # library dependencies into a single static library "new_target".
 function(shaderc_combine_static_lib new_target target)
-
   set(all_libs "")
   shaderc_get_transitive_libs(${target} all_libs)
-
-  set(libname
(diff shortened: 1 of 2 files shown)


dneto0 (approved):
This is much much much better.
Thank you!

The TARGET_OBJECTS generator was introduced in 3.1, and this project stuck to 2.8 for a long time.  But now we're on a much later CMake and can move to this.

Thank you for fixing this headache!

PR #1380, opened 2023-11-22
Expose setting relaxed Vulkan rules from glslang

dnfield:
This exposes https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GL_EXT_vulkan_glsl_relaxed.txt, which allows for compilation of many OpenGL GLSL shaders to Vulkan SPIRV.

Adds a simple test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ third_party/tint
 android_test/libs
 android_test/include
 .DS_Store
+.vscode/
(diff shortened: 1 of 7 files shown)


dneto0 (approved):
Looks good to me.

It would be nice to follow up to expose this on the glslc command line.

PR #1382, opened 2023-11-30
Fix for cmake policy CMP0148

stuart6854:
FindPythonInterp has been deprecated since CMake 3.12. 
This change makes CMake use FindPython for version > 3.12 (otherwise it uses FindPythonInterp).
Tested with CMake 3.27.

https://cmake.org/cmake/help/latest/policy/CMP0148.html


dneto0 (approved):
Thank you!

Issue #1391, opened 2024-01-25
SPIR_V: Generate nonsemantic shader debug information (with source)


tksuoran:
Enabling nonsemantic shader debug information would allow source debugging with RenderDoc, for example.

Essentially same as the following options for `glslang`:
```
  -gV         generate nonsemantic shader debug information
  -gVS        generate nonsemantic shader debug information with source
```

StefanPoelloth:
any progress on this?

PR #1392, opened 2024-01-25
Generate nonsemantic shader debug information

tksuoran:
Proof of concept fix for #1391.

Possibly `emitNonSemanticShaderDebugInfo` and `emitNonSemanticShaderDebugSource` should be controlled using dedicated command line arguments instead of just `-g`.

This worked for me and allowed me to use RenderDoc source debugging.

### DIFF
--- a/libshaderc_util/src/compiler.cc
+++ b/libshaderc_util/src/compiler.cc
@@ -341,6 +341,10 @@ std::tuple<bool, std::vector<uint32_t>, size_t> Compiler::Compile(
   options.generateDebugInfo = generate_debug_info_;
   options.disableOptimizer = true;
   options.optimizeSize = false;
+  if (generate_debug_info_) {
+    options.emitNonSemanticShaderDebugInfo = true;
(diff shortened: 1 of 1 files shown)


vinh-truong-sc:
Hey, we are also interested in this feature. Can someone look at this and merge it?

dneto0 (approved):
The code is good.
The policy question is that when using -g the downstream tools and driver have to undertand non-semantic extended instruction sets.  So in theory it's a backward incompatibility.   But this is about _debug info_ and it's being used in a development environment so it has to be deliberately turned on, and the remedy is easy: turn it back off.

So I think this is the right thing to do, even with the theoretical incompatibility.

PolarNick239:
Hi! Debug information is very useful for Vulkan-API debugging with RenderDoc. I hope it will be merged soon! :)

Trider12:
Why hasn't this been merged?

dj2:
There are failing tests which need to be fixed up. The `glslc_tests`, and a bunch of the shaderc test suites all failed with the patch applied.

Issue #1396, opened 2024-02-12
enable Glslang tests when Shaderc tests are enabled


dneto0:
Glslang tests are only enabled if the Cmake project is top-level.
See this change https://github.com/KhronosGroup/glslang/commit/cf1fbbff44c0900823b94a2122861d0d593432a8

dneto0:
https://github.com/KhronosGroup/SPIRV-Headers/pull/418 would update SPIRV-Headers to allow testing and installing.

PR #1397, opened 2024-02-12
Propagate test/install options to Glslang

dneto0:
This needs https://github.com/KhronosGroup/glslang/pull/3508 to be effective

Fixes: #1396

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -83,6 +83,11 @@ endif()
 
 if (NOT TARGET glslang)
   if (IS_DIRECTORY ${SHADERC_GLSLANG_DIR})
+    if (SHADERC_ENABLE_TESTS)
+      # Glslang tests are off by default. Turn them on if testing Shaderc.
+      set(GLSLANG_TESTS ON)
+    endif()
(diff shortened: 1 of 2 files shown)


third_party/CMakeLists.txt
```diff
@@ -83,6 +83,11 @@ endif()
 
 if (NOT TARGET glslang)
   if (IS_DIRECTORY ${SHADERC_GLSLANG_DIR})
+    if (SHADERC_ENABLE_TESTS)
+      # Glslang tests are off by default. Turn them on if testing Shaderc.
+      set(GLSLANG_ENABLE_TESTS ON)
```
alan-baker, line 88: I think the option is incorrectly named. The glslang CMakeLists.txt has:
>option(GLSLANG_TESTS "Enable glslang testing" ${GLSLANG_TESTS_DEFAULT})

```suggestion
      set(GLSLANG_TESTS ON)
```

dneto0: tested that, and it works. thanks
Tests fail, so we know they are running

alan-baker (approved):

PR #1401, opened 2024-02-23
Fix Python

moritz-h:
Fix executing the Python scripts during build from source.
The problem is that `${PYTHON_EXECUTABLE}` is undefined. This leads to the `add_custom_target` calls trying to execute the Python scripts directly, which probably just works most of the times. But if you have i.e. on Windows associated the .py extension with a code editor, the editor opens instead of the script beeing executed.

I think the problem is that `find_package(Python COMPONENTS Interpreter REQUIRED)` changed the casing to `${Python_EXECUTABLE}` compated to the previously used `${PYTHON_EXECUTABLE}` from `find_package(PythonInterp 3 REQUIRED)`.

As the main CMakeLists.txt defines `cmake_minimum_required(VERSION 3.17.2)` anyway, I just removed the version check for CMake <= 3.12 and updated to the new syntax.


dneto0 (approved):
Thanks!

PR #1417, opened 2024-06-03
Update DEPS

alan-baker:
* Roll spirv-tools, spirv-headers, and glslang deps
* Modify glslc tests where that depended on SampleCmp HLSL calls since glslang generates invalid SPIR-V


alan-baker:
See https://github.com/KhronosGroup/SPIRV-Tools/pull/5695 for the reason behind the test changes.

s-perron (approved):

PR #1420, opened 2024-06-17
Update dependencies, except Glslang

dneto0:


s-perron (approved):

s-perron:
RE2 requires VS2019, which is why the VS2017 builds are failing.

dneto0:
Looks like Abseil now requires VS2019 or later.  I'll update the README and the build status links.

PR #1440, opened 2024-08-07
update DEPS: glslang, spirv-tools, spirv-headers

dneto0:


dneto0:
This is a routine update to latest top-of-tree.  It passes on linux on my machine

s-perron (approved):

PR #1451, opened 2024-11-04
Update glslang dep to include GLSL_NV_cooperative_matrix2

jeffbolznv:


alan-baker (approved):

BtbN:
This PR had leads to errors like the following:

```
shaderc/third_party/glslang/SPIRV/disassemble.cpp:341:8: error: 'uint32_t' does not name a type
```

This was fixed in https://github.com/KhronosGroup/glslang/commit/76cc41d26f6902de543773023611e40fbcdde58b so the dependency might need bumped again.

PR #1463, opened 2024-12-17
kokoro: windows: use Python 3.12

dneto0:
crbug.com/350048185


dnovillo (approved):
Can't access https://crbug.com/350048185 but I suppose I don't need to read it to understand what this does.

dneto0:
> Can't access https://crbug.com/350048185 but I suppose I don't need to read it to understand what this does.

I'm just updating Python on all the bots used by the kokoro images I mangage

PR #1475, opened 2025-01-23
kokoro: use gcc 13

dneto0:


dneto0:
CI-macos-clang-release build succeeded (i looked at internal logs), but that status didn't propagate back.
I've triggered a rebuild.

PR #1478, opened 2025-01-29
kokoro linux: use 'lean' image based on ubuntu 24.04

dneto0:


dneto0:
oops. I was mistaken the build config settings have to be in the *same* directory as the presubmit.cfg and continuous.cfg.
So this doesn't actually work.
We have to run docker with the explicit image name.

PR #1480, opened 2025-01-31
kokoro: use lightweight image for license checker

dneto0:


dneto0:
The license checker image is only 88MB, 32MB compressed.  The image was downloaded and the checker ran all in 7 seconds.
That's very good.

PR #1491, opened 2025-04-12
update glslang dependency

jeffbolznv:
Update glslang dependency to include bfloat16 support (https://github.com/KhronosGroup/glslang/pull/3905/).


jeffbolznv:
Ping @dneto0 @alan-baker

dneto0 (approved):

PR #1493, opened 2025-04-16
Don't link glslang stub SPIRV library

jpr42:


jpr42:
I don't have a google account and their CLA tool requires me to make one.

"A Google account is required to sign the CLA. Google accounts are used to link individual and corporate CLAs to covered contributors and provide authentication for CLA management."

PR #1498, opened 2025-06-02
Fixed glslc binaries being build even if enable install is off.

Z3DBreaker:
For whatever reason even if this toggle is set to OFF it will still build the glsc binary so I added a if there and that seems to fix it.

### DIFF
--- a/glslc/CMakeLists.txt
+++ b/glslc/CMakeLists.txt
@@ -32,8 +32,8 @@ add_library(glslc STATIC
 shaderc_default_compile_options(glslc)
 target_include_directories(glslc PUBLIC ${glslang_SOURCE_DIR})
 
-if (SHADERC_ENABLE_WGSL_OUTPUT)
-  if (IS_DIRECTORY "${tint_SOURCE_DIR}/include")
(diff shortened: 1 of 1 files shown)


dj2:
Why guard building the binary on install? If you'd prefer to not build the binary, it would be better to add a custom flag to disable binary generation.

JennieLizz:
> There is already a build flag for binary? If so, then the install check shouldn't be needed, yea?

Yes, there are other binaries that do get cut off by this flag the only exception is glsc. I have no clue why.

JennieLizz:
If there is a flag I might've missed then please let me know.

dj2:
`glslc` is the only binary produced, is it not? There is also the shaderc library, but I think there is only one binary executable. There is currently no flag to put binaries behind if you don't want them built, it would have to be added.

But, the building of binaries should not be gated on install.

JennieLizz:
> `glslc` is the only binary produced, is it not? There is also the shaderc library, but I think there is only one binary executable. There is currently no flag to put binaries behind if you don't want them built, it would have to be added.
> 
> But, the building of binaries should not be gated on install.

Okay, Ill add something more like that then. Thanks!

PR #1499, opened 2025-06-02
Added a way to disable the glsc CLI executable.

JennieLizz:


JennieLizz:
I've tested this and it seems to work perfectly.

CMakeLists.txt
```diff
@@ -51,6 +51,16 @@ else()
   message(STATUS "Configuring Shaderc to avoid building examples.")
 endif()
 
+option(SHADERC_SKIP_EXECUTABLES "Skip building executables." ${SHADERC_SKIP_EXECUTABLES})
+if(NOT ${SHADERC_SKIP_EXECUTABLES})
+  set(SHADERC_ENABLE_EXECUTABLES ON)
+endif()
+if (${SHADERC_SKIP_EXECUTABLES})
```
dj2, line 58: I think this flag is backwards? If it's set to skip, it prints it's building.

CMakeLists.txt
```diff
@@ -51,6 +51,16 @@ else()
   message(STATUS "Configuring Shaderc to avoid building examples.")
 endif()
 
+option(SHADERC_SKIP_EXECUTABLES "Skip building executables." ${SHADERC_SKIP_EXECUTABLES})
```
dj2, line 54: Just to verify, if you don't set anything, this defaults to on, correct?

JennieLizz: I figured to leave the executables on by default since that's how it was originally set up, and I don't want to mess up anyone else's projects if they require it for whatever reason.

dj2: Yes, having them on by default is the right choice.

CMakeLists.txt
dj2: If I'm reading this right, if `SHADERC_SKIP_EXECUTABLES` is `true`, meaning we'll skip the creation of executables, the message printed is `Configuring Shaderc to build executables` which, seems like the opposite of what we've done, no?

dneto0: Right, this condition should be `${SHADERC_ENABLE_EXECUTABLES}`

dneto0 (changes requested):

CMakeLists.txt
```diff
@@ -51,6 +51,16 @@ else()
...
+if(NOT ${SHADERC_SKIP_EXECUTABLES})
+  set(SHADERC_ENABLE_EXECUTABLES ON)
+endif()
+if (${SHADERC_ENABLE_EXECUTABLES})
```
JennieLizz, line 58: I fixed the variable being the wrong one. I switched it to SHADERC_ENABLE_EXECUTABLES instead of the reversed one of SHADERC_SKIP_EXECUTABLES.

dj2 (approved):

PR #1508, opened 2025-07-31
Update DEPS

kpet:
To pick up GL_ARM_tensors in glslang and a few tensor-related fixes in SPIRV-Tools.


kpet:
Works for me, thanks!

PR #1516, opened 2025-09-04
When glslc is not built, don't install or test it

dneto0:
Fix the logic when SHADERC_SKIP_EXECUTABLES=ON

* Don't try to install the glslc executable
* Don't try to run tests that invoke glslc


glslc/test/CMakeLists.txt
```diff
@@ -21,10 +21,14 @@ if(${SHADERC_ENABLE_TESTS})
...
 
-  add_test(NAME glslc_tests
-    COMMAND ${Python_EXECUTABLE}
-    ${CMAKE_CURRENT_SOURCE_DIR}/glslc_test_framework.py
-    $<TARGET_FILE:glslc_exe> $<TARGET_FILE:spirv-dis>
-    --test-dir ${CMAKE_CURRENT_SOURCE_DIR})
-    
+  # These tests use the glslc executable, so they can only run if
+  # glslc has been built.
+  if(${SHADERC_ENABLE_EXECUTABLES})
```
dj2, line 26: I never understand this with CMake, why is this one `${SHADERC_ENABLE_EXECUTABLES}` and the other is `SHADERC_ENABLE_EXECUTABLES` ?

dneto0: same thing.  I just kept local style here.

dj2 (approved):

PR #1518, opened 2025-09-22
Add -fmax-id-bound flag

s-perron:
Add an option for shaderc to be able to set the max id bound used in the
compiler. This is useful for user who are hitting the limit, but are
able to use a higher value.

See https://github.com/KhronosGroup/SPIRV-Tools/issues/6260.

### DIFF
--- a/libshaderc_util/include/libshaderc_util/compiler.h
+++ b/libshaderc_util/include/libshaderc_util/compiler.h
@@ -319,6 +319,8 @@ class Compiler {
     preserve_bindings_ = preserve_bindings;
   }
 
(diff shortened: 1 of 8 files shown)


libshaderc_util/include/libshaderc_util/compiler.h
```diff
@@ -534,6 +536,8 @@ class Compiler {
   // True if the compiler should preserve all bindings, even when unused.
   bool preserve_bindings_;
 
+  uint32_t max_id_bound_ = 0x3FFFFF;
```
dneto0, line 539: Note: This limit comes from the SPIR-V spec.

dneto0 (approved):

PR #1546, opened 2026-04-11
Fix null pointer dereference in DependencyInfoDumpingHandler

YLChen-007:
## Description
This PR fixes a null pointer dereference vulnerability (SIGSEGV) in `DependencyInfoDumpingHandler::DumpDependencyInfo()` that occurs when `glslc` is invoked with the `-MD` flag, and the resulting dependency info output file cannot be opened for writing.

## Root Cause
When attempting to generate the `.d` dependency file in a read-only directory, a non-existent path, or when the disk is full, the utility function `shaderc_util::GetOutputStream()` correctly logs an error and returns `nullptr`. 

However, in `glslc/src/dependency_info.cc`, the stream pointer `dep_file_stream` was dereferenced unconditionally without a null check:
```cpp
*dep_file_stream << dep_string_stream.str();
```
This resulted in an immediate segmentation fault (SIGSEGV) and a process crash.

## Fix
Added a standard null check to verify `dep_file_stream` before writing to it. If it is null, the handler gracefully returns `false`. This aligns perfectly with the identical mitigation pattern previously established in `glslc/src/file_compiler.cc` (e.g. Commit `1d9790184b2e8fb726719deac80caaf6374daed7`).

## Testing
We have reproduced the crash locally and verified that with this patch, `glslc` now correctly exits with a graceful error output (`exit code 1`) instead of crashing when facing unwritable output dependency paths.

### DIFF
--- a/glslc/src/dependency_info.cc
+++ b/glslc/src/dependency_info.cc
@@ -50,6 +50,10 @@ bool DependencyInfoDumpingHandler::DumpDependencyInfo(
     std::ofstream potential_file_stream_for_dep_info_dump;
     std::ostream* dep_file_stream = shaderc_util::GetOutputStream(
         dep_file_name, &potential_file_stream_for_dep_info_dump, &std::cerr);
+    if (!dep_file_stream) {
+      // An error message has already been emitted to the stderr stream.
+      return false;
+    }
     *dep_file_stream << dep_string_stream.str();
     if (dep_file_stream->fail()) {
       std::cerr << "glslc: error: error writing dependent_files info to output "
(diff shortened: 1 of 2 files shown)


dneto0 (changes requested):
The logic looks ok.

Please add a test to https://github.com/google/shaderc/blob/main/glslc/test/option_dash_M.py 

Also, please sign the CLA.  I can't accept the changes without the CLA.

dneto0 (approved):
Thanks!

dneto0:
Ah, please rebase against main.
The old version of the code is trying to run against tools that have been removed from the bot image.

dneto0 (approved):

PR #1552, opened 2026-05-03
Fix UAF in CompileOptions copy/move constructor with custom includer

HARISHWAR-T:
## Problem

When SetIncluder() is called, the includer is stored in includer_ 
(unique_ptr) and its raw pointer is registered as C callback 
user_data via shaderc_compile_options_set_include_callbacks.

The copy constructor only clones options_ at the C level, which 
copies include_user_data as a raw pointer. When the original 
CompileOptions is destroyed, the copied object holds a dangling 
pointer that is called when resolving #include directives.

The move constructor had the same issue.

## Impact

Use-after-free / dangling virtual call when:
1. CompileOptions created and SetIncluder() called
2. Options object is copied or moved
3. Original is destroyed
4. Shader containing #include is compiled with the copy

## Fix

- Added Clone() to IncluderInterface (default returns nullptr)
- Copy constructor calls Clone() and re-registers via SetIncluder()
- If Clone() not implemented, callbacks are safely nullified
- Move constructor now transfers includer_ ownership correctly
- Tests updated to cover this lifecycle

### DIFF
--- a/libshaderc/include/shaderc/shaderc.hpp
+++ b/libshaderc/include/shaderc/shaderc.hpp
@@ -143,10 +143,18 @@ class CompileOptions {
   ~CompileOptions() { shaderc_compile_options_release(options_); }
   CompileOptions(const CompileOptions& other) {
     options_ = shaderc_compile_options_clone(other.options_);
+    if (other.includer_) {
+      if (auto* cloned_includer = other.includer_->Clone()) {
+        SetIncluder(std::unique_ptr<IncluderInterface>(cloned_includer));
+      } else {
+        shaderc_compile_options_set_include_callbacks(options_, nullptr, nullptr, nullptr);
+      }
+    }
   }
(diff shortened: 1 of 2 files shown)


HARISHWAR-T:
This issue results in a use-after-free due to lifetime mismatch between the C++ `IncluderInterface` ownership (`unique_ptr`) and the C-level `include_user_data` callback pointer.

The fix ensures:

* Safe ownership transfer in move operations
* Deep copy semantics via `Clone()` when available
* Safe fallback by nullifying callbacks when cloning is not supported

I verified the fix with lifecycle tests covering:

* Copy after SetIncluder()
* Move semantics
* Destruction of original instance before compilation

Happy to adjust the approach if there are preferred patterns for managing callback user_data in shaderc.
****

dneto0:
Please sign the CLA.

dneto0:
I can't accept contributions without a signed CLA.

Issue #1556, opened 2026-05-05
deprecate and eventually remove HLSL support


dneto0:
This follows deprecation and removal of HLSL support in Glslang.  https://github.com/KhronosGroup/glslang/issues/4210

PR #1557, opened 2026-05-05
Add SHADERC_ENABLE_HLSL option, defaulted to ON

dneto0:
The HLSL front end in Glslang is deprecated and will
eventually be removed.
https://github.com/KhronosGroup/glslang/issues/4210

Add CMake configuration option SHADERC_ENABLE_HLSL,
defaulting to ON.  In the future the default will
switch to OFF.

Adjust unit tests and glslc end-to-end tests.

Issue: #1556


dj2 (approved):

y
```diff
@@ -0,0 +1,18 @@
+-- Shaderc: build type is "Release".
```
dj2: Should this file be here?

dneto0: duh, no.

Also, I have to pull in the glslang dep!

PR #1560, opened 2026-05-13
Update dependencies

dneto0:


dneto0:
This pulls in fixes for glslang changes to compute shader derivatives

dj2 (approved):

PR #1561, opened 2026-05-13
Android.mk: support SHADERC_ENABLE_HLSL

dneto0:
two commits here.
- one to update the ndk build
- another to test it in kokoro


kokoro/scripts/ndk-build/build.sh
```diff
@@ -15,6 +15,9 @@
 #
 # Linux Build Script.
 
+# Enable Glslan'g backend? ON or OFF
```
dj2:
```suggestion
# Enable Glslang backend? ON or OFF
```

dj2 (approved):

kokoro/scripts/ndk-build/build.sh
dneto0: updated

PR #1565, opened 2026-05-14
Update dependencies

dneto0:


dneto0:
This will also let me check the new no-hlsl kokoro flows

dj2 (approved):

PR #1575, opened 2026-06-21
Sync #pragma shader_stage(<stage>) to -fshader-stage=<stage>

LBV2012-26:
This PR synchronizes the shader stages flags for the `-fshader-stage=` option with the `#pragma shader_stage(<stage>)` list in `glslc/src/main.cc`.  Such as `raygen`, `closest`, etc.


dneto0:
I think it's a bad idea to make new conventions for shader stage names that are aliases of existing names.

Let's stick to the conventions in Glslang:

https://github.com/KhronosGroup/glslang/blob/ce138e2c2d6992b31ff4cd2e955904637785a881/README.md?plain=1#L80 

Also, updates like this should come with tests.

dneto0 (changes requested):
see comments on the main part of this PR

LBV2012-26:
> I think it's a bad idea to make new conventions for shader stage names that are aliases of existing names.
> 
> Let's stick to the conventions in Glslang:
> 
> https://github.com/KhronosGroup/glslang/blob/ce138e2c2d6992b31ff4cd2e955904637785a881/README.md?plain=1#L80
> 
> Also, updates like this should come with tests.

Thanks for your response. But it's not the new aliases. Actually it's the same as custom pragma directives of shaderc (in libshaderc_util/src/shader_stage.cc), such as `#pragma shader_stage(raygen)`. In addition, the `tesseval` and `tesscontrol` are also not defined by glslang.
I submit this PR is not intended to define a new alias. Currently, `glslc --help` completely omits all ray-tracing and mesh shader stages (like `rgen`, `rahit`, etc.), even though they are fully supported in the code. I want to fix the outdated `-fshader-stage` help text while also standardize the behavior of the strings accepted by `#pragma shader_stage()` and the CLI flags. 
If you still think is improper I can close this PR. But please fix the inconsistent behavior between toolchain and the commands.
Test suite I'll commit later.

dneto0:
> Thanks for your response. But it's not the new aliases. Actually it's the same as custom pragma directives of shaderc (in libshaderc_util/src/shader_stage.cc), such as `#pragma shader_stage(raygen)`. In addition, the `tesseval` and `tesscontrol` are also not defined by glslang. I submit this PR is not intended to define a new alias. 

Ah, I see.  It's been so long I forgot about the `#pragma` mechanism.

> Currently, `glslc --help` completely omits all ray-tracing and mesh shader stages (like `rgen`, `rahit`, etc.), even though they are fully supported in the code. I want to fix the outdated `-fshader-stage` help text while also standardize the behavior of the strings accepted by `#pragma shader_stage()` and the CLI flags. If you still think is improper I can close this PR. But please fix the inconsistent behavior between toolchain and the commands. Test suite I'll commit later.

I'll build on top of this to add tests. (I'll push to your tree, thanks for allowing maintainers to edit).

dneto0 (approved):
with the new tests i added, this LGTM.
Thanks for your patience.

LBV2012-26:
> with the new tests i added, this LGTM. Thanks for your patience.

Sorry and it slipped my mind about adding test suit. Thank you for your review and merge.

PR #1587, opened 2026-07-30
update deps

dneto0:


dneto0:
this pulls in enhancements to spirv-tools validator, where it can better validate layouts in the presence of descriptor heap

dneto0:
TBR=dj2

PR #1589, opened 2026-08-01
Add NonSemantic.Shader.DebugInfo.100 options for Nvidia Nsight and/or RenderDoc source debugging

DannyArends:
### What

Adds two compile options that make shaderc emit NonSemantic.Shader.DebugInfo.100, equivalent to glslang's -gV / -gVS:

- shaderc_compile_options_set_generate_nonsemantic_debug_info, equivalent to -gV
- shaderc_compile_options_set_generate_nonsemantic_debug_source, equivalent to -gVS

Additionally, I have wired up matching CompileOptions::SetGenerateNonSemanticDebugInfo() / SetGenerateNonSemanticDebugSource() in the C++ wrapper.

### Why

set_generate_debug_info only emits OpLine/OpSource (glslang -g). Tools like NVIDIA Nsight Graphics need NonSemantic.Shader.DebugInfo.100 for function- and call-site-level source correlation (Flame Graph, Top-Down/Bottom-Up Calls). shaderc had no way to request it (see #1391), so callers had to drop to glslang or dxc directly. The pull requests makes these new flag independent from generate_debug_info so plain -g stays available as a fallback for shaders that fail to compile withNonSemantic info enabled.

### Usage: a Nsight-ready build

```C
    shaderc_compile_options_t options = shaderc_compile_options_initialize();
    shaderc_compile_options_set_generate_debug_info(options);               // -g:   OpLine/OpSource
    shaderc_compile_options_set_generate_nonsemantic_debug_source(options); // -gVS: NonSemantic + source
    shaderc_compile_options_set_optimization_level(options, shaderc_optimization_level_zero);
```

-g provides the line table, -gVS provides the call graph; Nsight uses both. Optimization must stay at zero. The size/performance passes can mangle or strip NonSemantic info even with the strip-guards in place.

### Notes

- Build against glslang >= 15.0.0; earlier versions point every DebugFunction at the same file, corrupting name/line correlation.
- Some shaders fail to compile with NonSemantic enabled; since the new flags are independent of -g, this is now a non-issue since builds need to opt in.

### DIFF
--- a/libshaderc_util/src/compiler.cc
+++ b/libshaderc_util/src/compiler.cc
@@ -363,6 +363,8 @@ std::tuple<bool, std::vector<uint32_t>, size_t> Compiler::Compile(
   std::vector<uint32_t>& spirv = compilation_output_data;
   glslang::SpvOptions options;
   options.generateDebugInfo = generate_debug_info_;
+  options.emitNonSemanticShaderDebugInfo = generate_nonsemantic_debug_info_;
+  options.emitNonSemanticShaderDebugSource = generate_nonsemantic_debug_source_;
   options.disableOptimizer = true;
   options.optimizeSize = false;
   // Note the call to GlslangToSpv also populates compilation_output_data.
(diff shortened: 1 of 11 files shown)


DannyArends:
Update: I have signed the CLA, let me know if anything else is needed from my side.

libshaderc_util/src/compiler.cc
```diff
@@ -454,28 +456,43 @@ void Compiler::SetForcedVersionProfile(int version, EProfile profile) {
 
 void Compiler::SetWarningsAsErrors() { warnings_as_errors_ = true; }
 
-void Compiler::SetGenerateDebugInfo() {
-  generate_debug_info_ = true;
-  for (size_t i = 0; i < enabled_opt_passes_.size(); ++i) {
-    if (enabled_opt_passes_[i] == PassId::kStripDebugInfo) {
-      enabled_opt_passes_[i] = PassId::kNullPass;
+void Compiler::UnstripDebugInfoPasses() {
```
dneto0: The function name reads weirdly.

I'd prefer:  `RemoveStripDebugInfoPass`.

Not a huge issue though.

dneto0 (changes requested):
This change should also include a command-line option on glslc, and please at least one unit test for each option, and glslc-based tests.

DannyArends:
Thanks for the review @dneto0. 

I've pushed changes addressing all points raised:

- Renamed `UnstripDebugInfoPasses` to `RemoveStripDebugInfoPass`.
- Added `-gV` and `-gVS` to glslc, with matching `--help` entries (updated the `HelpParameters` test output to match).
- Added unit tests for each option at both the C and C++ API layers, asserting that `NonSemantic.Shader.DebugInfo.100` is imported in the disassembly.
- Added glslc-based tests: `OptionGV` checks the import, and `OptionGVS` checks the import plus a source-bearing `DebugSource`.

A minor note on the `OptionGVS` test: both `-gV` and `-gVS` emit a `DebugSource`. The only structural difference is that `-gVS` adds a second operand (the embedded-source `OpString`). The test asserts two operands with an ID-independent regex (`DebugSource %\w+ %\w+`) rather than just matching on the source substring (which on its own isn't a reliable discriminator). Happy to adjust if you'd prefer a different approach.

I ran ctest --output-on-failure and everything passes locally.

PR #1592, opened 2026-08-21
Fix GLSLANG_ENABLE_INSTALL when SHADERC_SKIP_INSTALL is ON

### LINKED ISSUE #1435: `SHADERC_SKIP_INSTALL` cmake option broken
Running cmake with the `SHADERC_SKIP_INSTALL` option will break cmake:

```
...
**CMake Error: install(EXPORT "glslang-targets" ...) includes target "SPIRV" which requires target "SPIRV-Tools-opt" that is not in any export set.**
-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```
I think the issue is https://github.com/google/shaderc/blob/main/third_party/CMakeLists.txt#L90:
```
    set(GLSLANG_ENABLE_INSTALL $<NOT:${SKIP_GLSLANG_INSTALL}>)
```
I don't believe that "$<NOT" generator expression works during cmake's configure time, only at cmake's generation time.

rootkiller6788:
### Problem

Running `cmake` with `SHADERC_SKIP_INSTALL=ON` fails during generation:

```
CMake Error in third_party/glslang/CMakeLists.txt:
  install(EXPORT "glslang-targets" ...) includes target "SPIRV" which
  requires target "SPIRV-Tools-opt" that is not in any export set.
```

### Root cause

`third_party/CMakeLists.txt` sets

```cmake
set(GLSLANG_ENABLE_INSTALL $<NOT:${SKIP_GLSLANG_INSTALL}>)
```

Generator expressions are not evaluated at configure time. When `SHADERC_SKIP_INSTALL` is `ON`, the variable holds the literal string `$<NOT:ON>`, which CMake's `if()` treats as truthy, so glslang still enables its install rules. Since SPIRV-Tools' install is skipped in that configuration, glslang's `install(EXPORT glslang-targets)` references targets (`SPIRV`, `glslang`) whose dependency `SPIRV-Tools-opt` is not in any export set, and generation fails.

### Fix

Compute the boolean directly instead of relying on a generator expression:

```cmake
if (SKIP_GLSLANG_INSTALL)
  set(GLSLANG_ENABLE_INSTALL OFF)
else()
  set(GLSLANG_ENABLE_INSTALL ON)
endif()
```

### Verification

- `cmake -DSHADERC_SKIP_INSTALL=ON ...` previously failed with the error above; with the fix it configures successfully and glslang's `cmake_install.cmake` contains no install rules.
- The default configuration (without `SHADERC_SKIP_INSTALL`) still configures successfully and glslang's install rules are unchanged.

Fixes #1435


YGXXD:
I encountered the same problem.

PR #1598, opened 2026-09-10
Fix null pointer dereference on SPIR-V assembly failure

### LINKED ISSUE #1597: Null pointer dereference in shaderc SPIR-V assembly handling crashes glslc and API consumers
## Summary

`shaderc_assemble_into_spv()` returns a result object with a null internal `spv_binary` when SPIR-V assembly fails, but `shaderc_compilation_result_spv_binary::GetBytes()` unconditionally dereferences it at `libshaderc/src/shaderc_private.h:82`. Every caller that reads result bytes on a failed assembly crashes: the `glslc` CLI, the C API (`shaderc_result_get_bytes`), and the C++ API (`cbegin()`/`cend()`).

This is a denial-of-service of the compiling process via an attacker-supplied `.spvasm` file or assembly string. No memory corruption beyond the null read occurs, and no data is exposed. A one-line fix is proposed in the linked PR.

## Affected code

- `libshaderc/src/shaderc_private.h:82` — `GetBytes()` derefs `output_data_->code` without checking `output_data_`.
- `libshaderc/src/shaderc.cc:713-751` — `shaderc_assemble_into_spv()` only calls `SetOutputData()` inside the `if (assembling_succeeded)` branch; on failure `output_data_` stays null.
- `glslc/src/file_compiler.cc:182-184` — `EmitCompiledResult` builds a `string_piece` from `result.cbegin()`/`cend()` before checking success (checked later at line 207).
- `libshaderc/include/shaderc/shaderc.hpp:90,98` — `cbegin()`/`cend()` route to `shaderc_result_get_bytes`.

The GLSL-compile result type (`shaderc_compilation_result_vector`) is already null-safe; only the assembly result type is affected.

## Repro

```bash
# 1. Build shaderc (Release or ASAN)
# 2. Invalid assembly input:
printf 'not spv at all\n' > t3.spvasm
# 3. Crash:
glslc t3.spvasm -o /dev/null
#   -> SEGV (exit 139 Release; ASAN: null deref at address 0x0)
# 4. C API: shaderc_result_get_bytes() on failed assembly -> SEGV
# 5. Control: valid assembly compiles and exits 0
```

Reported to the Google OSS VRP (g.co/vulnz).

esoalba:
Fixes #1597

`shaderc_assemble_into_spv()` leaves `output_data_` null when assembly fails, but `GetBytes()` unconditionally dereferences it, crashing `glslc`, the C API (`shaderc_result_get_bytes`), and the C++ API (`cbegin()`/`cend()`) on any invalid `.spvasm` input.

This adds a null check before the dereference. Verified: invalid assembly now errors gracefully (exit 1 with an assembly error message) instead of segfaulting; valid assembly still compiles; C API returns `nullptr` on failure.

Unit test added (3cbbe7a): `AssembleStringTest.FailedAssemblyGetBytesIsNull` asserts that a failed assembly result has null bytes and zero length. Without the fix the test crashes with a null pointer dereference (SIGSEGV); with the fix it passes. Full test suite: 151 tests in `shaderc_test`, 126 in `shaderc_cpp_test`, all passing.

Reported via Google OSS VRP (g.co/vulnz).

### DIFF
--- a/libshaderc/include/shaderc/shaderc.h
+++ b/libshaderc/include/shaderc/shaderc.h
@@ -282,13 +282,13 @@ typedef struct shaderc_compile_options* shaderc_compile_options_t;
 // Any function operating on shaderc_compile_options_t must offer the
 // basic thread-safety guarantee.
 SHADERC_EXPORT shaderc_compile_options_t
-    shaderc_compile_options_initialize(void);
+shaderc_compile_options_initialize(void);
 
 // Returns a copy of the given shaderc_compile_options_t.
 // If NULL is passed as the parameter the call is the same as
 // shaderc_compile_options_init.
-SHADERC_EXPORT shaderc_compile_options_t shaderc_compile_options_clone(
-    const shaderc_compile_options_t options);
+SHADERC_EXPORT shaderc_compile_options_t
+shaderc_compile_options_clone(const shaderc_compile_options_t options);
 
 // Releases the compilation options. It is invalid to use the given
 // shaderc_compile_options_t object in any future calls. It is safe to pass
@@ -398,8 +398,7 @@ SHADERC_EXPORT void shaderc_compile_options_set_suppress_warnings(
 // a value listed in shaderc_env_version.  The 0 value maps to Vulkan 1.0 if
 // |target| is Vulkan, and it maps to OpenGL 4.5 if |target| is OpenGL.
 SHADERC_EXPORT void shaderc_compile_options_set_target_env(
-    shaderc_compile_options_t options,
-    shaderc_target_env target,
+    shaderc_compile_options_t options, shaderc_target_env target,
     uint32_t version);
 
 // Sets the target SPIR-V version. The generated module will use this version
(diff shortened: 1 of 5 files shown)


dneto0 (changes requested):
This should get a unit test also.

dneto0:
> This should get a unit test also.

I've added unit tests.  And rebased on top of #1603

dneto0 (approved):

esoalba:
Thanks @dneto0 — appreciate the review, and for taking the fix further. Hardening all the result getters against a null result object plus the NullResultObj and GetBytes* tests is a more complete fix than my original change, and the rebase on #1603 is appreciated.
On the test request: I had a complementary end-to-end regression test ready to push — shaderc_assemble_into_spv on invalid input, then asserting shaderc_result_get_bytes() returns nullptr on the failed result (verified it SIGSEGVs without the null check). I held off since your tests already cover the null-safety at the unit level. Happy to add it as an integration-level check if you think it's worth having; otherwise I'll leave the branch as-is.

PR #1603, opened 2026-09-11
Protect shaderc_compilation_result_t against null obj

dneto0:


libshaderc/include/shaderc/shaderc.h
```diff
@@ -282,13 +282,13 @@ typedef struct shaderc_compile_options* shaderc_compile_options_t;
 // Any function operating on shaderc_compile_options_t must offer the
 // basic thread-safety guarantee.
 SHADERC_EXPORT shaderc_compile_options_t
-    shaderc_compile_options_initialize(void);
+shaderc_compile_options_initialize(void);
```
dneto0, line 285: unfortunately clang-format has changed....
It's causing some spurious noise in this file.

The real changes are that I added more comments to the getters below.

dj2 (approved):

